PR #4, opened 2023-01-02
My two cents on mostro - list order command.

arkanoider:
My two cents ( to start I hope ):

Added the list offers action, you can request from nostril ( or cli when i work on that ) a command to request a pretty formatted table of a mostro like this:
mostro_table

Mostro will scan from db all orders with requested status from cli and send back a direct message for the user with the table.

Will work on cli side to create a specific command to make the trick.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -164,3 +164,22 @@ pub async fn next_event_kind(pool: &SqlitePool) -> anyhow::Result<i64> {
         Ok(10000)
     }
 }
+
+pub async fn find_orders_by_event_status(
+    pool: &SqlitePool,
+    status: &String,
+) -> anyhow::Result<Vec<crate::models::Order>> {
+    let order = sqlx::query_as!(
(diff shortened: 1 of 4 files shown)


grunch:
Hi @arkanoider this is great add, but this should be on a client not in mostro, we don't need to ask to mostro database because all this data is already on nostr relays, we should ask this to relays, I encourage you to do this same PR in [mostro-cli](https://github.com/MostroP2P/mostro-cli)

arkanoider:
You're right @grunch , I am looking to create the counterpart of this in the CLI. I am just trying to understand from NIPs if I can create a request with a range of values going from 10000 to 19000. Creating a single REQ for every single possible event is not so nice imo. Will look into it!

grunch:
Great, thanks to let me know about nip33, now with the implementation of it you don't need to worry about querying different kinds, we are going to work only with events with kind 30000.

Closing this issue

PR #9, opened 2023-03-16
Dev yadio quote for market price orders

arkanoider:
Hi @grunch !

Introduced a quote creation using Yadio API. Give a look.

Before hold invoice is created we check if order has 0 has sats amount --> this means user selected market price.

To be tested.


grunch (approved):
LGTM

PR #17, opened 2023-03-18
Adds dockerized relay for development purposes

bilthon:
I also modified the README a bit, with simple instructions on how to run the container. But for this to be complete, the `src/utils.rs` should also be modified to include the local relay URL (`ws://localhost:7000`) when in development mode.

I'm not sure how to do this, so I leave that open to alternatives. There could be logic that selects between two different sets of relays, depending on the execution context. Or maybe that list of relays could be loaded from an external configuration file or environment variables, and we have a different set of these for development and production mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /Cargo.lock
 .env
 mostro.db*
+
+# Relay data
+/relay/data/*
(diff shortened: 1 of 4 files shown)


arkanoider:
@bilthon welcome aboard!

PR #20, opened 2023-03-24
Hold invoice with market price

arkanoider:
Hi @grunch !

Finally I had some time to work on market price:

- Moved show_hold_invoice function in util.rs.
- Now both takebuy and take sell are up to work with market price order logic.

Everything need to be tested obviously, but we can move on a bit!


grunch (approved):
great contribution, tACK

PR #27, opened 2023-04-04
Scheduler

arkanoider:
Still in draft, now we can:

- Add env variable for hours of valid order
- The scheduler to check order is spawned in anothor tokio thread, I have to loop it so it runs always ( need to be tested deeply to avoid wrong behaviour of main loop)
- We have scheduled check of order older than creation time + env_time_hours, I have to make deletion from db of all entries.
- Will need testing at home cause at work i have issue to simulate all the flow

Cleaned up the code, spawn in the scheduler.rs was overcomplicating. Set 1 minute check at the moment.


grunch (approved):
LGTM

PR #47, opened 2023-04-19
Support to set local relay port through environment variable

diazemiliano:
Relay local port variable through environment variable.
If not set will default to `7000`

### DIFF
--- a/relay/docker-compose.yml
+++ b/relay/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     image: scsibug/nostr-rs-relay
     container_name: nostr-relay
     ports:
-      - '7000:8080'
+      - '${MOSTRO_RELAY_LOCAL_PORT:-7000}:8080'
     volumes:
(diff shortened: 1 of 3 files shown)


relay/docker-compose.yml
```diff
@@ -5,7 +5,7 @@ services:
     image: scsibug/nostr-rs-relay
     container_name: nostr-relay
     ports:
-      - '7000:8080'
+      - '${MOSTRO_RELAY_LOCAL_PORT:-7000}:8080'
```
grunch, line 8: why `-`7000 ?

diazemiliano: It was already the default. I just set it as default again if the user isn't specifying one through an env var. Maybe @bilthon can answer why he choose that port number.

grunch (approved):

grunch:
@diazemiliano thanks for your contribution, please for next contributions sign your commits

PR #69, opened 2023-05-04
Scheduler new tasks

arkanoider:
Hi @grunch ,

here a first draft pull request, just tested in a basic case with sell order published --> Taken without invoice --> After 15 minutes order has been republished.

I don't know if my idea is correct, but in the scheduler I check that order is WaitingBuyerInvoice or WaitingPayment and if it's older than 15 minutes I republish the order. Double check with you if it's not correct, anyway we have the scheduler running.


arkanoider:
Added delay on event rate sending. Not tested just compiled.

grunch (changes requested):
Testing I have this situation:

* Seller creates and market price order 
* Buyer take it
* Buyer didn't go further
* Order republished correctly but it doesn't clean the amount to 0

Same thing with orders created by buyer and took it by seller

arkanoider:
Thanks for testing @grunch !
Clear! Will look to fix all this and also the info string with a more coherent message.

arkanoider:
> Testing I have this situation:
> 
>     * Seller creates and market price order
> 
>     * Buyer take it
> 
>     * Buyer didn't go further
> 
>     * Order republished correctly but it doesn't clean the amount to 0
> 
> 
> Same thing with orders created by buyer and took it by seller

FIxed.

grunch:
> > Testing I have this situation:
> > ```
> > * Seller creates and market price order
> > 
> > * Buyer take it
> > 
> > * Buyer didn't go further
> > 
> > * Order republished correctly but it doesn't clean the amount to 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Same thing with orders created by buyer and took it by seller
> 
> FIxed.

Great, now I found that in this case it should set to null `master_buyer_pubkey` too, and in the case where the buyer is the maker the `master_seller_pubkey` should be set to null too.

arkanoider:
> Testing I have this situation:
> 
>     * Seller creates and market price order
> 
>     * Buyer take it
> 
>     * Buyer didn't go further
> 
>     * Order republished correctly but it doesn't clean the amount to 0
> 
> 
> Same thing with orders created by buyer and took it by seller

FIxed.

> > > Testing I have this situation:
> > > ```
> > > * Seller creates and market price order
> > > 
> > > * Buyer take it
> > > 
> > > * Buyer didn't go further
> > > 
> > > * Order republished correctly but it doesn't clean the amount to 0
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Same thing with orders created by buyer and took it by seller
> > 
> > 
> > FIxed.
> 
> Great, now I found that in this case it should set to null `master_buyer_pubkey` too, and in the case where the buyer is the maker the `master_seller_pubkey` should be set to null too.

Hi @grunch ,workin on this...just a doubt...

Seller creates order --> Buyer takes it --> No other actions from buyer --> Order reset to pending and also set to None master_buyer_pubkey

Buyer creates order --> Seller takes it --> No other actions from seller --> Order reset to pending and also set to None master_seller_pubkey ( and also master buyer ?? Don't understand if both need to go to None )

grunch:
> 

the counterpart fields must be set to null again, we need to reset the order to the exact same state that it was before was taken.

arkanoider:
> > 
> 
> the counterpart fields must be set to null again, we need to reset the order to the exact same state that it was before was taken.

Yes...just the counterpart right? not both...

grunch:
> > > 
> > 
> > the counterpart fields must be set to null again, we need to reset the order to the exact same state that it was before was taken.
> 
> Yes...just the counterpart right? not both...

the correct response is the order record must be the same than before, in this particular case only the counterpart should be reset but in other cases some fields of the maker should be reset too

arkanoider:
> > > > 
> > > 
> > > 
> > > the counterpart fields must be set to null again, we need to reset the order to the exact same state that it was before was taken.
> > 
> > 
> > Yes...just the counterpart right? not both...
> 
> the correct response is the order record must be the same than before, in this particular case only the counterpart should be reset but in other cases some fields of the maker should be reset too

Yes clear...i am.doing all the case chart to understand

arkanoider:
Working on scheduler @arkanoider , will add taken_at management to align to lnp2pbot cancel/republish logics.

src/db.rs
```diff
@@ -494,3 +515,28 @@ pub async fn update_order_seller_dispute(
...
+    let created_at = Timestamp::now();
+
+    let rows_affected = sqlx::query!(
+        r#"
+            UPDATE orders
+            SET
+            seller_dispute = ?1
+            WHERE id = ?2
+        "#,
+        seller_dispute,
```
grunch, line 534: `seller_dispute` is missing here

src/scheduler.rs
```diff
@@ -68,13 +69,17 @@ pub async fn cron_scheduler(sched: &JobScheduler) -> Result<(), anyhow::Error> {
...
-            let exp_seconds = var("EXP_SECONDS").unwrap().parse::<u64>().unwrap() / 60;
-
+            let exp_seconds = var("EXP_SECONDS").unwrap().parse::<u64>().unwrap_or(900) / 60;
 
-            info!("I run async every minute id {:?} - check for order to republish for late actions", uuid);
+            info!("Check for order to republish for late actions of users");
 
             let older_orders_list = crate::db::find_order_by_seconds(pool.as_ref().unwrap()).await;
 
+            // for or in older_orders_list.unwrap().into_iter(){
```
grunch, line 78: please remove commented code

grunch (approved):
LGTM 😄

Issue #70, opened 2023-05-06
Set a max routing fee on send_payment()


grunch:
We need to set a max routing fee for every payment we do to avoid leaking sats attack

https://github.com/MostroP2P/mostro/blob/main/src/lightning/mod.rs#L153

Here is an implementation on @lnp2pbot that can be used as a reference.

https://github.com/lnp2pBot/bot/blob/main/ln/pay_request.js#L13

arkanoider:
https://github.com/MostroP2P/mostro/pull/74

PR #74, opened 2023-05-09
max fee introduction

arkanoider:
Did a quick pr for #70.

@grunch review when you can.


grunch (approved):
this is looking gooooooooooooooooood!

PR #80, opened 2023-05-12
Bug fixes and refactoring related bot fee

grunch:


grunch:
I fixed the errors I found but this needs to be heavily tested, this can't have errors before production

PR #88, opened 2023-05-19
Retry test

arkanoider:
Basic retry mechanics for Yadio API for market price calculation. Exiting gently in case of 4 failed tentatives.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -3,7 +3,7 @@ use crate::lightning::LndConnector;
 use crate::messages;
 use crate::models::Yadio;
 use crate::{db, flow, RATE_EVENT_LIST};
-use anyhow::{Context, Ok, Result};
+use anyhow::{Context, Result};
 use dotenvy::var;
 use log::{error, info};
 use mostro_core::order::{NewOrder, Order, SmallOrder};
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -13,25 +13,57 @@ use nostr_sdk::prelude::*;
...
+use std::thread;
 use tokio::sync::mpsc::channel;
 use tonic_openssl_lnd::lnrpc::invoice::InvoiceState;
 use uuid::Uuid;
 
+pub async fn retries_yadio_request(req_string: &String) -> Result<reqwest::Response> {
+    let res = reqwest::get(req_string)
+        .await
+        .context("Something went wrong with API request, try again!")?;
+    // .json::<Yadio>()
```
grunch, line 25: Can you remove commented code? :D

src/util.rs
```diff
@@ -13,25 +13,57 @@ use nostr_sdk::prelude::*;
...
+
+    Ok(res)
+}
+
 /// Request market quote from Yadio to have sats amount at actual market price
 pub async fn get_market_quote(fiat_amount: &i64, fiat_code: &str, premium: &i64) -> Result<i64> {
     // Add here check for market price
     let req_string = format!(
-        "https://api.yadio.io/convert/{}/{}/BTC",
+        "https://api.yadio1.io/convert/{}/{}/BTC",
```
grunch, line 36: is this a typo or is it a new yadio domain?

grunch (changes requested):
thanks, just few comments down below

src/util.rs
arkanoider: Lol.was testing error...didn't think you look to it..will fix it!

PR #90, opened 2023-06-08
nip12 fiat name filter tag

arkanoider:
Added nip12 filter for fiat name ( tag f ) when order event is published by mostro.


grunch (approved):
LGTM!

PR #91, opened 2023-06-17
small test on fee calculations

arkanoider:
Small test added on main to try fee calculations.
Really stupid RNG, but i did not want to add dependecies.


grunch (approved):
LGTM! thanks buddy

arkanoider:
Welcome back boss! We need to move on!

PR #99, opened 2023-07-25
Modifies the init script to read database url from the toml file & print steps

bilthon:
It also only removes database files if they exist.


grunch (approved):
LGTM!

PR #100, opened 2023-07-28
Uses awk instead of grep in the init script, making it more robust

bilthon:
This is an improvement over the last commit.

The problem there was with older versions of `grep`, like the BSD version included with macOS, lack the `-P` option. So it would work in Linux, where it was tested, but fail on a Mac. 

This version uses `awk` instead, which is more universally compatible across different Unix-like systems.


grunch (approved):
LGTM

PR #104, opened 2023-08-20
Some improvements on config file and db file

arkanoider:
Hi @grunch @bilthon !

Some improvements on config files and db file:

- Switched `MOSTRO_CONFIG` variable from `lazy_static` to `OnceLock` which is included in [std lib](https://github.com/rust-lang/rust/pull/105587) now. Maybe you could have to update to latest Rust. Next time I will try to use `OnceCell` to remove also variable of `RATE_EVENT_LIST` from lazy static so we can remove the not maintained crate.
- I changed `settings.toml` to `settings.tpl.toml` in .init_db.sh for compatibility and made config path the one in which db file is searched at startup.

Now what is happening is that user will launch in root folder the db init then launches mostro that creates path for settings and db file, then will print out to copy both file in the folder.
Maybe we could improve in some way that file copy and db creation are automated, but I don't if it's a good idea...

Waiting for advice... ;)

### DIFF
--- a/init_db.sh
+++ b/init_db.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 echo "Reading database URL from settings.toml..."
-DATABASE_URL=$(awk -F'"' '/url *= */ {print $2}' settings.toml)
+DATABASE_URL=$(awk -F'"' '/url *= */ {print $2}' settings.tpl.toml)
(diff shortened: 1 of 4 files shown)


grunch (approved):
LGTM! 🥳

PR #107, opened 2023-09-07
removed global user vector

arkanoider:
Hi @grunch ,

to be tested with user rates...
I removed another global from main, more aligned to rust best practise. Putting here as draft, will look again tomorrow.


arkanoider:
Removed some comments, compiles correctly. To be fully tested anyway.

Just `MOSTRO_CONFIG` remain as static variable.

grunch (approved):
LGTM!!! 🚀

PR #109, opened 2023-09-13
Removed the case of panic if db is empty in scheduler

arkanoider:
Hi @grunch ,

small fix on scheduler.rs in case db is empty. This could lead to a panic for unwrap. Fixed it and removed to unwrap also.

Give a look!


grunch (approved):
LFGTM!!!!!

PR #116, opened 2023-09-20
Clap

arkanoider:
Hi @grunch ,

some improvements:

- Added `clap `crate for easier management of other parameters from terminal.
- Removed `pretty_env_logger` from cargo.toml, we can use just `tracing `crate.
- Inserted here from previous draft PR the tracing with less verbosity ( this could be enhanced, but for now it's working )


grunch (approved):
LGTM! 🚀

PR #117, opened 2023-09-22
used sdk code to avoid code duplication - reduced user_rate.rs file

arkanoider:
Hi @grunch ,

this need to be completely tested. Reduced a lot `user_rate.rs` file because nostr-sdk has implemented the same logic. So we are duplicating code. I admit i didn't test anything, but it should work, and in case we coul surely fix eventual issues to use the sdk code and avoid code duplication.

Waiting for comments...


grunch (approved):
LGTM

PR #128, opened 2023-10-10
New scheduler

arkanoider:
Hi @grunch ,

simplified a lot scheduler:

-- removed crate
-- used simple thread with sleep - we don't need so precise granularity like chron task at the moment.
-- easier and simpler to read
-- Least but not last now I don't use local instances of connect to relays all is working cloning main connect instace, much more clean and efficient.

Not tested because I am at work, let me know your thoughts about.


grunch (approved):
LGTM

PR #134, opened 2023-11-01
added configurable interval for user rate send events

arkanoider:
Hi @grunch ,

added configurable interval in seconds to send user rates events to relays.


grunch (approved):
LGTM 🪨

PR #154, opened 2023-11-29
Initial phase of refactor of mostro code to update protcol message

arkanoider:
@grunch ,

initial PR to fix something on` app.rs` and some include in sparse files...


grunch (approved):
LGTM

PR #163, opened 2023-12-12
Crud update

arkanoider:
Hi @grunch ,

first step to improve and clean database calls:

- Removed a quite big number of functions  in `db.rs` , as told in chat we need to fix an issue of rust compiler in the case of calls inside a spawned task, we will work on.
- I took the occasion to remove `log `crate moving to `tracing `in all files, it's more powerful for future uses and manage also all the functions of `log`.

let me know...


grunch (approved):
what a great PR, LGTM 🚀

PR #170, opened 2023-12-19
removed fn add_order from db.rs

arkanoider:
Hi @grunch ,

removed `fn add_order` from `db.rs`.

Update `mostro-core` before compile, I have added a default attrib to Order struct.

Not tested!


grunch (approved):
LGTM

PR #178, opened 2023-12-29
Includes buyer & seller pubkeys and order creation time upon order transition to Active state

bilthon:
With the seller as the maker, the information about each trade party's real public keys should be sent upon the hodl invoice confirmation with the `buyer_pubkey ` & `seller_pubkey ` as per the [documentation](https://mostro.network/messages/seller_pay_hold_invoice.html). This however was not happening. Another thing that was missing was the order creation time.

Please note that for this to work the dependencies must be updated as I'm using a modified `mostro-core` with modifications that can be seen [here](https://github.com/bilthon/mostro-core/commit/f594e32782d2b9f67afa5d662435ab7e3177d259). Basically just adding these two fields to the `SmallOrder` struct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ uuid = { version = "1.3.0", features = [
   "serde",
 ] }
 reqwest = { version = "0.11", features = ["json"] }
-mostro-core = "0.4.5"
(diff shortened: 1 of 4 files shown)


grunch:
fixed on #179 and #180

PR #183, opened 2024-01-06
No currency exchange rate

arkanoider:
Hi @grunch @bilthon ,

did some refactoring for error messages on `Yadio` request, should be better now:

- If 4 retries are over we get: 
```Rust
MostroError::NoAPIResponse => write!(f, "Price API not answered - retry")
```

-  No currency exchange on Yadio list:
```Rust
MostroError::NoCurrency => write!(f, "Currency requested is not present in the exchange list, please specify a fixed rate"),
```
-  Malformed answer from Yadio
```Rust
MostroError::MalformedAPIRes => write!(f, "Malformed answer from exchange quoting request")
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,6 +9,7 @@ pub enum MostroError {
     MinAmountError,
     WrongAmountError,
     NoAPIResponse,
+    NoCurrency,
 }
 
 impl std::error::Error for MostroError {}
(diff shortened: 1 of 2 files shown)


arkanoider:
Added a check to use fixed amount rate specified by user when new order is created.

Example:
```mostro-cli neworder -k sell -a 20000 -c XMR -f 10 -m SEPA```

Now takes 20000 sats as the amout the user want to exchange for 20 XMR, before we asked to Yadio api anyway causing wrong quote or no quote imo.

Let me know

arkanoider:
Added an extra check in `order.rs` to avoid negative amount ( with fixed or calculated ), don't know if it's necessary, but could be an extra safety check.

Feel free to merge it or no...easy to remove!

PR #193, opened 2024-01-29
removed thread sleep calls for sqlite update and create

arkanoider:
Hi @grunch ,

this could be a workaround to avoid the ugly `thread::sleep`.

Basically it seems that issue are present when show_hold_invoice is called near in time to another crud operation, my idea is to move the order variable as parameter and not only order_id, with this we call in the problematic case only one crud operation.
Also we remove a read from db because in show_hold_invoice we have the full order as parameter.

Give a look, I tried the sell flow and now invoice is correctly saved in mostro.db.


src/app/cancel.rs
```diff
@@ -91,8 +91,6 @@ pub async fn cancel_action(
                             &order.id
                         );
                     }
-                    // Update db
-                    let mut order = order.update(pool).await?;
```
bilthon, line 95: I'm not that familiar with the codebase yet, but this removal doesn't seem to be replaced by anything. Does it mean it was redundant? the `pool` seems to be a db connection so maybe this is flushing/committing changes to the db?

bilthon (approved):
This is surely about my pay grade, but since I was asked and it's actually a simple change, plus I ran the code and everything went smooth I can say I approve it.

I've added a brief comment regarding a few lines that were removed. It's more about satisfying my curiosity than addressing any critical issues.

src/app/add_invoice.rs
```diff
@@ -131,8 +130,6 @@ pub async fn add_invoice_action(
     let seller_pubkey = XOnlyPublicKey::from_str(&seller_pubkey)?;
     // We save the invoice on db
     order.buyer_invoice = Some(pr.clone());
-    let order = order.update(pool).await?;
```
arkanoider, line 134: Removed because in any case below we will update database.
If `preimage.is_some` we will call with `update_order_event` with order variable completely updated. So this call is redundant and could also be the cause of the missed database update.

src/app/add_invoice.rs
```diff
@@ -184,7 +181,7 @@ pub async fn add_invoice_action(
             None,
             &buyer_pubkey,
             &seller_pubkey,
-            order.id,
```
arkanoider, line 187: Same here, if we fall here I updated `show_hold_invoice` with order as parameter and not `order.id`.
`Show_hold_invoice` inside does the update call on order, and this has also the advantage of the order recall from database using just `order.id`.

src/app/cancel.rs
arkanoider, line 95: Same here, this update was redundant because the same call is made in `update_order_event`

arkanoider:
@bilthon @grunch 
Added a bunch of comment to explain with a bit more of detail what is my idea...hope it's more clear.
I think that could be the cause also of the wrong behaviour in some case of db operations.

PR #203, opened 2024-02-29
Invoice amount ln

arkanoider:
Hi @grunch ,

this is comprehensive of failure payment ( to improve ) and your fixes.

- improved is_valid_invoice fn: now checks inside which kind of invoice we have (ln address or bolt)
- Check if lnaddress amount is 0.
- removed a match on MostroError which imo was redundant.
- Some cosmetics to code

Still need to check amount when new buy order is created with invoice in one shot.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,4 +44,4 @@ clap = { version = "4.4.4", features = ["derive"] }
 lnurl-rs = "0.4.0"
 
 [patch.crates-io]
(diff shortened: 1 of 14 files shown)


grunch:
[this was a replacement](https://github.com/MostroP2P/mostro/pull/204) of this one? can we close this PR?

arkanoider:
Done!

PR #204, opened 2024-02-29
amount for ln address checks and many cosmetics

arkanoider:
Hi @grunch ,

this should be just with my modification above main trunk.

Give a look and we can discuss about in case.


grunch (approved):
LGTM, great add buddy 🥳

PR #209, opened 2024-03-12
refactoring admin_settle.rs

arkanoider:
Hi @grunch ,

give a look here, much reused code throwaway here in `admin_settle.rs` with reuse of r`elease.rs` do_payment fn.

Should be good imo...give a look, if I have not missed some state they should do the same logic.


grunch (approved):
LGTM

PR #222, opened 2024-03-24
Message refactoring

### LINKED ISSUE #211: Send message to buyer after the payment fails
Message should have a new action `payment-failed` and indicate the payment attempts left.

With this information clients could create a message like this:

`I tried to send you the sats but the payment of your invoice failed, I will try ${attempts} more times in ${pendingPaymentWindow} minutes window, please check your node/wallet is online`

arkanoider:
Hi @grunch ,

immagine

@bilthon meme syndrome got me and in the end i did this shit:

- Create a static global var for client instance, really too much functions using the same parameter.
 ```Rust
static NOSTR_CLIENT: OnceLock<Client> = OnceLock::new();
 ```
This has the advantage to call everywhere, just with import the `client` instance from main.
So we can do this:
```Rust
 NOSTR_CLIENT.get().unwrap().send_event(event).await?;
```
- This lead me to create two new wrapper `send_new_order_msg` and `send_cant_do_msg` to embed all the logic in a function, I think the message part is more clean now, we used a lot of send_dm around creating message...give a look if you like more.

- Also a lot of `my_key` parameters are remove from function, just get it in `send_dm` function, the result function just calls inside the keys like that:
```Rust
pub async fn send_dm(receiver_pubkey: &XOnlyPublicKey, content: String) -> Result<()> {
    info!("DM content: {content:#?}");
    // Get mostro keys
    let sender_keys = crate::util::get_keys().unwrap();

    let event =
        EventBuilder::new_encrypted_direct_msg(&sender_keys, *receiver_pubkey, content, None)?
            .to_event(&sender_keys)?;
    info!("Sending event: {event:#?}");
    NOSTR_CLIENT.get().unwrap().send_event(event).await?;

    Ok(())
}
```

And as meme pointed out this should fix #211 !

I could do the same for the few dispute messages if you approve, I did some quick testing and seems ok, but need some deeper tests.

@Catrya can you hear me??? :smile:

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -36,7 +36,7 @@ use tokio::sync::Mutex;
 
 pub async fn run(
     my_keys: Keys,
-    client: Client,
+    client: &Client,
     ln_client: &mut LndConnector,
     pool: Pool<Sqlite>,
     rate_list: Arc<Mutex<Vec<Event>>>,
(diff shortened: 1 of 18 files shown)


grunch (approved):
great refactoring thanks buddy, LGTM! 🚀

PR #229, opened 2024-03-25
updated nostr-sdk lib

arkanoider:
@grunch 

not tested, but compiled with latest nostr-sdk.

Updated also some other deps.

The only missing point is sqlx which involves also the crud crate...we will speak about it.


grunch (approved):
LGTM

Issue #220, opened 2024-03-22
Do not allow opening disputes if the seller has already released


Catrya:
When the seller releases the sats it no longer makes sense for either party to open a dispute because both have already fulfilled their part, but currently mostro is allowing it

I made a test sale and after I released the sats and they arrived to the buyer I opened a dispute without problems, then i resolved the dispute see:

```
Mostro sent you this message for order id: 0e419e99-b278-4974-b77c-a02705c5de44 at 12:56 date - 22/03/2024

Action: Release
Content: None

Mostro sent you this message for order id: 0e419e99-b278-4974-b77c-a02705c5de44 at 12:56 date - 22/03/2024

Action: PurchaseCompleted
Content: None

Mostro sent you this message for order id: 0e419e99-b278-4974-b77c-a02705c5de44 at 12:56 date - 22/03/2024

Action: RateUser
Content: None

Mostro sent you this message for order id: 0e419e99-b278-4974-b77c-a02705c5de44 at 12:57 date - 22/03/2024

Action: DisputeInitiatedByYou
Content: None

Mostro sent you this message for order id: 0e419e99-b278-4974-b77c-a02705c5de44 at 13:59 date - 22/03/2024

Action: AdminSettle
Content: None

```

arkanoider:
Starting fix via #230

grunch:
closing it via #230

PR #230, opened 2024-03-25
Added a check in dispute.rs

arkanoider:
Hi @grunch ,

this is for #220 from @Catrya .

Added a check like this in `dispute.rs`
```Rust
fn check_order_status_for_dispute(status: &str) -> bool {
    match Status::from_str(status).unwrap() {
        Status::Success => true,
        _ => false,
    }
}

```

I just check that order is in success state to block dispute creation for now, feel free to add other states.

### DIFF
--- a/src/app/dispute.rs
+++ b/src/app/dispute.rs
@@ -14,13 +14,6 @@ use sqlx::{Pool, Sqlite};
 use sqlx_crud::traits::Crud;
 use tracing::{error, info};
 
-fn check_order_status_for_dispute(status: &str) -> bool {
-    match Status::from_str(status).unwrap() {
-        Status::Success => true,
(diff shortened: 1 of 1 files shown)


Catrya:
Hi @arkanoider, if users canceled the order they can also open a dispute, so I thought that mostro could also verify that the order was not cancelled.

but speaking with @grunch he says that to open a dispute what we must do is check if the status of the order is active or fiat-sent, which is a much better idea than mine

arkanoider:
Sounds good! Wiill fix.

arkanoider:
@Catrya @grunch 

Dispute should be available only for `Active` and `FiatSent` orders now.

grunch (approved):
LGTM

PR #251, opened 2024-04-11
mostro stats first release

arkanoider:
Hi @grunch ,

here we go with the fixed pr.

Stats are basic, but it's a starting point. Now a file is saved every minutes with latest data collected. 
Values are updated inside mostrod at every event received, file is called `mostro_stats.json` with a simple struct like this:

```Rust
pub struct MostroMessageStats {
    received: u64,
    new_order: u64,
    new_dispute: u64,
    release: u64,
    bytes_recv: u64,
    max_msg_bytes: u64,
    min_msg_bytes: u64,
    last_msg_bytes: u64,
}
```


grunch:
due inactivity let's close this PR, if you have a new proposal let's create a new issue to re-think about this PR

PR #253, opened 2024-04-12
User rating ack message carries the rating back

bilthon:
Just makes mostro reply with the user rating in the ack message. This facilitates updating the UI in case messages come out of order and doesn't represent too much of an overhead.

I'm also updating the docs.


grunch (approved):
LGTM

PR #259, opened 2024-04-15
Idea to catch the great part of the errors

arkanoider:
Hi @grunch ,

added some catcher in the loop at the end, I think it's better and as improvement in the next days we could remove some matches in the functions below, cause now all errors are caught in the loop printing a warn message with error.
Example:

```Rust
Action::NewOrder => {
        if let Err(e) =
            order_action(msg, &event, &my_keys, &pool).await
        {
            tracing::warn!(
                "Error in {} with context {}",
                action,
                e
            );
        }
    }
```

Did a quick test with your cli command and it works:
```Rust
2024-04-15T21:07:43.464556Z  INFO mostrod::scheduler: Next tick for removal of older orders is Mon Apr 15 21:08:43 2024
2024-04-15T21:07:43.479253Z  INFO mostrod::scheduler: I run async every 60 minutes - checking for failed lighting payment
2024-04-15T21:08:25.779619Z  WARN mostrod::app: Error in NewOrder with context Incorrect invoice
2024-04-15T21:08:43.465449Z  INFO mostrod::scheduler: Check older orders and mark them Expired - check is done every minute
2024-04-15T21:08:43.466700Z  INFO mostrod::scheduler: Next tick for removal of older orders is Mon Apr 15 21:09:43 2024
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,9 +151,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.82"
+version = "1.0.81"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
+checksum = "0952808a6c2afd1aa8947271f3a60f1a6763c7b912d210184c5149b5cf147247"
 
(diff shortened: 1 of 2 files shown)


grunch (approved):
LGTM

PR #261, opened 2024-04-16
Added the check on min sats allowed in a new order

arkanoider:
Hi @grunch ,

as pointed out yesterday by @Catrya , when creating a new order user is allowed to make sats amount lower than minimum ( we checked only maximum ).

Then order is publishe and if someone takes it it's not allowed to do, but I think at this point is more meaningful to refuse new order with sats amout lower than minimum of settings.

Give a look, just a simple check added.

### DIFF
--- a/src/app/order.rs
+++ b/src/app/order.rs
@@ -50,10 +50,12 @@ pub async fn order_action(
             return Ok(());
         }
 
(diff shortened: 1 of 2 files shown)


grunch (approved):
LGTM 🥳

PR #269, opened 2024-04-18
small fix to return from settle_seller_hold_invoice if there are errors

arkanoider:
Hi @grunch,

small fix in  fn settle_seller_hold_invoice about returning in errors in case of issues during the flow


grunch (approved):
great job as always LGTM

PR #272, opened 2024-04-21
Removed other unwraps and minor cosmetics

arkanoider:
Hi @grunch,

as the title says some more unwraps removed and small cosmetics.


grunch (approved):
LGTM

Issue #264, opened 2024-04-16
bug: two admins can take the same dispute


Catrya:
I took a dispute using 2 admin users.
The buyer and seller received 2 messages that their dispute had been accepted.

When an administrator accepts a dispute, Mostro must not allow another administrator to accept it, to avoid contradictions.

that It's not a priority

```
Mostro sent you this message:

{"order":{"version":1,"id":"ea84da8c-8048-4061-80f0-78907b0b0f11","pubkey":null,"action":"admin-took-dispute","content":{"peer":{"pubkey":"00000018............................................................05"}}}}

Mostro sent you this message:

{"order":{"version":1,"id":"ea84da8c-8048-4061-80f0-78907b0b0f11","pubkey":null,"action":"admin-took-dispute","content":{"peer":{"pubkey":"000000909........................................................10"}}}}

Bye Bye!
```

arkanoider:
Hi @Catrya ,

clear! In your experience if  a solver ( not the admin ) takes a dispute an admin could take it too or we should think about only one operator assigned to a single dispute?

In this case I can fix just checking the state of the dispute is different from `initiated`.

Let me know your opinion!

Catrya:
Hi @arkanoider
 Well, first I had thought that if a solver took the dispute no one else could do it (not even the admin). But @grunch is right when he says that all the powers must be maintained to the admin.

So now I think it would be better if:
- if a solver takes a dispute, then no other solver can take it, but the admin can
- if an admin takes a dispute, no one else can take it

arkanoider:
Hi @Catrya ,

question for @grunch , I was looking to fix this and I have a doubt:

Dispute open --> A solver takes it and dispute goes `inProgress `status --> if an admin takes over the solver than the solver won't be able to close it right?

Dispute open --> A solver takes it and dispute goes `inProgress `status : at this stage is it possible that the admin takes the dispute in another status without closing it? Lookin to code it seems that all next step are conclusive for the dispute, but my doubt is just if there is another state after InProgress that could be, let me say a wait state. Then an admin takes over the solver in this state...what happens to dispute status...now it will be resetted to inProgress, and sounds right to me, but I am just asking if it's correct!

grunch:
> Dispute open --> A solver takes it and dispute goes `inProgress `status --> if an admin takes over the solver than the solver won't be able to close it right?

Correct, only the admin can take a dispute with status = `in-progress`, so no other solver can take a dispute from another solver
 
[...]

PR #277, opened 2024-05-01
Idea for fix issue #264

arkanoider:
Hi @grunch ,

first idea for fix the issue #264 .

Modified a bit this function `pubkey_event_can_solve`

```Rust
pub async fn pubkey_event_can_solve(
    pool: &Pool<Sqlite>,
    ev_pubkey: &PublicKey,
    status: Status,
) -> bool {
    if let Ok(my_keys) = crate::util::get_keys() {
        // Is mostro admin taking dispute?
        if ev_pubkey.to_string() == my_keys.public_key().to_string()
            && matches!(status, Status::InProgress | Status::Initiated)
        {
            return true;
        }
    }

    // Is a solver taking a dispute
    if let Ok(solver) = find_solver_pubkey(pool, ev_pubkey.to_string()).await {
        if solver.is_solver != 0_i64 && status == Status::Initiated {
            return true;
        }
    }

    false
}
```

In this way if a `solver` takes first a dispute we have to check that dispute is in `Initiated` state, which (imo) is the only one a solver can take.
While an admin can overtake also `InProgress` state, so now he will become the solver and other solvers cannot take the dispute again.

You will see that I removed also `pubkey_event_can_solve` from `admin_cancel.rs `and `admin_settle.rs` because ( still imo ) they are redundat because of `is_assigned_solver` that is responsible to check if the pubkey of message is the solver.
If, for example, an admin overtake a solver dispute he will be the new solver so the check below should be enough.

```Rust
   match is_assigned_solver(pool, &event.pubkey.to_string(), order_id).await {
        Ok(false) => {
            send_cant_do_msg(
                None,
                Some("Dispute not taken by you".to_string()),
                &event.pubkey,
            )
            .await;

            return Ok(());
        }
        Err(e) => {
            error!("Error checking if solver is assigned to order: {:?}", e);
            return Ok(());
        }
        _ => {}
    }
```

Let me know!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,9 +202,9 @@ dependencies = [
 
 [[package]]
 name = "async-wsocket"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d253e375ea899cb131b92a474587e217634e7ea927c24d8098eecbcad0c5c97a"
+checksum = "5c38341e6ee670913fb9dc3aba40c22d616261da4dc0928326d3168ebf576fb0"
 dependencies = [
  "async-utility",
  "futures-util",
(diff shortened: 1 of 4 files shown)


grunch (approved):
LGTM

PR #279, opened 2024-05-03
Info mostro message

arkanoider:
Added scheduled message with mostro infos - 5 minutes interval.

There was an error on file `src\lightning\mod.rs` line 161 here:

```Rust
let invoice = decode_invoice(payment_request)?;
        let hash = bytes_to_string(&payment_hash);
        let mostro_settings = Settings::get_mostro();
```

That function `signable_hash` did the trick, please check @grunch if it's right, seems we have the same result as before with just one call:
```Rust
let invoice = decode_invoice(payment_request)?;
        let payment_hash = invoice.payment_hash();
        let payment_hash = payment_hash.to_byte_array();
        let hash = bytes_to_string(&payment_hash);
        let mostro_settings = Settings::get_mostro();
```


src/nip33.rs
```diff
@@ -68,3 +69,61 @@ pub fn order_to_tags(order: &Order) -> Vec<(String, String)> {
...
+        ),
+        // hold_invoice_cltv_delta
+        (
+            "hold_invoice_cltv_delta".to_string(),
+            ln_settings.hold_invoice_cltv_delta.to_string(),
+        ),
+        // invoice_expiration_window
+        (
+            "invoice_expiration_window".to_string(),
+            ln_settings.hold_invoice_cltv_delta.to_string(),
```
grunch, line 120: This value is duplicated

src/scheduler.rs
```diff
@@ -21,10 +21,32 @@ pub async fn start_scheduler(rate_list: Arc<Mutex<Vec<Event>>>) {
...
+        Ok(keys) => keys,
+        Err(e) => return error!("{e}"),
+    };
+
+    tokio::spawn(async move {
+        loop {
+            info!("Sending info about mostro");
+
+            let tags = crate::nip33::info_to_tags(&mostro_pubkey.public_key());
+            let id = format!("Mostro npub {} info event", mostro_pubkey.public_key());
```
grunch, line 40: `d tag` value shouldn't have white spaces and must be more concise, let's change it to
 
```rust
let id = format!("info-{}", mostro_pubkey.public_key());
```

grunch (changes requested):
Can you add Mostro version and current main branch commit? if commit number is too much work just add the version and we do the commit part later

src/nip33.rs
arkanoider: Ups! Will remove

src/scheduler.rs
arkanoider: Yep!

src/nip33.rs
arkanoider: Added both in info message, need to be tested.

grunch (approved):
LGTM

PR #285, opened 2024-05-17
Range order

arkanoider:
Hi @grunch ,

some steps ahead also on mostro side, what's implemented now ( to be deeply tested anyway ):

- Updated `cargo.toml` for mostro-core 5.10
- Removed not null conditions from db file to avoid errors on db entry creation due to `max_amount` and `min_amount` null possible value
- If order is with range now we check if sats amount conversion is inside limit for both values, if order is no range order same logic as before
- Implemented same logic of lnP2P bot to diminish max amount or ranged order and mark it as successful ( this part need to be fully tested and improved in style )

Missing things in my mind:

- `TakeBuy` and `TakeSell` checks if amount received is in the range of order and in case send back a message to user.
- Testing everything


src/app/admin_take_dispute.rs
```diff
@@ -81,8 +81,12 @@ pub async fn admin_take_dispute_action(
     };
 
     let mut new_order = order.as_new_order();
-    new_order.master_buyer_pubkey = order.master_buyer_pubkey.clone();
-    new_order.master_seller_pubkey = order.master_seller_pubkey.clone();
+    new_order
+        .master_buyer_pubkey
+        .clone_from(&order.master_buyer_pubkey);
```
grunch, line 86: why is it better to use clone_from()

arkanoider: Suggestion of clippy, related to memory efficiency.

grunch:
I'm testing it and after try to create an order
```bash
$ mcli neworder -k sell -c ars -f 1000 -m MP
```

I get this in logs
```log
2024-05-17T13:29:29.219718Z  WARN mostrod::app: Error in NewOrder with context error returned from database: (code: 1299) NOT NULL constraint failed: orders.min_amount
```

grunch:
I could create a new range order, how can I take it? for this you should write new specs for different situation, here an example:

https://mostro.network/messages/take_sell.html

arkanoider:
> I could create a new range order, how can I take it? for this you should write new specs for different situation, here an example:
> 
> https://mostro.network/messages/take_sell.html

Yep! We will have to update specs with new range order case, let's complete the code to have a workin version and then I will write specs

arkanoider:
Added logic to check if fiat amount value sent from CLI is in range.
This won't passe CI test because of mostro-core new version

grunch:
closing it in favor of #286

PR #290, opened 2024-05-25
fix #288

### LINKED ISSUE #288: if it is a range order do not allow creating an order with the amount of sats
Currently, when you are going to make an order by range you can type:
```
mcli neworder -k sell -f 50-70 -a 200 -c cup -m test 
```
and it will be published without a problem.

If you are going to create an order by ranges, mostro should not allow you to also put the amount in sats because it does not make sense.

arkanoider:
@grunch 

this should fix #288


Catrya (approved):
works well

Catrya (approved):

Catrya (approved):

Catrya (approved):

grunch (approved):
LGTM

Issue #284, opened 2024-05-15
Add maker's reputation to nostr order event


grunch:
We should include the maker's completed orders and reputation when we publish the nostr event with the order

arkanoider:
You mean when we publish the event of a neworder? So using also number of completed orders and rating of the order creator?

arkanoider:
Should be fixed ( message format apart ) with #291 right?

PR #291, opened 2024-05-27
added maker rating on new order creation

arkanoider:
Hi @grunch ,

quick add to fix issue #284, now just inserted user rating in new order creation. Maker rating should be added with new fn:

```Rust
fn create_rating_string(rating: Option<Rating>) -> String {
    if rating.is_some() {
        format!("{:?}", rating.unwrap(),)
    } else {
        "No user reputation received".to_string()
    }
}
```

### DIFF
--- a/src/app/order.rs
+++ b/src/app/order.rs
@@ -55,12 +55,12 @@ pub async fn order_action(
                 _ => order.amount,
             };
 
-        // Check amount is positive - extra safety check
-        if quote < 0 {
-            let msg = format!("Amount must be positive {} is not valid", order.amount);
-            send_cant_do_msg(order.id, Some(msg), &event.pubkey).await;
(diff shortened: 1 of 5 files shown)


arkanoider:
Yes we have to make it better! I think maybe that just :
`"rating" : "3"`
could be enough?

arkanoider:
@bilthon quick improvement on the format, now you should have the json formatted `Rating` struct in the `rating` tag

```Rust
fn create_rating_string(rating: Option<Rating>) -> String {
    if let Some(rating) = rating {
        if let Ok(rating_json) = rating.as_json() {
            rating_json
        } else {
            "bad format value".to_string()
        }
    } else {
        "none".to_string()
    }
}


```

grunch (approved):
LGTM

PR #293, opened 2024-05-29
doc: updates the 'add-invoice' content

bilthon:
The documentation is not up to date with the contents of the `add-invoice` message.  A third field was added to the `payment_request` array and if not present the message is rejected.


grunch (approved):
LGTM

PR #294, opened 2024-05-29
feat: handles message parsing error & prints an error log

bilthon:
Without this mostro just silently rejects malformed or out of date messages from clients without even an error log entry.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -34,6 +34,7 @@ use sqlx::{Pool, Sqlite};
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tracing::info;
+use tracing::error;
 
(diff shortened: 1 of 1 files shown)


arkanoider (commented):
For me it's a merge without lookin, good to have more log on error in these case.

bilthon:
Perhaps more could be done, like sending a `CantDo` message to the client. But logging it was the minimal thing I could think of and that as an operator can help you understand what happened and why the client's message got dropped.

arkanoider:
> Perhaps more could be done, like sending a `CantDo` message to the client. But logging it was the minimal thing I could think of and that as an operator can help you understand what happened and why the client's message got dropped.

I agree 100%

grunch (approved):
LGTM

PR #300, opened 2024-06-19
Fix how we publish range fiat amount

grunch:
Code refactoring


arkanoider (commented):
Great work in direction of string removal @grunch ! LGTM!

Issue #303, opened 2024-06-21
Panic error when an admin tries to settle a canceled order


Catrya:
If buyer and seller open a dispute, but before solver resolves it they do a cooperative cancellation, if solver tries to do Settle, it will cause panic error and Mostro stops. Mostro should tell the solver that the order has already been cancelled and not try to pay the invoice.

![Captura desde 2024-06-20 21-41-17](https://github.com/MostroP2P/mostro/assets/140891948/7430a81b-e8b4-4dc5-b886-5865efd81ff3)

arkanoider:
Preparing a hotfix to avoid panic.

PR #310, opened 2024-06-25
hotfix #303

arkanoider:
@grunch @Catrya 

hotfix to avoid panic #303.

### DIFF
--- a/src/app/admin_settle.rs
+++ b/src/app/admin_settle.rs
@@ -6,7 +6,7 @@ use crate::NOSTR_CLIENT;
 
 use anyhow::{Error, Result};
 use mostro_core::dispute::Status as DisputeStatus;
-use mostro_core::message::{Action, Message};
+use mostro_core::message::{Action, Message, MessageKind};
 use mostro_core::order::{Order, Status};
 use nostr_sdk::prelude::*;
 use sqlx::{Pool, Sqlite};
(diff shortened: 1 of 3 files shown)


Catrya (approved):
There is no panic error anymore!

Catrya:
Although when admin does admsettle, the message Mostro sent is ¨Dispute not taken by you¨
But it should be something like ¨This order has already been cancelled by its users"
Maybe we could consider putting this in a new issue?

arkanoider:
> Although when admin does admsettle, the message Mostro sent is ¨Dispute not taken by you¨ But it should be something like ¨This order has already been cancelled by its users" Maybe we could consider putting this in a new issue?

I will continue on this issue @Catrya , will look to have a more consistent error for user in these case.

grunch:
> Although when admin does admsettle, the message Mostro sent is ¨Dispute not taken by you¨ But it should be something like ¨This order has already been cancelled by its users" Maybe we could consider putting this in a new issue?

Probably we should start doing the new stuff following what is asked in #312, so we don't have to send any text string to the user, we should send a message with an action that let know to the client what happened, you can see the list of actions [here](https://docs.rs/mostro-core/latest/mostro_core/message/enum.Action.html) so we can use `CooperativeCancelAccepted` but if you think we should use a different one we can discuss it here

Catrya (commented):
There is no panic error anymore, and returns the message "cooperative-cancel-accepted"

But now if there is a cooperative cancellation and the admin tries to cancel anyway, the message it returns is "Can't settle an order with status different than dispute!". But before it allowed it even though it had no effect

Catrya (commented):
There is no panic error anymore, and returns the message "cooperative-cancel-accepted"

But now if there is a cooperative cancellation and the admin tries to cancel anyway, the message it returns is "Can't settle an order with status different than dispute!". But before it allowed it even though it had no effect

arkanoider:
I will

arkanoider:
> But now if there is a cooperative cancellation and the admin tries to cancel anyway, the message it returns is "Can't settle an order with status different than dispute!". But before it allowed it even though it had no effect

I think that's just because I added the check only in `admin_settle` file and not in `admin_cancel`. Doing it in my new branch on Mostro.

PR #317, opened 2024-07-01
now cooperative cancelled order are no more managed by admin 

arkanoider:
Hi @grunch @Catrya ,

moved here PR for fix of #303 .

Added also check in `admin_cancel` file.

Let's see if now it's working.


Catrya (approved):
ACK

grunch (approved):
LGTM

PR #323, opened 2024-07-05
admin-settle-add-validation

Catrya:
Now if someone who is not a solver or admin tries to settle an order, they will get CantDo. If they are a solver or admin and the order is not in dispute, mostro will notified that they cannot do because of the status, and if it is in dispute but they did not taken it, will notified correctly.

The way Mostro has been working until now, if a user tries to settle an order regardless of its status and regardless of whether they are an admin or not they will see the message `Dispute not taken by you`, (except if it was cooperatively canceled). 

I haven't yet incorporated the the new actions without text strings

@grunch @arkanoider please test it :)

### DIFF
--- a/src/app/admin_settle.rs
+++ b/src/app/admin_settle.rs
@@ -1,4 +1,4 @@
-use crate::db::{find_dispute_by_order_id, is_assigned_solver};
+use crate::db::{find_dispute_by_order_id, is_assigned_solver, find_solver_pubkey};
 use crate::lightning::LndConnector;
 use crate::nip33::new_event;
 use crate::util::{send_cant_do_msg, send_dm, settle_seller_hold_invoice, update_order_event};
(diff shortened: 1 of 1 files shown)


arkanoider (commented):
For me it's a merge! Let's see Francisco...

@Catrya just do a cargo fmt and commit again to format all file in a nice way!

arkanoider (approved):

src/app/admin_settle.rs
```diff
@@ -29,24 +47,6 @@ pub async fn admin_settle_action(
         return Err(Error::msg("No order id"));
     };
 
-    match is_assigned_solver(pool, &event.pubkey.to_string(), order_id).await {
```
grunch, line 32: same that I asked in the other PR

could you explain why moving this block code to bottom?

grunch (changes requested):

Catrya:
I will close it and then upload it from the new fork I made

Catrya:
I will close it and then upload it from the new fork I made

PR #324, opened 2024-07-06
admin-cancel-add-validation

Catrya:
Now if someone who is not a solver or admin tries to cancel an order, they will get CantDo. If they are a solver or admin and the order is not in dispute, mostro will notified that they cannot do because of the status, and if it is in dispute but they did not taken it, will notified correctly.

I haven't yet incorporated the the new actions without text strings

@grunch @arkanoider, please test it ;)

### DIFF
--- a/src/app/admin_cancel.rs
+++ b/src/app/admin_cancel.rs
@@ -1,6 +1,6 @@
 use std::str::FromStr;
 
-use crate::db::{find_dispute_by_order_id, is_assigned_solver};
+use crate::db::{find_dispute_by_order_id, is_assigned_solver, find_solver_pubkey};
 use crate::lightning::LndConnector;
 use crate::nip33::new_event;
 use crate::util::{send_cant_do_msg, send_dm, update_order_event};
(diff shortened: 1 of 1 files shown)


arkanoider (approved):
Idem for this PR, good for me!

src/app/admin_cancel.rs
```diff
@@ -15,6 +15,24 @@ use sqlx::{Pool, Sqlite};
 use sqlx_crud::Crud;
 use tracing::{error, info};
 
+pub async fn pubkey_event_can_solve(pool: &Pool<Sqlite>, ev_pubkey: &PublicKey) -> bool {
```
grunch, line 18: You are duplicating this same function https://github.com/MostroP2P/mostro/blob/main/src/app/admin_take_dispute.rs#L16

We should never duplicate functions, if you need this function use the one we already have declared

src/app/admin_cancel.rs
```diff
@@ -28,24 +46,6 @@ pub async fn admin_cancel_action(
         return Err(Error::msg("No order id"));
     };
 
-    match is_assigned_solver(pool, &event.pubkey.to_string(), order_id).await {
```
grunch, line 31: could you explain why moving this block code to bottom?

Catrya:
I will close it and then upload it from the new fork I made

PR #329, opened 2024-07-15
Introduces a in-memory price cache

bilthon:
Introduces an in-memory price cache that can be used to reduce the latency of dynamically priced order creation.

### DIFF
--- a/src/bitcoin_price.rs
+++ b/src/bitcoin_price.rs
@@ -1,0 +1,38 @@
+use std::collections::HashMap;
+use std::sync::RwLock;
+use serde::Deserialize;
+use anyhow::Result;
+use tracing::info;
+use once_cell::sync::Lazy;
+
+const YADIO_API_URL: &str = "https://api.yadio.io/exrates/BTC";
+
+#[derive(Debug, Deserialize)]
+struct YadioResponse {
+    #[serde(rename = "BTC")]
+    btc: HashMap<String, f64>,
+}
+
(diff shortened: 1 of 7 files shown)


src/bitcoin_price.rs
```diff
@@ -0,0 +1,38 @@
...
+
+const YADIO_API_URL: &str = "https://api.yadio.io/exrates/BTC";
+
+#[derive(Debug, Deserialize)]
+struct YadioResponse {
+    #[serde(rename = "BTC")]
+    btc: HashMap<String, f64>,
+}
+
+static BITCOIN_PRICES: Lazy<RwLock<HashMap<String, f64>>> = Lazy::new(|| RwLock::new(HashMap::new()));
```
grunch, line 16: it's not possible to do this same thing using `OnceLock` as we've been doing for `MOSTRO_CONFIG` and `NOSTR_CLIENT` ?

arkanoider: Hi @grunch @bilthon ,

my two cents about this. I think it's right, because we have to update that static, this is exactly what [RwLock](https://doc.rust-lang.org/std/sync/struct.RwLock.html) does! On the contrary I used [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) is used to set a value once and then just reading it.
Lazyness is also correct otherwise we have to init the RwLock in the definition, so I agree with this solution.

Hope that answered to your doubt @grunch

grunch: perfect! thanks buddy

grunch (approved):
This is an amazing contribution, thanks @bilthon

PR #330, opened 2024-07-16
starting idea for resubscribe of invoices held

arkanoider:
Hi @grunch,

starting idea for ln invoice resubscribe.

Just added a db function to  get orders with held_time set and state is active.

Embedded in a fn invoice stuff to reuse the fn around and called at main level at the moment, still can't see other use cases then at restart of daemon, in case with the scheduler we have to be sure to don't call again resub on the same invoice I suppose. Maybe a field in db could be useful in this case? Or a new order field?

Waiting for comments if I am on the right path.


grunch (approved):
tACK

grunch:
This works really good, amazing job 😺

Issue #314, opened 2024-06-28
Allowing the user to cancel a take


dtonon:
As per subject, is there a technical reason because is not currently possible to cancel a take?
If so, it could be useful to explain that on the payment page.

arkanoider:
It should be not so complicated imo. Not seeing any issue about this. 
@grunch is the boss of protocol anyway, let's see what he thinks about.

grunch:
> As per subject, is there a technical reason because is not currently possible to cancel a take?
> If so, it could be useful to explain that on the payment page.

no reason, we should do that definitely

arkanoider:
Removed #311 because it's not good. Lookin back to the code @grunch i think we have all the logic to cancel a take, it's already done in `cancel.rs` and I think it works with CLI.
Now i cannot test it, maybe tonite. But is it just the fact that now works only from CLI:

- Here https://github.com/MostroP2P/mostro/blob/8588a2313a45ce9e542139e2d591970307d29b1a/src/app/cancel.rs#L59
- Here https://github.com/MostroP2P/mostro/blob/8588a2313a45ce9e542139e2d591970307d29b1a/src/app/cancel.rs#L65

These two calls do exactly the reset to pending state of the specific order.

@Catrya could you try a quick test with actual version?

arkanoider:
> As per subject, is there a technical reason because is not currently possible to cancel a take? If so, it could be useful to explain that on the payment page.

@dtonon could you confirm me that you only tested with web client? Probably the take cancel is yet functional, but only from CLI.

arkanoider:
> > @Catrya could you try a quick test with actual version?
> > Hi @arkanoider, understood, I'm out of home so I have less time to test, I'm prioritizing PR like the unsubscribe, @Catrya can you test this one please? 🥹
> 
> Yes of course, yesterday I also tested it in the current version and I'm almost sure that the cancel a take didn't work, but in a few minutes I'll test it again and I'll tell you guys

Great! From the code it should reset order to `Pending` state. Try this sequence using `mostro-cli`:

- Create a `sell` order.
- Take it with `takesell` from cli.
- Then from same user do a `cancel` command (not admin_cancel!).

Maybe I lost some memory on the code, but it should revert to pending state.

Catrya:
Canceling a take buy does work, it returns the order from waiting-payment to pending status. @arkanoider

grunch:
[...]

PR #331, opened 2024-07-16
Simple idea for cancel take

arkanoider:
Hi @grunch,

a simple idea for #314.

Just adding some checks inside cancel.rs could lead to rewind to pending state the taken order imo.

Let me know your thoughts.

In this way cancel command would cancel a take too.

### DIFF
--- a/src/app/cancel.rs
+++ b/src/app/cancel.rs
@@ -93,14 +93,14 @@ pub async fn cancel_action(
     }
 
     if order.kind == OrderKind::Sell.to_string()
-        && order.status == Status::WaitingBuyerInvoice.to_string()
+        && (order.status == Status::WaitingBuyerInvoice.to_string()
+            || order.status == Status::WaitingBuyerInvoice.to_string())
     {
         cancel_add_invoice(ln_client, &mut order, event, pool, my_keys).await?;
     }
 
     if order.kind == OrderKind::Buy.to_string()
-        && (order.status == Status::WaitingPayment.to_string()
-            || order.status == Status::WaitingBuyerInvoice.to_string())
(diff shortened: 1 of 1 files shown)


Catrya (commented):
It only allows to cancel a take of buy order, but not a take of a sell order.
Even though it cancels the take of buy order, it sends the message Not Allowed to the taker when cancell it.

arkanoider:
> It only allows to cancel a take of buy order, but not a take of a sell order.
> Even though it cancels the take of buy order, it sends the message Not Allowed to the taker when cancell it.

Thanks for the tests @Catrya !
Probably i did some typo logic should in the same way, will check. Message can be changed, no issue.

arkanoider:
> It only allows to cancel a take of buy order, but not a take of a sell order. Even though it cancels the take of buy order, it sends the message Not Allowed to the taker when cancell it.

You use the invoice in the take sell command @Catrya ?

arkanoider:
Closing this. Not the right way to do cancel take.

arkanoider:
Reopened PR - fixed.

Catrya (commented):
It allows users to cancel any take: buy or sell

But now this also happens:
1. taker takes the order and does his part
2. order goes to status `waiting-buyer-invoice` or `waiting-payment`
3. Maker sends `cancel` before doing his part
4. The order goes back to status `Pending` (and of course appears in listorders), but it should be cancelled

in the current version if the maker tries to cancel the order before doing his part Mostro doesn't allow it (it doesn't send a cantDo message but it just stays in the status waiting-buyer-invoice or waiting-payment)

Catrya (approved):
ACK
It works well!

grunch (approved):
LGTM

PR #341, opened 2024-08-07
Pow check of events incoming

arkanoider:
Hi @grunch ,

`mostrod` now checks the POW of incoming events.
Settings file now in `[mostro]` section has also pow value ( 0 means no pow ).

Added also pow requested by mostro in info event.

### DIFF
--- a/src/nip33.rs
+++ b/src/nip33.rs
@@ -168,7 +168,7 @@ pub fn info_to_tags(mostro_pubkey: &PublicKey) -> Vec<Tag> {
             vec![mostro_settings.fee.to_string()],
         ),
         Tag::custom(
-            TagKind::Custom(std::borrow::Cow::Borrowed("Pow")),
+            TagKind::Custom(std::borrow::Cow::Borrowed("pow")),
             vec![mostro_settings.pow.to_string()],
         ),
         Tag::custom(
--- a/src/nip33.rs
+++ b/src/nip33.rs
@@ -168,6 +168,10 @@ pub fn info_to_tags(mostro_pubkey: &PublicKey) -> Vec<Tag> {
             vec![mostro_settings.fee.to_string()],
         ),
         Tag::custom(
+            TagKind::Custom(std::borrow::Cow::Borrowed("Pow")),
+            vec![mostro_settings.pow.to_string()],
+        ),
+        Tag::custom(
             TagKind::Custom(std::borrow::Cow::Borrowed("hold_invoice_expiration_window")),
             vec![ln_settings.hold_invoice_expiration_window.to_string()],
         ),
--- a/settings.tpl.toml
+++ b/settings.tpl.toml
@@ -41,6 +41,8 @@ expiration_seconds = 900
 user_rates_sent_interval_seconds = 3600
 # Relay list event time interval
 publish_relays_interval = 60
+# Requested POW
+pow = 0
 
 
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -26,6 +26,7 @@ use crate::app::release::release_action;
 use crate::app::take_buy::take_buy_action;
 use crate::app::take_sell::take_sell_action;
 use crate::lightning::LndConnector;
+use crate::Settings;
 
 use anyhow::Result;
 use mostro_core::message::{Action, Message};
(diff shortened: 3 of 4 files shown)


src/nip33.rs
```diff
@@ -167,6 +167,10 @@ pub fn info_to_tags(mostro_pubkey: &PublicKey) -> Vec<Tag> {
             TagKind::Custom(std::borrow::Cow::Borrowed("fee")),
             vec![mostro_settings.fee.to_string()],
         ),
+        Tag::custom(
+            TagKind::Custom(std::borrow::Cow::Borrowed("Pow")),
```
grunch, line 171: please lower case to `pow`

arkanoider: yep!

arkanoider: Done!

grunch (changes requested):
I tested it with pow = 1, then I publish a new order with mostro-cli last commit and mostrod publish the order

If I try to cancel the event now I get `INFO mostrod::app: Not POW verified event!` and it doens't cancel, that is ok

arkanoider:
> I tested it with pow = 1, then I publish a new order with mostro-cli last commit and mostrod publish the order
> 
> If I try to cancel the event now I get `INFO mostrod::app: Not POW verified event!` and it doens't cancel, that is ok

I noticed this too...for example i tried with `pow=10 on mostrod` and `pow=9 on cli` and it can pass because could happen that you create an event on cli with id like:

0x`001`efrys....

0 - 4 bits to 0 so 4
0 - 4 bits to 0 so 4 other 0... 8 totally...
5 - in binary is 0101 so you have 10 zero leading bits and it passes...

I think that we need to think to pow as some robust one like 30 so cannot have a random lucky shot...if you put pow=1 on mostrod it's not so incredible that an event can be generated with the first digit that is zero.

This is my opinion, I can ask confirmation to @yukibtc in case or just look at the code.

grunch:
> > I tested it with pow = 1, then I publish a new order with mostro-cli last commit and mostrod publish the order
> > If I try to cancel the event now I get `INFO mostrod::app: Not POW verified event!` and it doens't cancel, that is ok
> 
> I noticed this too...for example i tried with `pow=10 on mostrod` and `pow=8 on cli` and it can pass because could happen that you create an event on cli with id like:
> 
> 0x`001`efrys....
> 
> 0 - 4 bits to 0 so 4 0 - 4 bits to 0 so 4 other 0... 8 totally... 1 - in binary is 0010 so you have 10 zero leading bits and it passes...
> 
> I think that we need to think to pow as some robust one like 30 so cannot have a random lucky shot...if you put pow=1 on mostrod it's not so incredible that an event can be generated with the first digit that is zero.
> 
> This is my opinion, I can ask confirmation to @yukibtc in case or just look at the code.

I see, great job, let's merge this 👶🏽

grunch (approved):
🐤

yukibtc:
> > I tested it with pow = 1, then I publish a new order with mostro-cli last commit and mostrod publish the order
> > If I try to cancel the event now I get `INFO mostrod::app: Not POW verified event!` and it doens't cancel, that is ok
> 
> I noticed this too...for example i tried with `pow=10 on mostrod` and `pow=8 on cli` and it can pass because could happen that you create an event on cli with id like:
> 
> 0x`001`efrys....
> 
> 0 - 4 bits to 0 so 4 0 - 4 bits to 0 so 4 other 0... 8 totally... 1 - in binary is 0010 so you have 10 zero leading bits and it passes...
> 
> I think that we need to think to pow as some robust one like 30 so cannot have a random lucky shot...if you put pow=1 on mostrod it's not so incredible that an event can be generated with the first digit that is zero.
> 
> This is my opinion, I can ask confirmation to @yukibtc in case or just look at the code.

Correct. A POW of 24-32 could be good.

---

Exists an option in the `Client` to set a min POW. Could be more efficient to use that instead of `Event::check_pow`, so the event will be almost immediately discarded instead of being fully parsed, verified and so on. The SDK execute these steps when receive a new relay message:
* Parse **raw** relay message
* Check message/event limits
* **Partially** parse event (only ID and pubkey) 
* Check blacklist (in-memory)
* Check POW <-- HERE
* Parse missing event fields and compose full event
* Check if it's expired
* Check if it's valid
* Check if exists in db or if was already seen
* Send notification to channels

You can set POW in `Client` with: 

```rust
let client = Client::default();
client.update_min_pow_difficulty(28);

// OR

let opts = Options::default().min_pow(28);
let client = Client::with_opts(signer, opts);
```

PR #350, opened 2024-09-02
Avoid a buyer taking the same order twice in status waiting-buyer-invoice

### LINKED ISSUE #311: A user can take a sell order multiple times in status waiting-buyer-invoice
A user can take a sell order multiple times while in status `waiting-buyer-invoice` ; each time they do so, a new event 38383 will be published.

If they take a range order and choose a fiat amount ¨X¨, when they take it again, even if they choose a different fiat amount, Mostro will ask them to send an invoice for the amount of Sats equivalent to ¨X¨.

I think Mostro updates the fiat amount correctly bc publishes an order discounting that amount of fiat when the trade ends. However, it does not update the amount of Sats each time the buyer takes the order.

Preventing a user from taking the same order twice in the "waiting-buyer-invoice" status helps avoid the other issue.

Example. sell order 100-700 cup
1st time taken: 500 cup (2219 sats)
Captura desde 2024-06-25 12-49-21

2nd time taken: 300 cup
Captura desde 2024-06-25 12-50-44

3rd time taken: 700 cup
Captura desde 2024-06-25 12-51-47

Catrya:
fix #311

### DIFF
--- a/src/app/take_sell.rs
+++ b/src/app/take_sell.rs
@@ -100,9 +100,9 @@ pub async fn take_sell_action(
         }
     };
 
-    // Buyer can take Pending or WaitingBuyerInvoice orders only
+    // Buyer can take Pending orders only
     match order_status {
-        Status::Pending | Status::WaitingBuyerInvoice => {}
+        Status::Pending => {}
         _ => {
             send_new_order_msg(
                 Some(order.id),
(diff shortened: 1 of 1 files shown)


src/app/take_sell.rs
```diff
@@ -100,9 +100,9 @@ pub async fn take_sell_action(
         }
     };
 
-    // Buyer can take Pending or WaitingBuyerInvoice orders only
+    // Buyer can take Pending orders only
     match order_status {
-        Status::Pending | Status::WaitingBuyerInvoice => {}
+        Status::Pending => {}
```
arkanoider, line 105: If @Catrya did a round of test and it work, for me it's a direct merge.

grunch (approved):
LGTM

PR #351, opened 2024-09-02
When taking an order, check the status first and then the quantity

### LINKED ISSUE #349: change check order when taking an order
When someone tries to take a range order that is not `pending`, if they do not put the amount in the range, Mostro returns `OutOfRangeFiatAmount`, but then if they put it within the range it returns `NotAllowedByStatus`.
I think Mostro should first check if it is in status `pending`, and then if it is within the fiat range or in the amount of sats allowed, and not the other way around.

Catrya:
fix #349 
Now when someone takes an order Mostro will first check the status of the order and then if it has a range it will check the quantity that was requested to be taken is within the range and not the other way around.

### DIFF
--- a/src/app/take_buy.rs
+++ b/src/app/take_buy.rs
@@ -45,20 +45,6 @@ pub async fn take_buy_action(
         return Ok(());
     }
 
-    // Get amount request if user requested one for range order - fiat amount will be used below
-    if let Some(am) = get_fiat_amount_requested(&order, &msg) {
-        order.fiat_amount = am;
-    } else {
-        send_new_order_msg(
-            Some(order.id),
-            Action::OutOfRangeFiatAmount,
-            None,
-            &event.sender,
-        )
-        .await;
-        return Ok(());
-    }
-
     let order_status = match Status::from_str(&order.status) {
         Ok(s) => s,
         Err(e) => {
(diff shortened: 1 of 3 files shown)


arkanoider:
@Catrya @grunch ,

i think this is correct, just don't know how `lnp2p` manage this. I mean now a range order cannot be taken from other people until it's completed ( with success or not ) . Is this correct or we want other users can take the same order while it's still active?

src/app/take_buy.rs
```diff
@@ -48,21 +48,7 @@ pub async fn take_buy_action(
...
-            Some(order.id),
-            Action::OutOfRangeFiatAmount,
-            None,
-            &event.pubkey,
-        )
-        .await;
-        return Ok(());
-    }
-
+
```
arkanoider, line 51: Moving this call below indeed is avoiding that a user can take another time the offer until it's completed ( successfully or not ). It could be good to me, but I don't know what is the behaviuor of lnp2p in this case, I mean a range order in lnp2p can have two pending orders simoultaneously?

Catrya: > Moving this call below indeed is avoiding that a user can take another time the offer until it's completed ( successfully or not ). It could be good to me, but I don't know what is the behaviuor of lnp2p in this case, I mean a range order in lnp2p can have two pending orders simoultaneously?

In lnp2pbot an order can only be taken in `pending` status. If it has a range, until it is completed it is not published again so that someone can take what is left of the range. I suppose that is the expected behavior for Mostro as well, I don't know if @grunch  want to do it differently here

But what is happening now in Mostro is that if someone takes an order with a range but it is already in any other status that is not `pending` because someone else is already taking it, or they are already taking it, if they put that they are going to take it with an amount outside the range, Mostro will respond `OutOfRangeFiatAmount`, and when that person later puts an amount within the range, Mostro will tell them that it is `NotAllowedByStatus`.

So what I propose is to save a step and have Mostro respond from the beginning that it cannot be taken `NotAllowedByStatus`

grunch (changes requested):
please solve the conflicts

arkanoider:
If conflicts are over I think this could be merged @grunch , make sense imo.

grunch (approved):
LGTM

PR #352, opened 2024-09-10
[DOC] Message actions to Clients

Catrya:
Add documentation with suggestions for messages that clients can show to users based on each action. 

I think that none of the actions that the client receives are missing, but it should be well reviewed. Also, please review the info that is in monospace format to verify if it is correctly.


grunch (approved):
LGTM

PR #364, opened 2024-10-07
Tonic-lnd

arkanoider:
First compiling version with new crate for tonic-lnd management.

[Fedimint](https://crates.io/crates/fedimint-tonic-lnd)

To be tested.

Really much easier with this crate, mostly aligned with old one.


src/lightning/mod.rs
```diff
@@ -40,9 +41,8 @@ impl LndConnector {
         let ln_settings = Settings::get_ln();
 
         // Connecting to LND requires only host, port, cert file, and macaroon file
-        let client = tonic_openssl_lnd::connect(
+        let client = fedimint_tonic_lnd::connect(
             ln_settings.lnd_grpc_host,
-            ln_settings.lnd_grpc_port,
```
grunch, line 45: I'm getting a panic on connection, can I have an example of what I need to change to connect?
```
thread 'main' panicked at /home/grunch/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tonic-0.10.2/src/client/grpc.rs:383:42:
path_and_query only is valid Uri: InvalidUriParts(InvalidUri(SchemeMissing))
```

arkanoider: Let me check I tested just that mostrod was starting, did not test a full trade, maybe I missed something.

PR #372, opened 2024-10-15
Fix settings.docker.toml to work with gprc and Polar

tincho70:
The lnd_grpc_host parameter must be set using the special DNS host.docker.internal so that Mostro (container) can communicate with Polar (host) via grpc.


grunch (approved):
LGTM

PR #393, opened 2024-11-13
feat: handles cancel request for already cancelled order & removes redundant check

bilthon:
When receiving a cancel request for an already cancelled order mostro just fails silently. This adds a check for that and makes mostro reply with a `cant-do` action response.

There's also a small redundant check that was removed.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for order cancellations to prevent processing of already canceled orders.
	- Improved logic for cooperative cancellations, ensuring funds are returned to the seller when applicable.

- **Bug Fixes**
	- Simplified condition checking for sell orders to streamline cancellation processes.

- **Documentation**
	- Updated messaging for buyers and sellers during the cancellation process to provide clearer communication.

### DIFF
--- a/src/app/cancel.rs
+++ b/src/app/cancel.rs
@@ -40,6 +40,21 @@ pub async fn cancel_action(
         }
     };
 
+    if order.status == Status::Canceled.to_string()
+        || order.status == Status::CooperativelyCanceled.to_string()
+        || order.status == Status::CanceledByAdmin.to_string()
+    {
+        error!("Order {} is already canceled", order_id);
+        send_cant_do_msg(
+            request_id,
+            Some(order_id),
+            Some("Order already canceled".to_string()),
+            &event.sender,
+        )
+        .await;
+        return Ok(());
+    }
+
     if order.status == Status::Pending.to_string() {
         // Validates if this user is the order creator
         if user_pubkey != order.creator_pubkey {
@@ -73,8 +88,7 @@ pub async fn cancel_action(
     }
 
     if order.kind == OrderKind::Sell.to_string()
-        && (order.status == Status::WaitingBuyerInvoice.to_string()
-            || order.status == Status::WaitingBuyerInvoice.to_string())
+        && (order.status == Status::WaitingBuyerInvoice.to_string())
     {
         cancel_add_invoice(ln_client, &mut order, event, pool, my_keys, request_id).await?;
     }


src/app/cancel.rs
```diff
@@ -40,6 +40,21 @@ pub async fn cancel_action(
...
 
+    if order.status == Status::Canceled.to_string()
+        || order.status == Status::CooperativelyCanceled.to_string()
+        || order.status == Status::CanceledByAdmin.to_string()
+    {
+        error!("Order {} is already canceled", order_id);
+        send_cant_do_msg(
+            request_id,
+            Some(order_id),
+            Some("Order already canceled".to_string()),
```
Catrya, line 51: The idea was to send messages with just the action, without a text string inside, so that clients can interpret that action and show a message to users with the text they want and in the language they want.

Catrya, line 51: The idea was to send messages with just the action, without a text string inside, so that clients can interpret that action and show a message to users with the text they want and in the language they want.

bilthon: Oh, sorry, I somehow missed this.

Yes, the idea is to replace custom messages with static codes that clients can map to localized, human-readable messages. However, I don’t think we have those static codes implemented yet, so for now, I decided to stick with the legacy method of using hard-coded, human-readable error messages.

Catrya: Actually there are, there are no longer any actions with text strings in the code, all the ones that had them were replaced by one of these: https://github.com/MostroP2P/mostro-core/blob/c0cd7bbd263a303e7aab46588362839e836f0ac0/src/message.rs#L37C1-L84C2

Perhaps in this case just send the cant do msg, but if is needit something more descriptive, create it as a new action
But if don't do it for now, at least keep it pending to don't forget later, because since there are no more actions with text string we could forget that was added one recently and not change it later. My 2 cents

bilthon: Oh I see. The thing is, we can have many many different reasons for errors like `CantDo`. Expanding this set for every new specific error we want to convey to clients would result in an unbounded number of actions, making `CantDo` itself too vague in the process.

Instead I'd suggest keeping the action list concise and grouping all specific error messages into a new enum, which we could call `ErrorMessage`. What do you say @grunch?

grunch: Sorry I missed this response, yes I think it makes sense @bilthon, if you want you want to do it please go first to mostro-core and then change it here

PR #402, opened 2024-12-15
testing new db approach on CI

arkanoider:
Hi @grunch , @Catrya , @bilthon 

being in office today and so a bit more in trouble to work on mostro testing, i tweaked a bit for the `mostro.db` "issue".
I mean... we also been there...time passes me or @grunch change something in sql files and the you pull latest work try to create an order...take it...and boom db error!! Why?! WTF! It's always the same ( @Catrya knows well :smile: ).

So you need to run `init_db.sh` and copy your new db in mostro folder.

I was trying to use build scritp (`build.rs`) to improve this and automatically generate at every build the correct db, but murphy's law hit me and I run into this so called deadlock issue:

https://lpalmieri.com/posts/cargo-px/#invoking-cargo

and i admit I was not able to exit from there easily.

Ok! After this what i am proposing here is to generate:

- mostro.db
- sqlx-data.json

from github actions. Not was super immediate, but I think it's working now. Some details:

- I put both files in  .gitignore so we can play locally and if we modify sql file correct files will be autogenerated from github and committed to repo.
- Result of this is that when you pull latest modification you should have both files updated, eventually the missing point is just the copy of mostro.db in mostro folder, means no more `init_db.sh` for normal user at least he wants to change locally sql files in migrations folder.
- So in the end the new `.yml` file in mostro will do all the job that `init_db` did locally but in the github world creating the file for us.

The downside is a commit more generated by the actions, not a big issue imo, but i am here to listen your opinion.

Take a look!

## Summary by CodeRabbit

- **New Features**
	- Enhanced CI workflow for Rust with integrated database setup and management.
	- Added environment variable `DATABASE_URL` for database configuration.

- **Bug Fixes**
	- Adjusted `.gitignore` to track variations of `mostro.db` while ignoring the specific file.
	- Added entry for `sqlx-data.json` in `.gitignore`.

- **Chores**
	- Updated permissions for workflow to allow commit and push actions.


grunch:
If I understand this correctly, with every new commit a new commit on mostrod repository will be generated automatically?

arkanoider:
> If I understand this correctly, with every new commit a new commit on mostrod repository will be generated automatically?

You're right!

Was not able to do it better,but now:
- We have always the latest  `mostro.db` ( and also local json file) generated for users ( devs will find another commit after their latest and this could be a downside )
- We just need eventually to remind users to overwrite their mostro.db file without using init_db 

Don't know if you like it, I see some advanteges, also if not a perfect solution. 
Still have to see the behaviour if you pull from repo and you have a mostro.db that must be updated.

grunch:
> You're right!
> 
> Was not able to do it better,but now:
> 
>     * We have always the latest  `mostro.db` ( and also local json file) generated for users ( devs will find another commit after their latest and this could be a downside )
> 
>     * We just need eventually to remind users to overwrite their mostro.db file without using init_db
> 
> 
> Don't know if you like it, I see some advanteges, also if not a perfect solution. Still have to see the behaviour if you pull from repo and you have a mostro.db that must be updated.

I think we should increment in double the commits, what about check if the file exists and if it is not we create a new one, I did this in a very primitive way on mostro-cli, but I'm sure you can do it way better than me 😃

arkanoider:
> > You're right!
> > Was not able to do it better,but now:
> > ```
> > * We have always the latest  `mostro.db` ( and also local json file) generated for users ( devs will find another commit after their latest and this could be a downside )
> > 
> > * We just need eventually to remind users to overwrite their mostro.db file without using init_db
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Don't know if you like it, I see some advanteges, also if not a perfect solution. Still have to see the behaviour if you pull from repo and you have a mostro.db that must be updated.
> 
> I think we shouldn't increment in double the commits, what about check if the file exists and if it is not we create a new one, I did this in a very primitive way on mostro-cli, but I'm sure you can do it way better than me 😃

Ok! Going to find a way to clone CLI autogeneration!

arkanoider:
I will open a different PR for the new implementation...

grunch:
> I will open a different PR for the new implementation...

ok let's close this one in favor of #411

PR #410, opened 2024-12-29
Mostro in Rust Workspace fashion

arkanoider:
Hi @grunch ,

here a MVP of workspace idea, take a look and think about, dropping here a list of advantages (imo):

- Everything in one place : `Mostro`, `Mostro-cli` and `Mostro-core` this means no more patch with local repo of mostro core while working now mostro-core is just a dep of mostro so we are still using local version.
- Add of xtask : xtask is a script idea to use all kind of commands of cargo and of a shell to manipulate things, you will see a created three basics: clean, build, builddb and buildall, you can easily understand what they do reading the code, for example now with `builddb` you will recreate db without shell but just pure rust and we can do all the same we do with shell. `Buildall` will do everything: clean, build db, and build all packages.
- I added just so we can easily and simple commands: do a `just -l` to get a list...

Ok just a starting point...

@bilthon @Catrya  let me know your opinion too...


grunch:
@arkanoider  this branch have a lot of conflicts, let's keep working this idea but please create a new branch from develop

arkanoider:
Yes. No rush...workspace is just a file organization thing, will redo it from develop branch as a feature branch.

PR #432, opened 2025-01-18
cancel in waiting-buyer-invoice or waiting-payment

### LINKED ISSUE #379: Allow a maker to cancel an order in status waiting-buyer-invoice or waiting-payment
If a user creates an order but wants to cancel it, if the order has already been taken by someone and is in status `waiting-buyer-invoice` or `waiting-payment`, the maker cannot cancel it and has to wait for the taker to cancel the order and go back to status `pending` to cancel it; but if the taker continues and does his part, then the maker cannot cancel it either, they has to wait for 15 min (or whatever time the mostro has set) to pass for the order to be cancelled because they has not done his part. IMHO this gives bad UX.

I think it would be good to allow the maker of an order to cancel it even if it is in status `waiting-buyer-invoice` or `waiting-payment`.

How it would work:
When a maker cancels the order it should go to status `canceled`.

Maker-Seller:
- When the order is taken, it goes to the status `waiting-buyer-invoice`. If the seller cancels before the buyer submits the invoice, a message must be sent to the buyer that the order was cancelled, and a message must be sent to the maker with the Cancelled action.
- If the buyer has already submitted the invoice, the order will go to the status `waiting-payment`. If the maker cancels it, the hold invoice must be cancelled, and cancellation messages must be sent to both the maker and the taker.

Maker-Buyer:
- When the order is taken, it goes to the status `waiting-payment`. If the buyer cancels before the seller pays the hold invoice, the hold invoice must be cancelled and a message canceling the order must be sent to both.
- If the order went to the status `waiting-buyer-invoice` bc the seller pays the hold invoice, if the buyer cancels, the hold invoice must be cancelled and a message canceling the order must be sent to both.

Catrya:
closes #379 

Added:

When the maker cancels an order in status `waiting-buyer-invoice` or `waiting-payment`, the order goes to status `canceled` and the taker will be notified that it was cancelled

When the taker cancels an order in the status `waiting-buyer-invoice` or `waiting-payment`, the order goes to status `pending`. The maker will be notified that it was cancelled only if the maker has already been notified to submit their invoice or pay the hold invoice

## Summary by CodeRabbit

- **New Features**
  - Enhanced order cancellation process with broader status handling.
  - Improved user role-based cancellation logic.
  - Added comprehensive notifications for order cancellations.

- **Bug Fixes**
  - Streamlined invoice cancellation handling.
  - Refined conditions for order status changes during cancellation.

### DIFF
--- a/src/app/cancel.rs
+++ b/src/app/cancel.rs
@@ -85,16 +85,103 @@ pub async fn cancel_action(
         return Ok(());
     }
 
-    if order.kind == OrderKind::Sell.to_string()
-        && order.status == Status::WaitingBuyerInvoice.to_string()
+    if order.status == Status::WaitingPayment.to_string()
+        || order.status == Status::WaitingBuyerInvoice.to_string()
     {
-        cancel_add_invoice(ln_client, &mut order, event, pool, my_keys, request_id).await?;
-    }
+        let (seller_pubkey, buyer_pubkey) = match (&order.seller_pubkey, &order.buyer_pubkey) {
+            (Some(seller), Some(buyer)) => (seller, buyer),
+            (None, _) => return Err(Error::msg("Missing seller pubkey")),
+            (_, None) => return Err(Error::msg("Missing buyer pubkey")),
+        };
 
-    if order.kind == OrderKind::Buy.to_string()
-        && order.status == Status::WaitingPayment.to_string()
-    {
-        cancel_pay_hold_invoice(ln_client, &mut order, event, pool, my_keys, request_id).await?;
+        let taker_pubkey: String;
+        if seller_pubkey == &order.creator_pubkey {
+            taker_pubkey = buyer_pubkey.to_string();
+        } else {
+            taker_pubkey = seller_pubkey.to_string();
+        }
+
+        if user_pubkey == order.creator_pubkey {
+            if let Ok(order_updated) = update_order_event(my_keys, Status::Canceled, &order).await {
+                let _ = order_updated.update(pool).await;
+            }
+
+            if let Some(hash) = &order.hash {
+                ln_client.cancel_hold_invoice(hash).await?;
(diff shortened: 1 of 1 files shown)


Catrya:
please @grunch @arkanoider take a look
I've changed the cancel code a bit, and although I've already tested it and it works well to me, remember that I'm pretty new to coding

arkanoider:
Hi @Catrya @grunch ,

ok I looked to your proposal @Catrya and if I got your point now the maker can cancel the order also in the `waiting-buyer-invoice` or `waiting-payment` states which was previously blocked by this checks in functions `cancel_add_invoice` and `cancel_pay_hold_invoice`:

**cancel_add_invoice**

```Rust
    if buyer_pubkey != &user_pubkey {
        // We create a Message
        send_cant_do_msg(request_id, Some(order.id), None, &event.rumor.pubkey).await;
        return Ok(());
    }
```

**cancel_pay_hold_invoice**

```Rust
    if seller_pubkey.to_string() != user_pubkey {
        // We create a Message
        send_cant_do_msg(request_id, Some(order.id), None, &event.rumor.pubkey).await;
        return Ok(());
    }
```

These checks blocks makers to cancel orders before taker make his step.

So looking back I completely agree with you and your modifications that looks good. @grunch do you remember if there's a good reason to block makers cancel operations in that cases?

@Catrya did you also tested it ( I suppose you're busy before getting famous in El Salvador ;) ) let me know if you fully tested your pr or you need some more confirmations!

So for me your pr is utACK.

In case we merge this, maybe I will try to split a bit the code to make it more readable, but for now we can go on like this.

Catrya:
Hi @arkanoider thanks for the review! I tested it and in all cases it worked well for me, but if you want to do more tests just in case, go ahead :)

arkanoider (approved):

arkanoider:
LGTM!

Issue #437, opened 2025-02-12
Put the user's reputation updated in the events 38383


Catrya:
When a user is rated, the reputation is updated in the mostrod database. However, if that user creates an order in the event 38383 appears, `"rating", "none"`
The user's reputation must be updated

arkanoider:
@grunch I can fix this, but we still want have it in the event? I mean, we will remove this in the end does it worth to fix it?

Catrya:
> [@grunch](https://github.com/grunch) I can fix this, but we still want have it in the event? I mean, we will remove this in the end does it worth to fix it?

Hi @arkanoider @grunch  The event I'm referring to is this one:
```
[
  "EVENT",
  "6ec901af-de9e-4ed3-970e-392eb2553b6d",
  {
    "id": "d17d8189fc83f12fa0a374df9af64b3ae9c5722236d27e3a58d60ffa31b1e1ce",
    "pubkey": "00000018c1ae3147f9010b5d768ec2b121acf1a67f238be8e2bdd7c1f9eef705",
    "created_at": 1739913346,
    "kind": 38383,
    "tags": [
      [
        "d",
        "3ea1d1a4-7e9e-409e-b6c6-8bbde2c33c21"
      ],
      [
        "k",
        "buy"
      ],
      [
        "f",
        "CUP"
      ],
      [
        "s",
        "pending"
      ],
      [
        "amt",
        "500"
      ],
      [
        "fa",
        "4"
      ],
      [
        "pm",
        "testReputacion"
      ],
      [
        "premium",
        "0"
      ],
      [
        "rating",
        "none"
      ],
      [
        "network",
        "mainnet"
      ],
      [
        "layer",
        "lightning"
      ],
      [
        "expiration",
        "1740042946"
      ],
      [
        "y",
        "mostro"
      ],
      [
        "z",
        "order"
      ]
    ],
    "content": "",
    "sig": "384e38ae93fb05ae88666f0b03cd1ad175c3d469d0186808bc45bcd5cff8573fd381de1773df0aaa44bac7d834e247ee70beb78ac7b2c3c41c0f73e3ddf937ce"
  }
]
```

The reputation in that event should not be missing because it is where users can know the reputation of whoever creates an order.

@grunch  Or is that going to be removed from there and I have confused it? Sorry

The event that I suppose should be removed in the future is the one for publishing the reputation, but in that event the reputation is being correctly set, there is nothing to fix there now:
```
[
  "EVENT",
[...]

PR #445, opened 2025-02-19
Feat: issue #437 implementation

arkanoider:
Hi @Catrya ,

i did the fix for rating of user not present in the nostr event when a new order is created. Please take a look and test when you're available.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined user reputation handling by removing the ability to fetch user ratings.
  - Updated the handling of reputation data to utilize a more versatile structure.

- **Chore**
  - Enhanced error handling for API requests and improved clarity in error reporting.
  - Adjusted imports to support new logic for user reputation retrieval.
  - Updated function signatures to provide more specific error types.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -200,11 +200,14 @@ pub async fn publish_order(
     let order_id = order.id;
     info!("New order saved Id: {}", order_id);
     // Get user reputation
-    let user = is_user_present(pool, initiator_pubkey.to_string())
+    let user = is_user_present(pool, identity_pubkey.to_string())
         .await
         .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
     // We transform the order fields to tags to use in the event
-    let tags = order_to_tags(&new_order_db, Some(user.total_rating));
+    let tags = order_to_tags(
+        &new_order_db,
+        Some((user.total_rating, user.total_reviews, user.created_at)),
+    );
     // nip33 kind with order fields as tags and order id as identifier
     let event = new_event(keys, "", order_id.to_string(), tags)
         .map_err(|e| MostroInternalErr(ServiceError::NostrError(e.to_string())))?;
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,7 +1,7 @@
-use crate::app::rate_user::get_user_reputation;
 use crate::bitcoin_price::BitcoinPriceManager;
 use crate::cli::settings::Settings;
 use crate::db;
+use crate::db::is_user_present;
 use crate::flow;
 use crate::lightning;
 use crate::lightning::invoice::is_valid_invoice;
@@ -12,8 +12,8 @@ use crate::nip33::{new_event, order_to_tags};
 use crate::MESSAGE_QUEUES;
 use crate::NOSTR_CLIENT;
 
-use anyhow::Context;
 use chrono::Duration;
+use fedimint_tonic_lnd::lnrpc::invoice::InvoiceState;
 use mostro_core::error::CantDoReason;
 use mostro_core::error::MostroError::{self, *};
 use mostro_core::error::ServiceError;
@@ -25,13 +25,11 @@ use sqlx::Pool;
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
 use sqlx_crud::Crud;
+use std::collections::HashMap;
 use std::fmt::Write;
 use std::str::FromStr;
 use std::thread;
 use tokio::sync::mpsc::channel;
-// use fedimint_tonic_lnd::Client;
-use fedimint_tonic_lnd::lnrpc::invoice::InvoiceState;
-use std::collections::HashMap;
 use tracing::info;
 use uuid::Uuid;
 
@@ -41,13 +39,15 @@ const MAX_RETRY: u16 = 4;
 pub async fn retries_yadio_request(
     req_string: &str,
     fiat_code: &str,
-) -> Result<(Option<reqwest::Response>, bool)> {
+) -> Result<(Option<reqwest::Response>, bool), MostroError> {
     // Get Fiat list and check if currency exchange is available
     let api_req_string = "https://api.yadio.io/currencies".to_string();
     let fiat_list_check = reqwest::get(api_req_string)
-        .await?
+        .await
+        .map_err(|_| MostroInternalErr(ServiceError::NoAPIResponse))?
         .json::<FiatNames>()
-        .await?
+        .await
+        .map_err(|_| MostroInternalErr(ServiceError::MalformedAPIRes))?
         .contains_key(fiat_code);
 
     // Exit with error - no currency
@@ -57,7 +57,7 @@ pub async fn retries_yadio_request(
 
     let res = reqwest::get(req_string)
         .await
-        .context("Something went wrong with API request, try again!")?;
+        .map_err(|_| MostroInternalErr(ServiceError::NoAPIResponse))?;
 
     Ok((Some(res), fiat_list_check))
 }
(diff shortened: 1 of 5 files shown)


src/util.rs
```diff
@@ -200,9 +200,11 @@ pub async fn publish_order(
     let order_id = order.id;
     info!("New order saved Id: {}", order_id);
     // Get user reputation
-    let reputation = get_user_reputation(&initiator_pubkey.to_string(), keys).await?;
+    let user = is_user_present(pool, initiator_pubkey.to_string())
+        .await
+        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
     // We transform the order fields to tags to use in the event
-    let tags = order_to_tags(&new_order_db, reputation);
```
grunch, line 205: We need to show to the counterpart 3 fields, `total_rating`, `total_reviews` and `days_operating`, we need to refactor to that

Catrya (commented):
Hi @arkanoider when i try to create a new order this appears in mostrod logs

```
2025-02-19T15:53:44.827163Z  WARN mostrod::app: Error in NewOrder with context Order id not present in database
```

src/util.rs
arkanoider:
```Json
[
    "rating",
    "4.5"
],
[
    "reviews",
    "54"
],
[
    "days_operating",
    "456"
],
```

Is it good like this @grunch ?
Days operating filed I suppose it's calculated from user creation time in mostrod users table in database right?

arkanoider:
> Hi @arkanoider when i try to create a new order this appears in mostrod logs
> 
> ```
> 2025-02-19T15:53:44.827163Z  WARN mostrod::app: Error in NewOrder with context Order id not present in database
> ```

I did a big shit! :) Will check asap...in office I was not able to test.

arkanoider:
@Catrya give another try to this please.

arkanoider:
Modified `rating` tag as JSON object:
```JSON
[
  "rating",
  "{\"total_reviews\":1,\"total_rating\":3.0,\"days\":33}"
],
```

Catrya (commented):
when a order is in pending status is ok
```        "rating",
        "[\"rating\",{\"total_reviews\":1,\"total_rating\":5.0,\"days\":0}]"
```

But when changes the status the Days are wrong:
```
        "rating",
        "[\"rating\",{\"total_reviews\":0,\"total_rating\":0.0,\"days\":20138}]"
```

Catrya:
The other strange thing I see is that the ratings are not being updated in my database, however they are displayed correctly in event 38383.
@arkanoider Could you verify on your side that they are updated in the DB please?

image

Catrya:
I was thinking that it would be good to add to the message that the Maker receives to pay the hold invoice if he is the seller, or to put the invoice  if he is the buyer, the reputation of the Taker, the number of ratings and the days he has been using a Mostro.

Do you think it would be better if I open an issue for that or that it is addressed within this same PR?
@arkanoider @grunch

arkanoider:
> when a order is in pending status is ok
> 
> ```
>         "[\"rating\",{\"total_reviews\":1,\"total_rating\":5.0,\"days\":0}]"
> ```
> 
> But when changes the status the Days are wrong:
> 
> ```
>         "rating",
>         "[\"rating\",{\"total_reviews\":0,\"total_rating\":0.0,\"days\":20138}]"
> ```

Ok got it, just need to manage the `None` case better. Now we just use the data of reputation when order is created, when there is an update of order state or when we create a child order we set the data to None option. 

Maybe we can always add rating data to all messages related to orders? What do you think?

@Catrya @grunch

arkanoider:
> > when a order is in pending status is ok
> > ```
> >         "[\"rating\",{\"total_reviews\":1,\"total_rating\":5.0,\"days\":0}]"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But when changes the status the Days are wrong:
> > ```
> >         "rating",
> >         "[\"rating\",{\"total_reviews\":0,\"total_rating\":0.0,\"days\":20138}]"
> > ```
> 
> Ok got it, just need to manage the `None` case better. Now we just use the data of reputation when order is created, when there is an update of order state or when we create a child order we set the data to None option.
> 
> Maybe we can always add rating data to all messages related to orders? What do you think?
> 
> @Catrya @grunch

Mmmmhhh...i will try to answer to myself, maybe it was correct to have only in a new order event the reputation of the maker, which makes sense because the taker can evaluate that after taking all the event update of the order could be without rating tag ( o set to null or zero ). Maybe just when we have a child new order from a range order we can publish it back for the new order, also if it will be the same as previous because users won't have rate for each other yet.

Let me know what you think guys... 
@Catrya @grunch

arkanoider:
@Catrya if you can do a round of test with latest here let me know. 

- Rating appears only in event of `new order` ( maker creation or child orders ).
- Rating is cancelled during messages of a trade, no more meaningful imo. It is useful when a taker decide if he wants to take an order.
- Downside in child orders is that user have not rated each other at that moment so rating is not updated.

Let me know your thoughts and results.

@grunch 

Removed other use of `anyhow` my objective it to completely remove dependence from cargo.toml, we are near.

Catrya:
Well, I tested it again, for me it's ok

The bug that I thought wouldn't let me try it out isn't actually that bad, the publication of child orders does work, but if the amount left is less than the initial range limit, it sends a command that it shouldn't send, but that has nothing to do with this PR

grunch:
> I was thinking that it would be good to add to the message that the Maker receives to pay the hold invoice if he is the seller, or to put the invoice if he is the buyer, the reputation of the Taker, the number of ratings and the days he has been using a Mostro.
> 
> Do you think it would be better if I open an issue for that or that it is addressed within this same PR? @arkanoider @grunch

yes please create a new issue for that, let's not include it here and keep this one small

grunch:
> > when a order is in pending status is ok
> > ```
> >         "[\"rating\",{\"total_reviews\":1,\"total_rating\":5.0,\"days\":0}]"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But when changes the status the Days are wrong:
> > ```
> >         "rating",
> >         "[\"rating\",{\"total_reviews\":0,\"total_rating\":0.0,\"days\":20138}]"
> > ```
> 
> Ok got it, just need to manage the `None` case better. Now we just use the data of reputation when order is created, when there is an update of order state or when we create a child order we set the data to None option.
> 
> Maybe we can always add rating data to all messages related to orders? What do you think?
> 
> @Catrya @grunch

I think the rating data is only needed on pending orders

arkanoider:
> > > when a order is in pending status is ok
> > > ```
> > >         "[\"rating\",{\"total_reviews\":1,\"total_rating\":5.0,\"days\":0}]"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > But when changes the status the Days are wrong:
> > > ```
> > >         "rating",
> > >         "[\"rating\",{\"total_reviews\":0,\"total_rating\":0.0,\"days\":20138}]"
> > > ```
> > 
> > Ok got it, just need to manage the `None` case better. Now we just use the data of reputation when order is created, when there is an update of order state or when we create a child order we set the data to None option.
> > 
> > Maybe we can always add rating data to all messages related to orders? What do you think?
> > 
> > @Catrya @grunch
> 
> I think the rating data is only needed on pending orders

yep! it should be like that now!

grunch:
> > > when a order is in pending status is ok
> > > ```
> > >         "[\"rating\",{\"total_reviews\":1,\"total_rating\":5.0,\"days\":0}]"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > But when changes the status the Days are wrong:
> > > ```
> > >         "rating",
> > >         "[\"rating\",{\"total_reviews\":0,\"total_rating\":0.0,\"days\":20138}]"
> > > ```
> > 
> > 
> > Ok got it, just need to manage the `None` case better. Now we just use the data of reputation when order is created, when there is an update of order state or when we create a child order we set the data to None option.
> > Maybe we can always add rating data to all messages related to orders? What do you think?
> > @Catrya @grunch
> 
> Mmmmhhh...i will try to answer to myself, maybe it was correct to have rating tagvonly in a new order event the reputation of the maker, which makes sense because the taker can see it and use it to take a decision.
> 
> After taking an order all the event update of the could be without rating tag ( or set to null or zero ), maybe just when we have a child new order from a range order we can publish it back for the new child order. The "bad" thing in this case is that rating is the same as before because users won't have sent the ratings to each other.
> 
> Let me know what you think guys... @Catrya @grunch

I agree, in the future mostro can update that order event with the new rating data, but let's not get complicated right now with that

arkanoider:
I tested a bit on my side too, seems doing what we need now. If also @Catrya @grunch agree after testing we could merge on develop.

Catrya:
tACK @arkanoider @grunch

grunch (approved):
utACK

PR #454, opened 2025-02-28
Correct full privacy mode check when orders arrives

arkanoider:
@grunch @Catrya 

fix for full privacy order coming in mostrod, now events are published also if order is in full privacy.

Removed an useless map_err call which was creating not precise error message remapping again a `MostroError` coming from below.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new function to streamline the generation of tags for orders based on user privacy status, enhancing the order processing experience.

- **Refactor**
	- Simplified error management during order publication for more consistent and reliable processing.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -163,6 +163,33 @@ pub fn get_expiration_date(expire: Option<i64>) -> i64 {
     }
     expire_date
 }
+/// Check if the order is full privacy or normal and return the tags accordingly
+pub async fn check_full_privacy_order(
+    new_order_db: &Order,
+    pool: &SqlitePool,
+    identity_pubkey: &PublicKey,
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -163,6 +163,33 @@ pub fn get_expiration_date(expire: Option<i64>) -> i64 {
     }
     expire_date
 }
+/// Check if the order is full privacy or normal and return the tags accordingly
+pub async fn check_full_privacy_order(
```
grunch, line 167: if this function is returning tags the name should be according to it

arkanoider: take a look and tell me if this is more meaningful.

grunch: awesome!

Issue #490, opened 2025-06-03
Bug when assigning child purchase orders by range


Catrya:
When a range buy order is republished, that child order is assigned by Mostrod to the Taker of the first order, not the buying Maker.
This only happens with buy orders, not sell orders.
To reproduce it:

1. A creates a purchase order with a range
2. B takes it with a quantity less than the maximum order
3. They complete the order normally
4. The child order is republished and Mostrod sends the `new-order` message to B, as if they were the order's creator.
5. A is unaware of that child order; the last message they received from Mostrod was the purchase completed message, and they rate your counterparty

arkanoider:
@Catrya fix in #491

Catrya:
fixed in #491

PR #491, opened 2025-06-05
fix(#490): child order of buy order was not sent to maker, but to taker

arkanoider:
@Catrya 

this is the fix for #490 .
Tested with a sell order and a buy order - now child order seems correctly sent to makers.

### DIFF
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -219,6 +219,9 @@ async fn handle_child_order(
             _ => {}
         }
 
+        let destination_key = child_order
+            .get_creator_pubkey()
+            .map_err(MostroInternalErr)?;
         child_order.creator_pubkey = next_trade_pubkey.to_string();
         child_order.next_trade_index = None;
         child_order.next_trade_pubkey = None;
@@ -230,7 +233,7 @@ async fn handle_child_order(
             new_order.id,
             Action::NewOrder,
             Some(Payload::Order(new_order)),
-            next_trade_pubkey,
+            destination_key,
             Some(next_trade_index as i64),
         )
         .await;


Catrya (commented):
Hi @arkanoider 
Now in purchase orders Mostrod sends the `new-order` message of the child order to the correct user, but when someone takes the child order that goes to `waiting-buyer-invoice` status, it sends the `add-invoice` message to the taker of the first order. In the database, in `creator_pubkey`, there is the trade key of the taker, not the maker (i mean the next trade key, no the key the used in 1st time)

In sales orders it still works well

arkanoider:
great...i will revert this. It's weird it did not work before, will look deeper.

Catrya (commented):
Hi @arkanoider mostro crashes when it receives the release action in a purchase range order.
```
2025-06-06T02:24:02.104833Z  INFO mostrod::util: Sending DM, Event ID: be5822df7adf89c9dca0dd339b88b66e097efeb9fb7854b4f8e1f1ff37825db4 with payload: "{\"order\":{\"version\":1,\"request_id\":null,\"trade_index\":null,\"id\":\"02f51039-bd0f-492b-989d-e0fd7695f806\",\"action\":\"released\",\"payload\":null}}"
thread 'main' panicked at src/app/release.rs:187:74:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

arkanoider:
@Catrya seems working on my side. Tested also privacy orders and seems ok. I fixed and refactored the code for better readability. Improved privacy: now child order master seller/buyer key is salted in case of normal user, previously they were the same of parent order.
As usual you have the final sentence on tests! :wink:

Catrya (approved):
tACK great job @arkanoider

PR #494, opened 2025-06-07
check privacy mode before rating users

### LINKED ISSUE #477: Remove from the logs that a user was not rated in full privacy mode
When a user is in full privacy mode and is rated (their counterpart doesn't know they are in that mode), this error appears in the logs because Mostro has no one to assign the rating to:
```
2025-04-16T15:09:55.011825Z  WARN mostrod::app: Error in RateUser with context Error in database access: Error caused by Error in database access: no rows returned by a query that expected to return at least one row
```
If someone runs a Mostro instance and sees this error, they may think something is wrong. 
Deleting that log completely prevents confusion, and it doesn't add much to having that information there.

Catrya:
fix #477 
- Check if order is in full privacy mode before rating
- Skip rating silently for privacy-protected users

@grunch @arkanoider testing claude code. I checked it and tested it and it works fine, but please check it too.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved privacy handling by preventing reputation updates for users in full privacy mode. Ratings will now be skipped for users who have enabled full privacy.

### DIFF
--- a/src/app/rate_user.rs
+++ b/src/app/rate_user.rs
@@ -9,8 +9,7 @@ use sqlx::{Pool, Sqlite};
 pub fn prepare_variables_for_vote(
     message_sender: &str,
     order: &Order,
-) -> Result<(String, String, bool, bool), MostroError> {
-    let mut counterpart: String = String::new();
+) -> Result<(String, bool, bool), MostroError> {
     let mut counterpart_trade_pubkey: String = String::new();
     let mut buyer_rating: bool = false;
     let mut seller_rating: bool = false;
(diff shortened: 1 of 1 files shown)


Catrya:
> @Catrya
> 
> tACK on my side.
> 
> I removed the now unused counterpart return value which is now derived from another function. Added a cargo doc documentation at head of rate function.
> 
> As usual you have the last sentence on testing! Do some testing on your side, for me it's a merge.

tested and it works fine @grunch

grunch (approved):
utACK

PR #503, opened 2025-07-02
Start Os wrapper helper

arkanoider:
@grunch,

very basic `Dockerfile`  that simply build mostrod and copy db inside a folder. This approach is then used by start os wrapper that will pull the mostro docker image and use it inside their wrapper for dockerization using another layer.

If you need to define more things in docker file let me know. This works and create an image.

This is one i created for testing:
https://hub.docker.com/repositories/arkanoider

Added an rpc command to get mostro version, using grpcurl tool we can ask version to mostrod, this is useful as sanity check when it will bootup from StartOs:

```Bash
$ grpcurl -plaintext -proto proto/admin.proto -d '{}' 127.0.0.1:50051 mostro.admin.v1.AdminService/GetVersion
{
  "version": "0.14.0"
}
```

I have added also a cargo feature `startos` , i will use the feature in the dockerfile to create startOs package, this flags will add:

- password flag to set from cli encryption password of db
- cleartext flag to avoid password encryption of db

we need this for the launch of mostrod in startos to avoid interactive part of CLI with users for password db encryption.

## Summary by CodeRabbit

* **New Features**
  * Added RPC endpoints for version retrieval and database password validation.
  * Introduced CLI options for database password management, including a cleartext flag (available with the "startos" feature).
  * Added example clients to test version retrieval and database password validation RPCs.

* **Improvements**
  * Enhanced Docker workflows and CI to build, tag, and publish StartOS images for multiple platforms.
  * Added a StartOS Docker image/service and runtime instructions in documentation.
  * Strengthened database password handling and setup flow.

* **Bug Fixes**
  * Improved error handling during database password setup.

* **Chores**
  * Removed an obsolete Docker startup script.
  * Updated test invoice data.


grunch (changes requested):
This looks amazing, what about the nitpick comments from the rabbit ? I think those about the dockerfile are legit but not sure the other one

instructions.md
```diff
@@ -0,0 +1,22 @@
+# Mostro Instructions
```
grunch: I think we should follow that `standard` of naming md files in uppercase

Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Build stage
```
grunch: we have a docker/ directory with also a Dockerfile and other files, is that deprecated right? shouldn't be deleted from the repository in this same PR?

instructions.md
arkanoider: yep! will.fix

Dockerfile
arkanoider: yes this is also an open point to me, old dockerfiles worked for spin up mostrod or a local relay, maybe this is still useful?
No docker.compose support for startOs so juat dockerfile is used

grunch: we should have a way of users can create a mostrod node with docker but avoiding to have two dockerfiles, we don't want to confuse users, so is up to you to decide how to do this

arkanoider: yesss....i don't.think.we need relay dockerized...i will proceed removing old.files.
Uaers can choose relays from Startos ui and in case setup their ones. Will proceed.like that if it's.ok also for you.

arkanoider: > we should have a way of users can create a mostrod node with docker but avoiding to have two dockerfiles, we don't want to confuse users, so is up to you to decide how to do this

Yep! I think i can add to it docker folder and let user make it with something like make `mostro-startos`, will try this conservative approach, so we can keep :

- Mostro dockerized locally
- Relay dockerized locally
- Mostro packaged for startos

arkanoider:
Added new ideas for building and pushing to `docker hub` startos images. Now they are pushed to my arkanoider docker hub, we can clearly change it. Let me test all @grunch and in case we can merge if you like this solution with a conservative approach, without losing @tincho70  nice work!

Makefile
```diff
@@ -43,3 +43,44 @@ docker-down:
...
+uninstall:
+	start-cli package uninstall $(ID)
+
+.PHONY: logs
+logs:
+	start-cli package logs $(ID)
+
+.PHONY: clean
+clean:
+	rm -f $(ID).s9pk
```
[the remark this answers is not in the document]
arkanoider: yep! wlll.fix this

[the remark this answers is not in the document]
arkanoider: no i will.find a way to read directly package name and version from cargo.toml, maybe cargo manifest? any advice @coderabbitai?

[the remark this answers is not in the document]
arkanoider: the binary name is mostrod and the package name is mostro. I want to create package of docker with name mostro. Can you prepare a commit for me with grep/sed solution? @coderabbitai

PR #504, opened 2025-07-04
fix: ensure maker always gets notified when their order is taken

Catrya:
Currently when a buy order is taken, the maker receives a `waiting seller to pay` message, so they know how their order is progressing. However, when a sell order is taken, its maker isn't notified that it's in `waiting buyer invoice` state, creating a period where that maker does not see their order as Pending in clients but also doesn't know what's happening with their order, which can confuse new users.

This PR ensures that when a sell order is taken, mostrod sends a `waiting buyer invoice` message to the sell order maker.

## Summary by CodeRabbit

* **New Features**
  * Added a new notification to inform sellers when their order is taken and the buyer needs to add an invoice.

* **Documentation**
  * Improved and clarified comments for better understanding of notification processes.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -640,7 +640,8 @@ pub async fn show_hold_invoice(
         order.trade_index_seller,
     )
     .await;
-    // We send a message to buyer to know that seller was requested to pay the invoice
+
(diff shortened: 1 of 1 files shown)


arkanoider:
utACK

PR #507, opened 2025-07-10
Fix: Set event expiration time to 24 hours, improved time to check or…

### LINKED ISSUE #506: Inconsistencies in event timing and status changes
- Pending order events don't expire after 24 hours, but after 36 hours, even though the mostro info event says it's after 24 hours. 
` [ "expiration_hours",   "24" ]`
This happens because 12 hours are being added to the expected duration. 
src/nip33
```
Tag::custom(
TagKind::Custom(Cow::Borrowed("expiration")),
vec![(order.expires_at + Duration::hours(12).num_seconds()).to_string()],
),
```

- When orders are taken, if the taker doesn't respond, they return to pending after approximately 21 minutes. I looked at two orders, and one took 21 minutes and the other 22 minutes and 17 seconds. It should return to pending after 15 minutes.

- If the maker doesn't respond, it's taking approximately 27 minutes (27 minutes and 7 seconds) to cancel the order. It should be canceled after 15 minutes.
```
["hold_invoice_expiration_window", "900"],
["invoice_expiration_window", "900"],
```

Catrya:
…ders to be canceled

fix #506 

- Now the kind 38383 event of a pending order has a duration of 24 hours, not 36.
- Now `job_cancel_orders()` is executed every 1 minute, so that orders do not accumulate for 15 minutes to be processed all at once, and there is greater consistency in the time of an order in status `waiting-buyer-invoice` or `waiting-payment`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted order status handling to treat expired orders as canceled.
  * Modified the order cancellation job to run every 60 seconds for consistent timing.

### DIFF
--- a/src/nip33.rs
+++ b/src/nip33.rs
@@ -94,7 +94,7 @@ fn create_status_tags(order: &Order) -> Result<(bool, Status), MostroError> {
     match status {
         Status::WaitingBuyerInvoice => Ok((order.is_sell_order().is_ok(), Status::InProgress)),
         Status::WaitingPayment => Ok((order.is_buy_order().is_ok(), Status::InProgress)),
-        Status::Canceled | Status::CanceledByAdmin | Status::CooperativelyCanceled => {
+        Status::Canceled | Status::CanceledByAdmin | Status::CooperativelyCanceled | Status::Expired=> {
             Ok((true, Status::Canceled))
         }
         Status::Success | Status::CompletedByAdmin => Ok((true, status)),
@@ -162,7 +162,7 @@ pub fn order_to_tags(
             ),
             Tag::custom(
                 TagKind::Custom(Cow::Borrowed("expiration")),
-                vec![order.expires_at.to_string()],
+                vec![(order.expires_at + Duration::hours(12).num_seconds()).to_string()],
             ),
             Tag::custom(
                 TagKind::Custom(Cow::Borrowed("y")),
--- a/src/nip33.rs
+++ b/src/nip33.rs
@@ -1,6 +1,7 @@
 use crate::config::settings::Settings;
 use crate::lightning::LnStatus;
 use crate::LN_STATUS;
+use chrono::Duration;
 use mostro_core::prelude::*;
 use nostr::event::builder::Error;
 use nostr_sdk::prelude::*;
--- a/src/nip33.rs
+++ b/src/nip33.rs
@@ -1,7 +1,6 @@
 use crate::config::settings::Settings;
 use crate::lightning::LnStatus;
 use crate::LN_STATUS;
-use chrono::Duration;
 use mostro_core::prelude::*;
 use nostr::event::builder::Error;
 use nostr_sdk::prelude::*;
@@ -163,7 +162,7 @@ pub fn order_to_tags(
             ),
             Tag::custom(
                 TagKind::Custom(Cow::Borrowed("expiration")),
-                vec![(order.expires_at + Duration::hours(12).num_seconds()).to_string()],
+                vec![order.expires_at.to_string()],
             ),
             Tag::custom(
                 TagKind::Custom(Cow::Borrowed("y")),
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -372,7 +372,7 @@ async fn job_cancel_orders() {
                     next_tick.format("%a %b %e %T %Y")
                 );
             }
-            tokio::time::sleep(tokio::time::Duration::from_secs(exp_seconds as u64)).await;
+            tokio::time::sleep(tokio::time::Duration::from_secs(60)).await;
         }
     });
 }


src/scheduler.rs
```diff
@@ -372,7 +372,7 @@ async fn job_cancel_orders() {
                     next_tick.format("%a %b %e %T %Y")
                 );
             }
-            tokio::time::sleep(tokio::time::Duration::from_secs(exp_seconds as u64)).await;
```
arkanoider, line 375: I don't think this is good @Catrya because now you are hardcoding the value of 60 seconds while now it's configurable from settings.toml file here:
```Toml
# Expiration of pending orders
expiration_seconds = 900
```

So you can just set 60 as value in settings.toml and you have the same result, but with more flexibility for admins.

src/nip33.rs
```diff
@@ -163,7 +162,7 @@ pub fn order_to_tags(
             ),
             Tag::custom(
                 TagKind::Custom(Cow::Borrowed("expiration")),
-                vec![(order.expires_at + Duration::hours(12).num_seconds()).to_string()],
```
arkanoider: Ok...i think that also this is not good, but I understand what you saw @Catrya and it's more tricky... TL;DR this expiration time is different from the scheduler job that checks pending.
This function in `scheduler.rs`:
https://github.com/catrya/mostro/blob/d142d31e4e3c5eee4c6bd01375e328901137cf55/src/scheduler.rs#L380
checks every 60 seconds ( hardcoded ) if an order is pending and is alive from more than 24 hours, after that mostro marks its state in the database to `expired`. So orders are still published from relays, but their state is expired, client have the responsability to manage the new state and be careful because in some case of relay issues or downtime you could have both expired version of message and also pending.
Think that you have 3 relays to which mostrod sends messages but at the moment mostrod send the message with new expired status one relay is down...it will still sends out the pending state of the order.
Now comes the timer you cancelled...it's a second level of protection to avoid these cases....

What you removed is related to this:
https://nips.nostr.com/40

which is related to a tag of nostr protocol that avoid to an event to be published from relays, we added this as a good practise to avoid tons of expired orders floating around in nostr and also in case of some mostro issue we are more safe that older orders are no more propagated at nostr protocol level.
So...to recap...we add that extra 12 hours of expired state order and then these events are no more republished from relays at protocol level, as described by NIP40.

What we can discuss if 12 hours is a good number or to reduce it...

Let's ask also @grunch opinion about...

arkanoider (commented):
Hi @Catrya !

i commented your code modifications, and I hope my explanations are clear, TL;DR I don't think we need to modify these things, we can discuss in case if 12 hours of extra time after order is expired in mostro database.

@grunch opinion is more the welcome!

src/scheduler.rs
Catrya: > I don't think this is good @Catrya because now you are hardcoding the value of 60 seconds while now it's configurable from settings.toml file here:
> 
> ```toml
> # Expiration of pending orders
> expiration_seconds = 900
> ```
> 
> So you can just set 60 as value in settings.toml and you have the same result, but with more flexibility for admins.

Hi @arkanoider, I disagree, these are actually two different things:

**`expiration_seconds = 900`** (from settings.toml):
- Defines when an order becomes eligible for cancellation (after 15 minutes stuck)
- Used in the database query: `taken_at < now - expiration_seconds`
- This stays configurable as intended

**`tokio::time::sleep(60)`** (scheduler frequency):
- Defines how often we check for stuck orders (every 1 minute)
- Before it was `exp_seconds` (900), so we only checked every 15 minutes

The issue with using `exp_seconds` for sleep was:
- Orders would pile up for 15 minutes before getting processed
- A user stuck at minute 1 would have to wait 14 more minutes for no reason. In the worst case, they could see their order in waiting-buyer-invoice or waiting-payment status for up to 29 minutes instead of the expected 15. This also creates inconsistencies since orders would get canceled or return to pending at completely different times depending on when exactly they got stuck in the 15-minute window.
- Database load would spike every 15 minutes instead of being spread out every minute

What I'm proposing:
- Check every minute for orders stuck longer than 15 minutes (`expiration_seconds`)
- Better UX: orders get unstuck within ~1 minute instead of waiting up to 15 minutes
- Better performance: load gets distributed instead of those periodic spikes

The `expiration_seconds` setting stays fully configurable for admins to set the timeout threshold.

src/nip33.rs
Catrya: > Ok...i think that also this is not good, but I understand what you saw @Catrya and it's more tricky... TL;DR this expiration time is different from the scheduler job that checks pending. This function in `scheduler.rs`: https://github.com/catrya/mostro/blob/d142d31e4e3c5eee4c6bd01375e328901137cf55/src/scheduler.rs#L380 checks every 60 seconds ( hardcoded ) if an order is pending and is alive from more than 24 hours, after that mostro marks its state in the database to `expired`. So orders are still published from relays, but their state is expired, client have the responsability to manage the new state and be careful because in some case of relay issues or downtime you could have both expired version of message and also pending. Think that you have 3 relays to which mostrod sends messages but at the moment mostrod send the message with new expired status one relay is down...it will still sends out the pending state of the order. Now comes the timer you cancelled...it's a second level of protection to avoid these cases....
> 
> What you removed is related to this: https://nips.nostr.com/40
> 
> which is related to a tag of nostr protocol that avoid to an event to be published from relays, we added this as a good practise to avoid tons of expired orders floating around in nostr and also in case of some mostro issue we are more safe that older orders are no more propagated at nostr protocol level. So...to recap...we add that extra 12 hours of expired state order and then these events are no more republished from relays at protocol level, as described by NIP40.
> 
> What we can discuss if 12 hours is a good number or to reduce it...
> 
> Let's ask also @grunch opinion about...

I'll read NIP-40 to understand it better, but either way we need to change something because the way it is now, orders stay showing as pending for 12 extra hours, and clients display them. Plus if a user tries to take one of these orders, Mostro doesn't send a message saying they're expired , instead it allows them to take it and continue with the order process.

In the case of lnp2pbot, the event expires at 23 hours, so when orders get marked as expired in the database, relays stop showing them and users can not take it.

src/scheduler.rs
arkanoider: Sorry for looking to wrong function! You're fuckin right, I mean we can lower granularity of check to 1 minute.

src/nip33.rs
arkanoider: Looking in to this we found a regression bug, it was introduced when aligning order status to nip69 which has these order status in docs:
`< Status >: pending, canceled, in-progress, success.`

We use these values in a function that prepares status tag for a new event, but if status is not in the ones below event is not prepared and sent:

```Rust
fn create_status_tags(order: &Order) -> Result<(bool, Status), MostroError> {
    // Check if the order is pending/in-progress/success/canceled
    let status = order.get_order_status().map_err(MostroInternalErr)?;

    match status {
        Status::WaitingBuyerInvoice => Ok((order.is_sell_order().is_ok(), Status::InProgress)),
        Status::WaitingPayment => Ok((order.is_buy_order().is_ok(), Status::InProgress)),
        Status::Canceled | Status::CanceledByAdmin | Status::CooperativelyCanceled => {
            Ok((true, Status::Canceled))
        }
        Status::Success | Status::CompletedByAdmin => Ok((true, status)),
        Status::Pending => Ok((true, status)),
        _ => Ok((false, status)),
    }
}
```
So what happens is that having status expired now the event is no more published, fix should be easy like:
```Rust
fn create_status_tags(order: &Order) -> Result<(bool, Status), MostroError> {
    // Check if the order is pending/in-progress/success/canceled
    let status = order.get_order_status().map_err(MostroInternalErr)?;

    match status {
        Status::WaitingBuyerInvoice => Ok((order.is_sell_order().is_ok(), Status::InProgress)),
        Status::WaitingPayment => Ok((order.is_buy_order().is_ok(), Status::InProgress)),
        Status::Canceled | Status::CanceledByAdmin | Status::CooperativelyCanceled => {
            Ok((true, Status::Canceled))
        }
        Status::Success | Status::CompletedByAdmin => Ok((true, status)),
        Status::Pending => Ok((true, status)),

        Status::Expired => Ok((true, status)),

        _ => Ok((false, status)),
    }
}
```

or we can make it fall in canceled state like this:
```Rust
fn create_status_tags(order: &Order) -> Result<(bool, Status), MostroError> {
    // Check if the order is pending/in-progress/success/canceled
    let status = order.get_order_status().map_err(MostroInternalErr)?;

    match status {
        Status::WaitingBuyerInvoice => Ok((order.is_sell_order().is_ok(), Status::InProgress)),
        Status::WaitingPayment => Ok((order.is_buy_order().is_ok(), Status::InProgress)),

        Status::Canceled | Status::CanceledByAdmin | Status::CooperativelyCanceled | Status::Expired => {
            Ok((true, Status::Canceled))
        }
        
        Status::Success | Status::CompletedByAdmin => Ok((true, status)),
        Status::Pending => Ok((true, status)),
        _ => Ok((false, status)),
    }
}
```

What's your opinion guys @Catrya @grunch about the event status in that case? It's going to become `canceled` or  adding an expired state to nip69?

After that things will come back as before, so after N hours (24 by default) order is marked to expired and event is published, and after other 12 hours ( we can also lower this as @Catrya is requesting ) it's also discarded at protocol level by Nostr.

Catrya: > What's your opinion guys @Catrya @grunch about the event status in that case? It's going to become canceled or adding an expired state to nip69?

I think it's preferable to use canceled, rather than modifying the nip69 just for a status that perhaps other p2p don't use.

> and after other 12 hours ( we can also lower this as @Catrya is requesting ) it's also discarded at protocol level by Nostr.

I only wanted to remove the extra 12 hours for the pending event, but if the status change event is going to be published, then I think it's fine to leave it (and even increase it in the future).   
I noticed that all events for the same order keep the same expiration time. For example, if an order has been in pending status for 23 hours and then it's taken, there will only be 13 hours left before the events expire. If the trade takes more than 13 hours (unlikely, but possible), Mostro might try to publish a success event with a created_at timestamp after the expiration.

arkanoider: > I think it's preferable to use canceled, rather than modifying the nip69 just for a status that perhaps other p2p users don't use.

I agree! We can ask @grunch opinion and fix it.

arkanoider: > I noticed that all events for the same order keep the same expiration time. For example, if an order has been in pending status for 23 hours and then it's taken, there will only be 13 hours left before the events expire. If the trade takes more than 13 hours (unlikely, but possible), Mostro might try to publish a success event with a created_at timestamp after the expiration.

Good point! I think that user waiting 12 hours could be typically a dispute case. But ok in case we have surely options to improve this:
- when order change state update `expires_at` field with actual time + 12h
- just add more time to nip40 expiration 
- sending a notifications to user some time before order expires from nostr with nip40.

Tomorrow I will retest expired order state making it collapse in `Canceled` ( in case just do on your side like my second snippet ) and if you want to accelerate expiration you can locally tweak this line:

https://github.com/MostroP2P/mostro/blob/dbd8912bb3a4b17154f3c3583656cda19f514203/src/util.rs#L185

just change Duration::hours to Durations::minute so you can just set 1 minute to make an order expire.

I think that you can modify this pr and then it's mergeable.

arkanoider:
Ok did a little test with this snippet for fixing:
```Rust
fn create_status_tags(order: &Order) -> Result<(bool, Status), MostroError> {
    // Check if the order is pending/in-progress/success/canceled
    let status = order.get_order_status().map_err(MostroInternalErr)?;

    match status {
        Status::WaitingBuyerInvoice => Ok((order.is_sell_order().is_ok(), Status::InProgress)),
        Status::WaitingPayment => Ok((order.is_buy_order().is_ok(), Status::InProgress)),

        Status::Canceled | Status::CanceledByAdmin | Status::CooperativelyCanceled | Status::Expired => {
            Ok((true, Status::Canceled))
        }
        
        Status::Success | Status::CompletedByAdmin => Ok((true, status)),
        Status::Pending => Ok((true, status)),
        _ => Ok((false, status)),
    }
}
```

Confirm on my side that now after expire time order is publishe with `canceled` state correctly!

arkanoider (approved):

src/nip33.rs
grunch: `canceled` is useful for this case, let's use that

grunch:
> Doing a review of which apps, relays implement [nip-40](https://github.com/nostrability/nostrability/issues/211). Are y'all pursuing nip-40 expiration timestamp?

@alltheseas  yeah, we implemented many NIPs in mostro 😃

Issue #526, opened 2025-10-01
Dispute initiator incorrectly identified in admin view


Catrya:
If it's a purchase order, it shows the solver that the buyer initiated the dispute, but if it's a sale order it shows that it was the seller

Sale Order: The buyer initiated the dispute but the admin sees the seller's pubkey as initiator
```
    initiator_pubkey: "c2a2ba9c04d2e33bb4666a17cc53bdb67f276324481e5011b5aae59554c36f8e",
    buyer_pubkey: Some(
        "b6fe1bc9c813c3b846dc602e1b476790983d6fa6568783c623a8affa2f42696d",
    ),
    seller_pubkey: Some(
        "c2a2ba9c04d2e33bb4666a17cc53bdb67f276324481e5011b5aae59554c36f8e",
```

Purchase Order: The seller opened the dispute, but the admin sees that the buyer is shown as the initiator
```
    initiator_pubkey: "55d8185c98b9b9c762dd23612e6f702e67a548923c784c5097dacfdf4cbac0f4",
    buyer_pubkey: Some(
        "55d8185c98b9b9c762dd23612e6f702e67a548923c784c5097dacfdf4cbac0f4",
    ),
    seller_pubkey: Some(
        "3ad583fe7ad5b4b307d4b5e5a590f5dd34f66681415901e0c673ea54fbf87523",
```

arkanoider:
Fixed via #527

PR #527, opened 2025-10-01
Fix for #526

arkanoider:
@grunch @Catrya ,

this is the fix for #526 , now initiator of dispute is correctly sent to admin( solver ) when the dispute is taken, previously, my bad, I was sending the pubkey or order creator.


Catrya (approved):
tACK

PR #552, opened 2025-12-10
  fix: correct amount calculations in seller/buyer messages

### LINKED ISSUE #551: Mostrod is not sending the correct amount of Satss in the payload for an order.
When a seller receives the `pay-invoice` message, even though they receive the hold invoice correctly, the amount they are told to pay is incorrect because it doesn't include the order fee.

Then, when the status changes to `Active`, both the seller and the buyer receive incorrect amounts, as they only receive the `order amount` without adding or subtracting the fee, as applicable.

This doesn't affect the functionality of Mostrod, but it does affect the user experience, as it can be confusing for them.

Catrya:
fix #551 
  - pay-invoice: Show order amount + fee
  - buyer-took-order: Show order amount + fee
  - hold-invoice-payment-accepted: Show order amount - fee

## Summary by CodeRabbit

* **Bug Fixes**
  * Invoice and order messages now use adjusted amounts so sellers are credited including transaction fees and buyers are charged net of fees; hold invoices and confirmation messages reflect these corrected amounts.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/app/add_invoice.rs
+++ b/src/app/add_invoice.rs
@@ -86,7 +86,7 @@ pub async fn add_invoice_action(
 
         // We send a confirmation message to seller
         let mut seller_order = SmallOrder::from(active_order.clone());
-        seller_order.amount = active_order.amount + active_order.fee;
+        seller_order.amount = active_order.amount.saturating_add(active_order.fee);
         enqueue_order_msg(
             None,
             Some(active_order.id),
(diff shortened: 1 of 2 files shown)


grunch (approved):
LGTM

arkanoider (approved):
Good job @Catrya ! Another detail that was silently there since long time, maybe i will look in the next days if I am able to compact in some way the two buffer variables to send messages to seller and buyer, but it's totally right!

uTACK!!

@grunch can merge this!

PR #599, opened 2026-02-17
Fix restore session dispute initiator

### LINKED ISSUE #598: The restore-session response does not include who initiated each dispute
This was already solved for dispute Nostr events in #563, but the restore payload was not updated alongside it.

Without this field, clients must fetch a past dispute event from a relay to determine the initiator, which is unreliable and not guaranteed to succeed. The restore response should be the source of truth for session recovery because if the relay does not have the event cached, the client has no way to determine the initiator at all.

## Current behavior

Disputes returned in the restore payload look like this:

```json
"disputes": [
  {
    "dispute_id": "<Dispute Id>",
    "order_id": "<Order Id>",
    "trade_index": 4,
    "status": "initiated"
  }
]
```

## New behavior

The `initiator` field should be included:

```json
"disputes": [
  {
    "dispute_id": "<Dispute Id>",
    "order_id": "<Order Id>",
    "trade_index": 4,
    "status": "initiated",
    "initiator": "seller"
  }
]
```

## Notes

- No schema migration is required. The data already exists in `orders.buyer_dispute` and `orders.seller_dispute` columns and can be derived via a JOIN in the restore query.

Related: #548, #563 and [protocol](https://mostro.network/protocol/restore_session.html )

BraCR10:
**Depends on: https://github.com/MostroP2P/mostro-core/issues/129**

The `restore-session` response was missing the `initiator` field in disputes. The data already existed in `orders.buyer_dispute` and `orders.seller_dispute` , so no schema changes were needed.

**Proposed changes:**

- Added `initiator: Option<String>` to `RestoredDisputesInfo` in `mostro-core`
- Added `buyer_dispute: bool` and `seller_dispute: bool` to `RestoredDisputeHelper` so those values can be read from the existing JOIN and used to derive the initiator after decryption
- Both existing queries (encrypted and non-encrypted) were updated, the JOIN with `orders` was already in place
- For non-encrypted databases: `initiator` is derived directly in SQL via a `CASE` on `buyer_dispute` / `seller_dispute`
- For encrypted databases: the same columns are fetched in the existing JOIN and `initiator` is derived in Rust after decrypting the master keys, following the current behaviour as #563 does

Closes #598 

## Summary by CodeRabbit

* **New Features**
  * Disputes now record and display the initiator (buyer, seller, or unknown), improving visibility into who opened a dispute.
  * Restored and listed dispute records include additional flags used to more accurately determine and populate the initiator when retrieving or restoring dispute data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ reqwest = { version = "0.12.1", default-features = false, features = [
   "json",
   "rustls-tls",
 ] }
-mostro-core = { version = "0.7.0", features = ["sqlx"] }
+mostro-core = { version = "0.7.1", features = ["sqlx"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 clap = { version = "4.5.45", features = ["derive"] }


mostronator (approved):
Re-reviewed after 3 new commits (`1095e698`, `dad0bbe6`, `69087484`).

Good improvements:
- Using `DisputeInitiator` enum instead of raw strings — type-safe and consistent with mostro-core
- Explicit handling of `(true, true)` conflict with `tracing::warn!` — better than silently returning `None`
- mostro-core bumped to 0.7.1

CI passes. LGTM ✅

grunch (approved):
LGTM

PR #625, opened 2026-03-05
feat: add days field to user rating event (kind 38384)

### LINKED ISSUE #624: Add days field to user rating event (kind 38384)
The user rating event (kind 38384) currently includes total_reviews, total_rating, last_rating, max_rate, and min_rate, but it does not include the days field (days since first trade).

The days field is already calculated and included in the rating tag within order events (kind 38383), but it's missing from the standalone user rating event.

**Use Case**

Userswants to see their own reputation in their account/profile screen. Currently, users can only see their reputation when they create an order (because the rating tag is included in the order event).

By adding days to the kind 38384 event, clients can query this public event and display the complete user reputation without needing to create an order or request it from Mostro.

**Proposed Solution**

Add the days tag to the rating event in src/app/rate_user.rs, calculated from the user's created_at timestamp:
```
let days = (Timestamp::now().as_u64() - user.created_at as u64) / 86400;
tags = tags.push(Tag::custom(
    TagKind::Custom("days".into()), 
    vec![days.to_string()]
));
```
This change only affects mostrod — no changes needed in mostro-core.

Expected Event Structure
```
{
  "kind": 38384,
  "tags": [
    ["d", "<user_trade_pubkey>"],
    ["total_reviews", "10"],
    ["total_rating", "48"],
    ["last_rating", "5"],
    ["max_rate", "5"],
    ["min_rate", "4"],
    ["days", "120"],  // ← new field
    ["y", "mostro"],
    ["z", "rating"]
  ]
}
```

Catrya:
fix #624 
Users should be able to see their reputation on the client they use. By adding it to the standalone rating event (kind 38384), clients can query it directly.

The mobile client already expects this field (`lib/data/models/rating.dart`) but always receives 0 because mostrod  wasn't sending it.

The change was made only in mostrod (not in the `Rating` struct of mostro-core) for these reasons:

  - `days` is not a rating attribute, it's derived from the user's `created_at` timestamp. Conceptually it doesn't  belong in the `Rating` struct alongside `total_reviews`, `last_rating`, etc.
  - The mobile client doesn't use mostro-core, it's a Flutter/Dart project with its own models, so modifying mostro-core wouldn't benefit any client

## Summary by CodeRabbit

* **Improvements**
  * Reputation calculations now include account age (days since creation) as an explicit rating tag, refining how reliability and experience are reflected in scores.
* **Tests**
  * Added unit tests validating account-age calculations across normal, zero, negative, and partial-day scenarios.
* **Documentation**
  * Updated rating event example to show the new "days" tag in event payloads.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -55,7 +55,7 @@ jobs:
             file_args="$file_args --file $f"
           done
           echo "Testing mutants in: $changed_rs"
-          cargo mutants $file_args --output mutants.out
+          cargo mutants $file_args
 
       - name: Upload mutation report
         uses: actions/upload-artifact@v4
@@ -180,7 +180,7 @@ jobs:
         run: cargo install cargo-mutants
 
       - name: Run full mutation testing
-        run: cargo mutants --output mutants.out
+        run: cargo mutants
         # Note: Do NOT fail on low score initially (report only mode)
         continue-on-error: true
 


arkanoider:
> @grunch @arkanoider The failing mutant (replace update_user_reputation_action -> Ok(())) is pre-existing and unrelated to this PR. It survives because no test executes this async function directly, it requires a SQLite pool, Nostr keys, and an UnwrappedGift, meaning full DB and protocol mocks to test end-to-end.
> 
> All 7 other mutants (including the 5 from the new days calculation) were caught. This one only appears because cargo-mutants analyzes touched lines and the function was modified when adding the calculate_days_since_creation call.
> 
> Covering it would require integration tests with full mocks, which is out of scope for this change.

uhhh ok, if mutants are merged we have to prepare basically always to these errors, I will take a look to fix the mutants for the files involved this morning. In case i will pr your pr...

PR #692, opened 2026-04-02
fix: prevent stale Order writes from causing duplicate payments

Catrya:
fix #691 
When buyer payment fails, `check_failure_retries` and `payment_success` perform full-row `order.clone().update(pool)` using a stale `Order` struct captured minutes earlier by `tokio::spawn`. This overwrites fields modified by concurrent processes, causing duplicate dev fee payments and duplicate buyer payments.                                                    
                                                                                                                      
  ## Root cause                                                                                                       
                                                            
  `do_payment` spawns async tasks that hold an `Order` copy. While the task waits for LND (can take minutes), other processes modify the same order in the DB, the dev fee cycle pays and writes `dev_fee_paid=true`, or another task succeeds and writes `status=success`. When the stale task finally calls `check_failure_retries`, the full-row write reverts these changes:                                                                               
                                                                                                                      
  - `dev_fee_paid` reverts to `false` → dev fee cycle pays again (duplicate)                                          
  - `status` reverts from `success` to `settled-hold-invoice` → buyer gets asked for a new invoice and gets paid again (duplicate)

  ## Changes                                                                                                          
                                                            
  - **`check_failure_retries`**: replace full-row write with targeted SQL that only updates `failed_payment` and `payment_attempts` — no longer overwrites `status`, `dev_fee_paid`, `dev_fee_payment_hash`, or any other field                                                        
  - **`payment_success`**: replace full-row write with targeted SQL that only updates `status` and `event_id`, with `WHERE status = 'settled-hold-invoice'` guard to prevent double success transitions from concurrent tasks                                                          
  - **`release_action` and `admin_settle_action`**: persist `settled-hold-invoice` status to DB before calling `do_payment` previously this status change only existed in memory and was persisted as a side-effect of the full-row writes  

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Abort processing when a concurrent order status change is detected to prevent duplicate notifications, disputes handling, or payments.
  * Persist status/event markers earlier to reduce race conditions between status checks, dispute handling, messaging, and payment steps.

* **Performance**
  * Reduce unnecessary writes by updating only the specific fields that changed (dev fees, payment attempts, status/event), lowering contention and overwrite risk.

* **Tests**
  * Added a test ensuring targeted dev-fee updates do not overwrite concurrent status changes.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #693, opened 2026-04-02
fix: use targeted updates for dev_fee writes

arkanoider:
Hi @Catrya ,

this could be a hardening to your #692 , using targeted update also in `dev_fee.rs` now we have completely decoupled updates of the payments.

- release, settle and all other logic just touch their part of order fields.
- dev_fee touch just the parts of dev fees.

Added also a test for regression.

Take a look!


Catrya (approved):
tACK

PR #701, opened 2026-04-09
fix: reject orders with invalid fiat_amount

Catrya:
- Reject orders with fiat_amount < 0 in all cases
  - Reject orders with fiat_amount == 0 for non-range orders
  - Add tests for both zero and negative fiat_amount scenarios

## Summary by CodeRabbit

* **Bug Fixes**
  * Orders with negative fiat amounts are now rejected.
  * Orders with zero fiat amounts are rejected for standard (non-range) orders while range orders remain supported.
* **Tests**
  * Added tests verifying rejection of non-range zero-amount orders and negative-amount range orders.

### DIFF
--- a/src/app/order.rs
+++ b/src/app/order.rs
@@ -345,7 +345,10 @@ mod tests {
         );
 
         let result = order_action(&ctx, msg, &event, &keys).await;
-        assert!(result.is_err());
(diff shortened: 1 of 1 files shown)


grunch:
duplicated #700

PR #704, opened 2026-04-10
refactor: load Nostr private key from file instead of config

gotcha:
It is more secure to put secrets in their own files, in this case, the nostr private key.

- Change nsec_privkey to nsec_privkey_file in settings
- Add async load_nostr_private_key() helper function that loads private key from file
- Make get_keys() async and use load_nostr_private_key()
- Update all get_keys() call sites to await
- Add default_pubkey parameter to order_to_tags() to avoid making it async
- Add vendor to .gitignore

## Summary by CodeRabbit

* **New Features**
  * Setup wizard now saves the Nostr private key to a secure file and reports the saved path.

* **Chores**
  * Configuration switched to reference a private-key file path instead of embedding the key.
  * Key loading and event handling moved to async for improved reliability.
  * Added .gitignore entry to exclude vendor files.

* **Documentation**
  * Docs updated with instructions for the file-based Nostr key.

* **Tests**
  * Added validation tests for the new key-file requirements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 mostro.db*
 mostro.log
 lnurl-test-server/target
+vendor
 
 # IDE's
 .idea
--- a/README.md
+++ b/README.md
@@ -492,8 +492,8 @@ payment_retries_interval = 60  # seconds between retries
 #### Nostr Configuration
 ```toml
 [nostr]
-# Your Mostro daemon's private key (nsec format)
-nsec_privkey = 'nsec1...'
+# Path to a file containing your Mostro daemon's private key (nsec format)
+nsec_privkey_file = '/home/user/.mostro/nostr-key.nsec'
 
 # Relays to connect to
 relays = [
@@ -508,6 +508,10 @@ relays = [
 # Using rana (https://github.com/grunch/rana)
 rana --vanity mostro
 ```
+
+Save the generated `nsec` in the file referenced by `nsec_privkey_file`.
+Mostro expects that file to contain only the `nsec` value.
+Inline `nostr.nsec_privkey` is no longer supported and startup will fail until you move the key into a file.
 
 **Important**: Never reuse keys between Mostro instances. Each daemon needs a unique identity.
 
--- a/docs/STARTUP_AND_CONFIG.md
+++ b/docs/STARTUP_AND_CONFIG.md
@@ -89,8 +89,10 @@ Configuration is loaded from `~/.mostro/settings.toml` (template: `settings.tpl.
   - Example (absolute path; use a real path — **do not** use `~`; SQLx does not expand tilde): `"sqlite:///home/youruser/.mostro/mostro.db"`
   - Default: `"sqlite://mostro.db"`
 
-**Nostr** (`src/config/types.rs:47-54`):
-- `nsec_privkey` (String): Mostro's Nostr private key in nsec format
+**Nostr** (`src/config/types.rs`):
+- `nsec_privkey_file` (String): Path to a file containing Mostro's Nostr private key in nsec format
+  - The file should contain only the `nsec` value
+  - Inline `nsec_privkey` is no longer supported and causes startup validation to fail
 - `relays` (Vec<String>): List of Nostr relay URLs for event broadcasting
   - Default: `['ws://localhost:7000']`
   - Note: At least one relay required
--- a/settings.tpl.toml
+++ b/settings.tpl.toml
@@ -19,7 +19,7 @@ payment_attempts = 3
 payment_retries_interval = 60
 
 [nostr]
-nsec_privkey = 'nsec1...'
+nsec_privkey_file = '/home/user/nostr-key.nsec'
 relays = ['ws://localhost:7000']
 
 [mostro]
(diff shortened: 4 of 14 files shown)


arkanoider:
Thanks for the PR @gotcha ! Always welcome, please fix the request of @coderabbitai to complete the work, than we will evaluate the pr!

AndreaDiazCorreia:
Hi @gotcha, thanks for the contribution. After review we've decided to close this PR: moving the nsec from settings.toml to a separate plain-text file on the same filesystem doesn't meaningfully improve the threat model (real protection comes from FS permissions, disk encryption, or a secrets manager), and the implementation is a hard breaking change. It rejects the existing nsec_privkey inline with a fatal error, so every current Mostro operator on 0.17.x would fail to start after upgrading. If you'd like to revisit this, a smaller non-breaking approach (e.g. optional MOSTRO_NSEC env var with precedence over the config) would be easier to land. Thanks again.

gotcha:
Hi @AndreaDiazCorreia 

I would like to expand the conversation. On one hand on the security aspect, and on the other hand on the backward-compatibility for users.

We proposed to configure the nsec via a path to a file to be able to write the nsec to a small file in a memory-only filesystem (a tmpfs mount like /run/secrets/), and have mostro read the file at startup.

It sounds more complicated than an environment variable, but it exists precisely because environment variables leak, and the surface area is wider than it looks. The process environment is readable at /proc/<pid>/environ by anything running as the same user, inherited by every child process you fork, and routinely captured by crash reporters and monitoring agents that snapshot the environment when an unhandled exception occurs — one bad stack trace and the nsec would be sitting in a third-party SaaS. Add the mundane leaks on top — set -x in a CI job, a pasted one-liner saved to shell history, env vars surfaced by docker inspect or kubectl describe pod to anyone with cluster read access — and the "in-memory only" mental model behind 12-factor falls apart quickly.

A file-based approach sidesteps all of this. Permissions are per-file and enforced by the kernel, so a service only sees the secrets it was explicitly granted — not the whole set. Child processes don't inherit access implicitly the way they inherit the environment. Because the filesystem lives in RAM, the bytes never touch a physical disk, survive a reboot, or end up in a backup snapshot. It's not an accident that systemd credentials, Kubernetes projected volumes, Vault Agent, and Docker/Podman secrets all converged on this same pattern — they're responses to the same decade of environment-variable incident reports.

Worth noting: the tmpfs part only matters for production secrets. For local development, any regular file with fake or non-production credentials works fine — the application code is identical, you just point it at ~/.mostro/nostr-key.nsec instead of /run/secrets/nostr-key.nsec, which makes the pattern easy to adopt without extra infrastructure on a developer laptop.

You are right that breaking without escape hatch is not really acceptable for users.  Would it be better to instead produce an error message explaining to the user that he can run a command to migrate the nsec from the general config file to a separate file.

The error message would look like this:

"""
The nostr nsec should not be inline in the `settings.toml` configuration file.
In order to fix this issue, run `mostrod move-nsec` before restarting.
"""

This implies that the new version would also have this move-nsec feature.

I would appreciate your feedback before merging #713.

grunch:
> Hi @AndreaDiazCorreia
> 
> I would like to expand the conversation. On one hand on the security aspect, and on the other hand on the backward-compatibility for users.
> 
> We proposed to configure the nsec via a path to a file to be able to write the nsec to a small file in a memory-only filesystem (a tmpfs mount like /run/secrets/), and have mostro read the file at startup.
> You are right that breaking without escape hatch is not really acceptable for users. Would it be better to instead produce an error message explaining to the user that he can run a command to migrate the nsec from the general config file to a separate file.
> 
> The error message would look like this:
> 
> """ The nostr nsec should not be inline in the `settings.toml` configuration file. In order to fix this issue, run `mostrod move-nsec` before restarting. """
> 
> This implies that the new version would also have this move-nsec feature.
> 
> I would appreciate your feedback before merging #713.

Hi @gotcha,

Thanks, this is a very useful expansion of the discussion.

I think your security argument is correct. Using a separate file for the Nostr secret is a better long-term direction than keeping the secret inline in the general config, and also better than relying on environment variables as the final model for production deployments. The reasons you described make sense: narrower access surface, no implicit inheritance to child processes, better fit with tmpfs/runtime secret mounts, and alignment with how modern secret delivery is handled in systemd, containers, and orchestration environments.

That said, I do not think we should introduce this as a hard break in the same step.

The main concern is operational backward compatibility. Today, users already have working deployments with `nsec_privkey` configured inline. If a new version refuses to start until a manual migration command is executed, that creates a real risk of outages during upgrades, especially for automated or unattended deployments. In practice, this would turn a security improvement into a service interruption.

So I think the safest path is to split this into two stages:

1. Merge Andrea's PR #713 first, which adds support for loading the nsec from an environment variable.
2. Then open a new PR implementing the file-based secret approach you propose.
   In that same PR, we should introduce a deprecation warning for `nsec_privkey` when it is found inline in the config.

This gives us a clean transition plan:
- current minor line: support the new mechanisms and emit a clear warning that `nsec_privkey` is deprecated and will be removed
- next minor line (0.18.x): remove inline `nsec_privkey` support entirely

That way we move toward the more secure file-based model without breaking existing users abruptly, and we give operators a full minor release cycle to migrate safely.

PR #712, opened 2026-04-24
feat(bond): anti-abuse bond phase 0 foundation

grunch:
Adds the inert data and configuration layer for the anti-abuse bond feature described in docs/ANTI_ABUSE_BOND.md (issue #711):

- [anti_abuse_bond] settings block, opt-in and disabled by default, with Settings::is_bond_enabled() gate.
- bonds table + Bond sqlx model + CRUD helpers under src/app/bond/.
- BondRole / BondState / BondSlashReason enums with parse roundtrips.
- Pure compute_bond_amount helper: max(amount_sats * order_amount, base_amount_sats).

No trade flow, scheduler, or LND path is touched in this PR; every code path added here is behind the disabled-by-default gate. Later phases (taker lock, dispute slash, payout, timeout slash, maker bond, range orders) will hook into the same data model.

## Summary by CodeRabbit

* **New Features**
  * Anti-abuse bond infrastructure with database storage, configuration options, and amount calculation logic.

* **Documentation**
  * Specification document detailing bond feature phases, lifecycle, database schema, configuration surface, payout workflows, and acceptance tests.

### DIFF
--- a/docs/ANTI_ABUSE_BOND.md
+++ b/docs/ANTI_ABUSE_BOND.md
@@ -46,9 +46,9 @@ Add a new optional section to `settings.toml`. Missing section ≡
 # Master switch. When false the bond code paths are dead.
 enabled = false
 
-# Bond amount = max(amount_sats * order_amount_sats, base_amount_sats)
-# amount_sats is a fraction of the order amount (0.01 = 1%).
-amount_sats = 0.01
+# Bond amount = max(amount_pct * order_amount_sats, base_amount_sats)
+# amount_pct is a unitless fraction of the order amount (0.01 = 1%).
+amount_pct = 0.01
 # Floor, in sats. Prevents trivial bonds on small orders.
 base_amount_sats = 1000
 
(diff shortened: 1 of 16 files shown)


grunch:
@coderabbitai review

AndreaDiazCorreia (commented):
Hey! I did an initial pass on this with AI assistance, and a couple of things came up that I wanted to run by you. I'm not 100% sure they're actual concerns, so take them as questions rather than blockers.

1. Should we skip serializing preimage on the Bond struct?

The Bond struct in src/app/bond/model.rs derives Serialize and the preimage field is a plain Option<String>. Phase 0 doesn't serialize bonds anywhere external, but I was wondering. Once Phases 2/3/5 wire this up to audit events, logs, or RPC responses, isn't there a risk that a Bond ends up serialized somewhere it shouldn't, leaking the preimage? My AI flagged that anyone with the preimage could claim the HTLC before Mostro does. Wouldn't it be safer to add #[serde(skip_serializing)] on preimage (and maybe payout_invoice) now, while the surface is tiny, rather than having to remember later?

2. The migration schema seems to jump ahead of Phase 0

The SQL already includes parent_bond_id, child_order_id, slashed_share_sats (Phase 6) and payout_routing_fee_sats (Phase 3), but the spec doc in §5.1 shows a smaller table (range_parent_id instead of parent_bond_id plus child_order_id, without the other two). I get the appeal of one migration vs. multiple ALTER TABLEs, but shouldn't the spec doc be updated to match what actually shipped? Otherwise it's going to drift even more as the later phases land.

3. find_bond_by_order_and_role plus LIMIT 1 feels risky for Phase 6

The query at src/app/bond/db.rs:33 assumes there's at most one row per (order_id, role), but in Phase 6 a maker parent bond would have multiple rows sharing order_id and role = 'maker' (one per child). There's no UNIQUE constraint either, so LIMIT 1 becomes non deterministic. Not a bug today since nothing calls it, but shouldn't we either filter by parent_bond_id IS NULL or return Vec<Bond> until Phase 6 nails down the semantics?

Again, sorry if any of this is off base. The AI surfaced it and I didn't want to just sit on it. I'll give it a proper review myself tomorrow, but wanted to leave these here in the meantime

grunch:
@coderabbitai review

PR #713, opened 2026-04-24
feat: support MOSTRO_NSEC_PRIVKEY env var for Nostr private key

AndreaDiazCorreia:
## Motivation

Storing the Nostr private key inline in `settings.toml` makes it easy to leak it accidentally: config files end up in logs, backups, bug reports, and chat messages when operators ask for help. It also makes integration with Docker secrets, systemd credentials, and similar tools harder than it should be.

This PR lets operators provide the nsec via the `MOSTRO_NSEC_PRIVKEY` environment variable, which is the dominant pattern across Nostr daemons (Alby Hub, NWC, OpenClaw) and follows 12-factor conventions.

## Changes

- Add `MOSTRO_NSEC_PRIVKEY` env var support. Precedence: env var >
  `<settings_dir>/.env` (auto-loaded via `dotenvy`) > `settings.toml`.
- Make `nsec_privkey` optional in `settings.toml` via `#[serde(default)]`, so
  operators relying on the env var can omit the field entirely.
- Update the setup wizard to ask where to store the generated key: either in
  `~/.mostro/.env` (recommended, `chmod 600`) or inline in `settings.toml`.
- Documentation updated with examples for shell, systemd, and Docker.
- Unit tests for all precedence cases (env wins, empty env falls back,
  whitespace-only env ignored, no env keeps TOML, trimming, TOML without the
  field).

## Backward compatibility

Fully backward-compatible. Existing `nsec_privkey = 'nsec1...'` in `settings.toml` keeps working exactly as before. The env var only takes effect when explicitly set.

## Summary by CodeRabbit

* **New Features**
  * Nostr private key can be provided via the MOSTRO_NSEC_PRIVKEY environment variable.
  * App auto-loads a local .env file and applies an environment-variable override with defined precedence.
  * Configuration wizard now offers choosing to store the private key in a .env file (recommended) or inline in settings.

* **Documentation**
  * Updated guides with environment variable usage, trimming/blank-value behavior, and precedence order.


gotcha:
I do understand how using an environment variable improves security wise compared to nsec inline in config. 

However, I am uneasy with the .env file support and its inclusion in the setup wizard. Once there is .env support that includes secrets, we would be back at that place where there is a risk to leak for instance when versioning production setup.

AndreaDiazCorreia:
Hi @gotcha , fair point.

I pushed a small change to the wizard so the two storage options are presented neutrally, without the "recommended" / "legacy" labels. The preamble also mentions that setting MOSTRO_NSEC_PRIVKEY via systemd, Docker or a secrets manager is a valid path, so nobody is nudged into a .env file if they don't want one.

The intent here is purely additive. We're not deprecating inline nsec_privkey or asking anyone to migrate. Operators who want stronger isolation can opt in, the rest keep working unchanged.

The file-path option you described fits naturally on top of this and can land in a follow-up PR whenever there's appetite.

AndreaDiazCorreia:
@coderabbitai review

grunch (approved):
LGTM

Issue #714, opened 2026-04-24
Migrate to mostro-core 0.9.1 centralized NIP-59 transport (wrap_message / unwrap_message / validate_response)


grunch:
## Context

`mostro-core` 0.9.1 ships a centralized NIP-59 transport layer that unifies how the Mostro node and every Mostro client build and consume gift-wrapped messages. Today this repo rolls its own wrap/unwrap plumbing against `mostro-core = 0.8.4` (see `Cargo.toml:73`), duplicating logic that now lives upstream and drifting from the reference client implementation.

Reference: [mostro-core `docs/NIP59_TRANSPORT.md`](https://github.com/MostroP2P/mostro-core/blob/main/docs/NIP59_TRANSPORT.md).

Goal: replace the in-tree gift-wrap code with `mostro_core::nip59::{wrap_message, unwrap_message, validate_response}` so wire-format changes, signature rules, and replay/expiration semantics are owned by one crate.

## New public API (mostro-core 0.9.1)

```rust
pub struct WrapOptions { pub pow: u8, pub expiration: Option<Timestamp>, pub signed: bool }
// Default: pow = 0, expiration = None, signed = true

pub struct UnwrappedMessage {
    pub message: Message,
    pub signature: Option<Signature>,
    pub sender: PublicKey,       // trade pubkey (rumor author == seal signer)
    pub created_at: Timestamp,   // rumor timestamp
}

pub async fn wrap_message(
    message: &Message,
    trade_keys: &Keys,
    receiver: PublicKey,
    opts: WrapOptions,
) -> Result<Event, MostroError>;

pub async fn unwrap_message(
    event: &Event,
    trade_keys: &Keys,
) -> Result<Option<UnwrappedMessage>, MostroError>;

pub fn validate_response(
    message: &Message,
    expected_request_id: Option<u64>,
) -> Result<(), MostroError>;
```

### Contracts that change

1. **Inbound unwrap is ternary**, not binary:
   - `Ok(Some(unwrapped))` → for us, process it.
   - `Ok(None)` → NIP-44 outer decrypt failed; the event is not addressed to these keys (iterate to the next candidate trade key without logging an error).
   - `Err(_)` → protocol violation: wrong kind, bad seal, `SenderMismatch`, malformed or non-verifying inner signature. Warn and drop.
2. `nostr-sdk 0.44` enforces `rumor.author == seal.signer`, so the `event.sender != event.rumor.pubkey` branches in today's code become unreachable after the migration.
[...]

PR #716, opened 2026-04-24
refactor(nip59): route send_dm through mostro_core::wrap_message (NIP-59 migration PR 2/6)

grunch:
Second PR in the NIP-59 transport migration tracked in #714. Outbound side only — replaces the hand-rolled rumor + gift-wrap chain in \`send_dm\` with a call to \`mostro_core::nip59::wrap_message\`.

## Summary

- \`src/util.rs::send_dm\`: delete the manual \`(Message, Option::<String>::None)\` tuple, \`EventBuilder::text_note(...).build(...)\` rumor, tag assembly, and \`EventBuilder::gift_wrap(...)\` call. Replace with:

  \`\`\`rust
  let opts = WrapOptions { pow: 0, expiration, signed: false };
  let event = wrap_message(&message, sender_keys, receiver_pubkey, opts).await?;
  \`\`\`

- \`signed: false\` — the Mostro node holds **no** trade keys, so the inner rumor stays \`(Message, None)\` as before. Note: \`mostro-core\` 0.9.1 does **not** auto-skip signing based on \`trade_index\`; \`signed\` is the explicit switch. The issue description suggested \`signed: true\` would behave conditionally, which was incorrect — verified against \`mostro-core-0.9.1/src/nip59.rs\`.
- Public \`send_dm\` signature unchanged; every caller in \`scheduler.rs\`, \`admin_add_solver\`, \`admin_cancel\`, \`admin_take_dispute\`, \`last_trade_index\` compiles and runs unchanged.
- \`?\` for error propagation now that \`wrap_message\` returns \`MostroError\` directly (previously the \`EventBuilder::gift_wrap\` error was manually mapped to \`NostrError\`).

## Test changes

\`test_send_dm\` was a placeholder that happened to pass only because the old \`EventBuilder::gift_wrap\` path errored in a no-client unit-test environment — the original inline comment even admitted this. Post-migration, \`wrap_message\` is pure offline crypto and succeeds; the \`if let Ok(client) = get_nostr_client()\` fall-through then returns \`Ok(())\`. But inside the full suite, a different test initialises \`NOSTR_CLIENT\` globally, which makes any \`send_dm\` assertion race-dependent on execution order.

Rewrote the test as \`test_send_dm_outbound_wrap_shape\`:

- No longer calls \`send_dm\` (would be flaky under parallel execution).
- Exercises \`wrap_message\` + \`unwrap_message\` directly, which is the crypto actually moved by this PR.
- Asserts the three invariants that matter for the Mostro → user direction: unsigned tuple, \`sender = Mostro identity\`, inner message round-trips byte-for-byte.

## Respecting the open-question answers on #714

| Question | Answer | Applied here |
|---|---|---|
| 1. Replay window | Caller-side, keep 10s guard | Inbound only, lands in PR 3 |
| 2. \`validate_response\` | Skip (node is push-only) | Not imported |
| 3. Admin bridge shape | Mechanical migration first | Still on \`UnwrappedGift\`; PR 5 handles it |
| 4. PoW | Defer to separate issue | \`pow: 0\` |

## Test plan

- [x] \`cargo check --bin mostrod --tests\` clean
- [x] \`cargo clippy --bin mostrod --all-features --tests -- -D warnings\` clean
- [x] \`cargo test --bin mostrod\` — 239 passed, 0 failed
- [ ] Smoke test against a real relay (not required for this layer — wrap output is byte-compatible with the previous \`EventBuilder::gift_wrap\` output since \`wrap_message\` uses the same underlying \`nip59\` primitives)

## Scope fence

- **No** inbound changes: \`src/app.rs::run\`'s \`Kind::GiftWrap\` branch, \`check_trade_index\`, and every handler reading \`event.rumor.pubkey\` stay exactly as they are.
- **No** RPC admin-bridge changes.
- **No** new handler-signature migration.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -509,23 +509,19 @@ pub async fn send_dm(
     );
     let message = Message::from_json(payload)
         .map_err(|_| MostroInternalErr(ServiceError::MessageSerializationError))?;
-    // We compose the content, as this is a message from Mostro
-    // and Mostro don't have trade key, we don't need to sign the payload
-    let content = (message, Option::<String>::None);
-    let content = serde_json::to_string(&content)
-        .map_err(|_| MostroInternalErr(ServiceError::MessageSerializationError))?;
-    // We create the rumor
-    let rumor = EventBuilder::text_note(content).build(sender_keys.public_key());
-    let mut tags: Vec<Tag> = Vec::with_capacity(1 + usize::from(expiration.is_some()));
 
-    if let Some(timestamp) = expiration {
-        tags.push(Tag::expiration(timestamp));
-    }
-    let tags = Tags::from_list(tags);
+    // Mostro identity holds no trade keys, so the inner rumor tuple is
+    // `(Message, None)` — `signed = false`. `wrap_message` builds the
+    // rumor, seal, outer gift wrap, ephemeral key, random-timestamp
+    // tweak and `p` tag for us; we only pick the wrap-level knobs.
+    // PoW is deferred to a follow-up (see issue #714, open question 4).
+    let opts = WrapOptions {
+        pow: 0,
+        expiration,
+        signed: false,
+    };
+    let event = wrap_message(&message, sender_keys, receiver_pubkey, opts).await?;
 
-    let event = EventBuilder::gift_wrap(sender_keys, &receiver_pubkey, rumor, tags)
-        .await
-        .map_err(|e| MostroInternalErr(ServiceError::NostrError(e.to_string())))?;
     info!(
         "Sending DM, Event ID: {} to {} with payload: {:#?}",
         event.id,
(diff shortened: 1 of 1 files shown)


grunch:
Blocked on #717 (revert of #715).

A production regression surfaced: the Mostro mobile app — and any client still on `mostro-core` 0.8.x — produces gift wraps where `rumor.pubkey != seal.pubkey` (identity signs the seal, trade key authors the rumor). `nostr-sdk` 0.44 added a strict equality check that rejects those wraps with `SenderMismatch` ("sender public key mismatch" in the server log). Details and revert PR: #717.

This PR depends on `mostro-core` 0.9.1 / `nostr-sdk` 0.44 via #715, so it has to stay idle until the migration plan is redone with coordinated client updates (tracking on #714). Leaving it open so the refactor diff is preserved for the eventual retry rather than closing and re-authoring it.

Issue #703, opened 2026-04-10
Add BOLT12 offer support for buyer invoices / payment destinations


satoshisound:
## Request

Please consider adding **BOLT12 offer support** to Mostro as an additional buyer payment destination format.

## Why

Today Mostro appears to rely on:
- BOLT11
- Lightning Address / LNURL
- LND + hold invoice architecture

That works, but it makes compatibility dependent on invoice generation or LNURL HTTP reachability. Supporting BOLT12 would improve interoperability with newer Lightning tooling and BIP353-style payment identity.

## Suggested scope

- keep existing BOLT11 / Lightning Address / LNURL behavior
- add BOLT12 offer parsing/validation
- allow BOLT12 as a buyer payout destination
- document backend/node requirements and limitations

## Key question

Since Mostro currently appears to be built around **LND gRPC + hold invoices**, is the main blocker that the current backend does not support the needed BOLT12 flow? 

I’d be interested in this feature because it would make Mostro easier to use with newer wallets which utilize Breez Liquid SDK (which uses BOLT12) and less dependent on classic LNURL flows.

PR #720, opened 2026-04-26
feat(bolt12): add BOLT12 offer payout support via LNDK

QaidVoid:
## Related Issue
Partial fix for #703. BIP-353 DNS resolution will follow as a separate PR.

## Summary
Lets buyers submit a BOLT12 offer (`lno1…`) as their payout destination. Mostro routes the payout through an optional [LNDK](https://github.com/lndk-org/lndk) daemon running alongside LND. Opt-in; default-off behavior is byte-identical to current Mostro.

## Motivation
BOLT12 closes UX gaps in BOLT11: offers are reusable (no per-retry invoice round-trip) and receiver identity is hidden via blinded paths. Mostro's escrow flow keeps using BOLT11 hold invoices (BOLT12 has no hold-invoice equivalent), so this PR only touches the buyer-payout step — the one place where the destination is chosen by an external party.

LNDK is the standard sidecar for BOLT12 with LND today: LDK implements offers, LND provides routing and onion-message forwarding, LNDK glues them together. Avoids forking LND or maintaining a parallel routing stack.

## Architecture
```text
Buyer  ---- Nostr ---->  Mostro  ----gRPC---->  LNDK  ----gRPC---->  LND
(sends lno1…)             |                      |                    |
                          +--BOLT11/LNURL path---|------gRPC--------->+
                                                 |
                                                 +- onion messaging -+
```

`do_payment` becomes a thin dispatch over a new `classify()` that routes BOLT12 offers to `do_payment_bolt12` (LNDK) and everything else to a renamed `do_payment_lnd` whose body is byte-for-byte identical to the prior `do_payment`. `pay_offer_validated()` does a defensive `GetInvoice → validate → PayInvoice` because LNDK's one-shot `PayOffer` doesn't verify the fetched invoice's amount or expiry. The LNDK client is threaded through `AppContext` and `AdminServiceImpl` so dispute resolution and admin settle/cancel can pay BOLT12 buyers as well.

Defensive checks in `validate_offer` reject non-BTC currency, mismatched fixed amounts, elapsed expiry, `Quantity::Bounded(n > 1)`, zero-amount orders, and any offer received while `lndk_enabled = false`. Wire-level fee tolerance is sent as absolute msats (`PayInvoiceRequest.fee_limit`), not LNDK's whole-percent field, so a small `max_routing_fee` doesn't get rounded up by ~10×.

## Configuration
All new fields live in the existing `[lightning]` block. Older `settings.toml` files parse unchanged via struct-level `#[serde(default)]`. Operator setup (LND build tags, custom-message flags, LNDK install, baked-macaroon recipe, troubleshooting) lives in `docs/LNDK_SETUP.md`. `lndk_tls_domain` lets remote LNDK deployments override the default localhost SNI; `lndk_min_invoice_expiry` (default 60s) keeps the freshness floor appropriate for short-lived BOLT12 invoices.

## Backward compatibility
- Default-off; no LNDK process required.
- BOLT11 / LNURL / LN-Address paths byte-identical.
- Pre-fetched BOLT12 invoices (`lni1…`) rejected at validation rather than failing later.
- Older `settings.toml` files parse unchanged.

## Scope and limitations
Deferred to follow-ups: BIP-353 (separate PR), BOLT12 offer creation (dev-fee receipt still uses BOLT11), BOLT12-aware retries (today shares the BOLT11 retry loop), seller→Mostro escrow leg (BOLT12 has no hold-invoice equivalent).

### Automated
- `cargo test`: 251 passed, 0 failed.
- `cargo clippy --all-targets --all-features`: clean.
- `cargo fmt --check`: clean.
- Unit coverage for `offers` (classify shapes, `validate_offer` rejection cases including `Quantity::Bounded(n>1)` and zero-amount orders, `validate_fetched_invoice` amount / expiry / sanity bounds).

### Regtest topology
- `bitcoind` regtest.
- Two LND v0.20.1-beta nodes (`lnd`, `lnd2`) built with `peersrpc signrpc walletrpc` and started with `--protocol.custom-message=513 --protocol.custom-nodeann=39 --protocol.custom-init=39`.
- LNDK v0.3.0 sidecar on each LND (`:7000`, `:7001`).
- CLN (Core Lightning) as a third node, peered with both LNDs and used as the BOLT12 offer issuer.
- Channels: `lnd2 → lnd` 500k sats, `cln → lnd` 500k sats, `lnd2 ↔ cln` peered for the BOLT12 reply path.

### LNDK wire-format check
LNDK v0.3.0 was triggered with the exact wire format Mostro now sends — absolute msat `fee_limit` and `payer_note: None`. From the LNDK server log:

```text
INFO lndk::server - Received a request: PayOfferRequest {
  ...,
  payer_note: None,
  fee_limit: Some(200),
  fee_limit_percent: None
}
TRACE lndk::offers::lnd_requests - Peer already known while connecting as introduction node: <CLN pubkey>
INFO lndk::onion_messenger - Sending outgoing onion message to <CLN pubkey>.
```

LNDK parsed the request, looked up the introduction node, and emitted the onion message — the new `PayInvoiceRequest.fee_limit` (msat) + `payer_note: None` encoding is wire-compatible with LNDK v0.3.0. mostrod boots cleanly against the same LNDK (`LNDK connector initialized for BOLT12 offer payouts`) with no panics, and resubscribes pending orders without errors.

### End-to-end Mostro flow on the same topology

1. **BOLT12 payout.** Maker creates a buy order with `buyer_invoice = lno1...` issued by CLN. Hold-invoice flow runs unchanged. On `release`, mostrod calls `LNDK::GetInvoice → validate → PayInvoice`. After `lnd2` was peered with CLN (so the BOLT12 reply path through it works), the retry loop converged with `BOLT12 offer paid, preimage=0cad817d…`. Channel-level proof: `lnd → cln` shifted by exactly 100,000 sats.
2. **BOLT11 regression.** Maker creates a buy order with payout = a freshly-minted `lnbcrt500u…` invoice on the seller LND. `release` triggers `do_payment_lnd`. mostrod logs `Order…: Invoice with hash: b21a252e… paid!` and transitions to `success`. No BOLT12 path touched.

## Summary by CodeRabbit

* **New Features**
  * Experimental support for paying BOLT12 offers via an optional LNDK payout path, including offer validation, invoice fetching, and payout handling.
  * New config options to enable/disable LNDK, set gRPC endpoint/TLS/macaron, timeouts, expiry and fee limits.

* **Documentation**
  * Added LNDK setup guide and updated startup/config docs describing BOLT12/LNDK configuration and troubleshooting.


Catrya (changes requested):
Hi @QaidVoid Thanks for the work on this PR, it’s a great feature. Before moving forward, it should be split into two. This PR touches 20 files, we recommend keeping PRs limited to fewer files and focused on a single feature at a time.
It mixes BOLT12 (via LNDK) and BIP-353 (DoH resolver), which have different dependencies, risks, and operational implications. They shouldn’t be reviewed or shipped as a single unit.
BIP-353 depends on BOLT12, but not the other way around. It makes more sense to land BOLT12 first, and then in other PR add BIP-353.

Catrya:
@QaidVoid I also asked Claude Code (Opus 4.7) to do a deep review, and here’s its feedback, please take this along with CodeRabbit’s review into account when working on the other PRs. Happy to discuss any of these. Thanks again for the work, splitting and addressing the issues above will get this landed cleanly.

Issues found in BOLT12 (PR 1 scope)

Blocking

- Admin/dispute paths cannot pay BOLT12 buyers. `src/rpc/service.rs` constructs `AppContext` with `lndk: None` in all four admin actions (`admin_cancel`, `admin_settle`, `admin_add_solver`, `admin_take_dispute`). If `admin_settle` or a dispute resolution triggers a payout to a buyer whose `buyer_invoice` is a BOLT12 offer, `do_payment_bolt12` will fail with `"BOLT12 offer received but LNDK is disabled"`. Disputes for BOLT12 buyers will get stuck. Either pass the LNDK client through to the RPC layer, or hoist it to a global like `MOSTRO_CONFIG`.
- `fee_limit_percent` calculation inflates fee tolerance ~10×. In `src/lightning/lndk.rs`: `let fee_fraction = ln.lndk_fee_limit_percent.unwrap_or(mostro_settings.max_routing_fee); let fee_limit_percent = ((fee_fraction * 100.0).ceil().max(1.0)) as u32;`
- LNDK expects whole-number percent; `max_routing_fee` is a fraction. With the typical `max_routing_fee = 0.001` (0.1%), this becomes `ceil(0.1).max(1.0) = 1 → 1%`, a 10× increase versus the BOLT11 path. With `0.0001` (0.01%) the ratio is 100×. Operators will reasonably expect the same `max_routing_fee` setting to mean the same thing on both paths. Consider using `PayInvoiceRequest.fee_limit` (absolute msat: `ceil(amount_msats * fee_fraction)`) instead of `fee_limit_percent` to preserve precision.
- `Quantity::Bounded(n) if n.get() < 1` is dead code. In `validate_offer` (`src/lightning/offers.rs`), `NonZeroU64::get()` is always ≥ 1, so the branch is unreachable and any `Bounded(n)` is silently accepted. If an offer requires a minimum quantity > 1, Mostro accepts it and the failure surfaces later as a confusing LNDK `GetInvoice` error. The intent was probably to reject offers that can't satisfy a single-unit purchase — please rework that match.
- `expected_msats == 0` silently accepts any offer amount. Same function, the `if expected_msats != 0 && amount_msats != expected_msats` skip lets a zero-amount order accept an offer with an arbitrary fixed amount. Defense-in-depth should error here, not accept.

Serious

- Hardcoded SNI `"localhost"`. `src/lightning/lndk.rs` does `.domain_name("localhost")` regardless of the configured `lndk_grpc_host`. Remote LNDK deployments are blocked. Please make the TLS domain name configurable (e.g. `lndk_tls_domain`, default `"localhost"`).
- `payer_note: format!("mostro:{order_id}")` leaks correlation to the recipient wallet. The receiver sees `mostro:<uuid>` and can correlate the order ID (which is publicly known via Nostr) with their wallet. Consider omitting the payer note or using a truncated hash without the `mostro:` prefix.
- `min_expiry_secs = invoice_expiration_window` may be overly strict. BOLT12 invoices are short-lived by design (fetched right before payment). Reusing the 15-minute hold window as the minimum remaining lifetime can reject legitimately fresh invoices. Worth a separate, smaller setting (e.g. `bolt12_min_invoice_expiry`, default `60s`).
- Vendored `proto/lndkrpc.proto` pinned at `v0.3.0` by hand. If LNDK upstream changes a field, Mostro silently breaks at runtime. At minimum, document the LNDK version requirement loudly in `LNDK_SETUP.md`. Ideally pull the proto from a tagged release as part of the build.
- `do_payment_bolt12` is fire-and-forget without panic/error visibility. `src/lightning/lndk.rs` `tokio::spawn` returns immediately; if the task panics, nothing is reported. This matches the existing LND streaming behavior, so it's consistent — but worth instrumenting (`JoinHandle` capture, or tracing span on the spawned task).
- No integration tests for LNDK gRPC. Bugs like the `fee_limit_percent` one above wouldn't be caught by the current unit tests. The PR mentions a regtest topology was used, but there's nothing automated. For an experimental opt-in feature this is acceptable for now, but worth tracking in a follow-up.

Minor

- `kind_tag(InvoiceFormat::Bolt12Invoice)` returns `Some("bolt12_invoice")` but `is_valid_invoice` always rejects that variant. Dead path; either remove it or add a comment that it exists for forward compatibility.
- Onion-message feature bit detection (in `main.rs`) — checking bits `38/39` advertises onion-message support, but doesn't guarantee the operator actually started LND with `--protocol.custom-message=513`. The warning is informative only; consider linking the exact config line in the message.
- Macaroon scope in docs. The setup guide should explicitly recommend a baked macaroon (`lncli bakemacaroon`) with minimum permissions rather than `admin.macaroon`. Compromise of LNDK currently means compromise of LND admin.

Issues to address in BIP-353 (PR 2)

Please don't ship these as-is in any future PR; flagging now so they're on the radar:

- Critical: BIP-353 resolution is one-shot at order creation, defeating offer rotation. In `src/util.rs`, `validate_invoice` resolves `user@domain` to a BOLT12 offer and the resolved offer gets written to `order.buyer_invoice`. At payout time (potentially hours later), `do_payment` reads the cached offer and pays it without re-resolving DNS. The whole point of BIP-353 is that the wallet can rotate the offer in DNS between the time the buyer provides their address and payout — caching the offer at order creation eliminates that. The original `user@domain` should be stored separately (e.g. `buyer_invoice_alias`) and re-resolved at payout time, with the cached offer or LNURL as fallback.
- Privacy: full `user@domain` is logged at `info` and `warn` levels. In `bip353.rs` and the BIP-353 block in `src/util.rs::validate_invoice`. Any log aggregator (Loki, ELK, Datadog) captures buyer payment identifiers. Please log only the domain or a hashed local-part.
- Privacy: Cloudflare DoH default exposes payment metadata to a third party. Every resolution sends `<user>.user._bitcoin-payment.<domain>` from the Mostro node's IP to Cloudflare — i.e. "Mostro is about to pay `user@domain`". This isn't documented as a privacy trade-off in `STARTUP_AND_CONFIG.md`. Consider defaulting to a less-centralized option, or at minimum documenting the implication and recommending operators run a local resolver.
- `bip353_skip_dnssec` is a footgun with no runtime warning. Without DNSSEC validation, an attacker on the DNS path can redirect payouts. The flag is documented as "regtest only" in a comment, but nothing prints a loud warning at startup if it's enabled in production. At minimum, log a banner when the flag is set.
- `"DNSSEC validated"` is trust-the-resolver, not end-to-end. The AD flag comes from the DoH resolver — if the resolver is compromised or the TLS to it is intercepted, the flag is forgeable. The doc should be explicit that Mostro trusts the resolver to have validated DNSSEC, rather than implying end-to-end DNSSEC verification.

QaidVoid:
@coderabbitai review

QaidVoid:
Hi @Catrya, thanks for the thorough review. I've pushed fixes addressing all the feedback. This PR is now focused only on BOLT12. Ready for another look when you have a moment.

grunch:
@QaidVoid thank you for your valuable contribution, this is a great new feature that Mostro users would appreciate.

Unfortunately this PR is getting too big, we are already in production with mostros around the world and we can't afford to introduce a bug.

When we need to implement features like this one we split it on smaller phases, you can see this example on closed issues on recently we are doing a big and sensitive multiple phases PR to add anti abuse bond (https://github.com/MostroP2P/mostro/pull/731)

The spec must indicate all phases and what do each one in detail and update the spec indicating when each one is completed.

Every phase must be backwards compatible if possible, if not we should indicate it in the spec at the beginning to prepare a minor release in the future.

If you have doubts let me know

QaidVoid:
@grunch, thanks for your feedback. I'll look into splitting out the feature into multiple phases PR..

grunch:
> @grunch, thanks for your feedback. I'll look into splitting out the feature into multiple phases PR..

Thanks @QaidVoid we appreciate it 🙏🏽

PR #728, opened 2026-05-08
feat(bond): align AntiAbuseBondSettings with spec — slash split, claim window, drop unused dispute flag

grunch:
## Summary

Three changes to \`[anti_abuse_bond]\` so the runtime config matches
the \`docs/ANTI_ABUSE_BOND.md\` spec:

- **Add \`slash_node_share_pct: f64\` (default \`0.5\`).** Fraction of
  a slashed bond the node retains; the rest goes to the winning
  counterparty. Funds solver compensation for dispute work without
  introducing a new payment rail. \`0.0\` keeps the legacy
  winner-takes-all semantics; \`1.0\` makes the bond a pure sybil
  cost.

- **Add \`payout_claim_window_days: u32\` (default \`15\`).** Long-stop
  forfeiture: if the wronged counterparty never submits a payout
  invoice within this window, the bond closes as \`forfeited\` and
  the node retains the counterparty share too.

- **Remove \`slash_on_lost_dispute: bool\`.** The spec drops it
  explicitly in §3 — dispute slashes are expressed by the solver
  per-resolution via the \`BondResolution\` payload (Phase 2). The
  flag was opt-in dead code with no runtime behaviour, so removal
  does not change behaviour. A toml file with
  \`slash_on_lost_dispute = true\` from before this lands continues
  to parse (serde ignores unknown fields on this struct), so no
  operator action is needed at upgrade time.

Doc comments on \`slash_on_waiting_timeout\` and \`payout_max_retries\`
gain context: the former notes the \`BondResolution\` design for
disputes; the latter clarifies it counts only \`send_payment\` retries
against an already-received invoice (independent from the claim
window).

\`settings.tpl.toml\` updated to match.

No behaviour change in Phase 0/1 — the new fields are read by Phase
3+ scheduler logic that has not landed yet.

## Test plan

- [x] \`cargo build\`
- [x] \`cargo test config::\` — 24/24 green (incl. new
      \`toml_slash_node_share_pct_and_claim_window_override\`)
- [x] \`cargo test bond::\` — 32/32 green (no regression)
- [x] \`cargo clippy --all-targets\` — clean

## Summary by CodeRabbit

* **Chores**
  * Updated anti-abuse bond configuration settings: removed the `slash_on_lost_dispute` flag and added `slash_node_share_pct` and `payout_claim_window_days` parameters for enhanced control over bond payouts and dispute slashing mechanisms.

### DIFF
--- a/settings.tpl.toml
+++ b/settings.tpl.toml
@@ -100,7 +100,14 @@ port = 50051
 # base_amount_sats = 1000
 # # "take" | "create" | "both"
 # apply_to = "take"
-# slash_on_lost_dispute = true
+# # Note: there is no `slash_on_lost_dispute` flag — dispute slashes are
+# # solver-directed via the `BondResolution` payload (Phase 2).
 # slash_on_waiting_timeout = false
+# # Fraction of a slashed bond that the node retains (the rest goes to the
+# # winning counterparty). Funds solver compensation for dispute work.
+# slash_node_share_pct = 0.5
 # payout_invoice_window_seconds = 300
 # payout_max_retries = 5
+# # Days the winner has to claim their share by submitting a bolt11; after
+# # this the bond closes as `forfeited` and the node retains everything.
+# payout_claim_window_days = 15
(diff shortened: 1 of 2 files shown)


grunch:
@coderabbitai review

PR #747, opened 2026-05-22
feat(price): Phase 0 — multi-source price module foundation

grunch:
## What

First atomic PR of the multi-source price rollout ([`docs/PRICE_PROVIDERS.md`](docs/PRICE_PROVIDERS.md) §9, **Phase 0**), which removes the Yadio single point of failure. This is the **foundation only**: the pluggable provider interface, the pure aggregation core, the price store, and the typed `[price]` config. **Purely additive** — no HTTP, no scheduler change, no consumer change, no behavior change.

The module is not yet referenced by `main`/handlers, so it carries a scoped `#![allow(dead_code)]` on `src/price/mod.rs`, **removed in Phase 1** when `PriceManager` wires the registry into the scheduler and `get_bitcoin_price`.

## New `src/price/`

- **`provider.rs`** — `PriceProvider` trait (async, object-safe via `async-trait` so the registry can hold `Vec<Box<dyn PriceProvider>>`); `Quote` (`PerBtc` direct | `PerBase` fiat-cross); `ProviderId`; `ProviderError`; and `ProviderHealth` circuit-breaker state (pure, clock injected; exponential backoff + cap).
- **`aggregate.rs`** — the pure §6 core: `combine` (median + outlier guard; cleans `NaN`/`inf`/`≤0`; mean for 2, single for 1), `resolve_per_base` (fiat-cross → per-BTC against the USD anchor, §6.3), `aggregate_tick` (the §5.3 pipeline; uppercases codes so providers that disagree on casing still combine, §6.6).
- **`store.rs`** — `PriceStore` (RwLock map) with last-known-good semantics (currencies absent from a tick keep their prior entry, §6.4) and a staleness-checked `get` returning a local `PriceError` (mapped onto `MostroError` in Phase 1/4 per §10.2).
- **`config.rs`** — `PriceSettings` + `ProviderConfig` (`enabled`/`url`/`fallback_urls`/`api_key`/`token`/`only`/`except`) with §7 defaults and validation (`only`⊕`except`, outlier range).

## Wiring

`Settings` gains `Option<PriceSettings>` + `Settings::get_price()` (mirrors `get_bond`); every literal `Settings { .. }` constructor (wizard + test helpers) updated. Adds the `async-trait` dependency.

## Tests (38 new) — Phase 0 acceptance

- **combine**: 0/1/2/≥3 sources, outlier discarded at the boundary (inclusive), all-equal, `NaN`/`inf`/`≤0` rejected.
[...]

PR #748, opened 2026-05-22
fix(price): tolerate null rates in Yadio /exrates/BTC response

grunch:
## Diagnosis

Running `mostrod` on `main` logs every tick:

```
ERROR mostrod::scheduler: Failed to update Bitcoin prices: Error caused by Error serializing message
```

Yadio's `GET /exrates/BTC` now returns a **`null`** value for at least one currency:

```json
{ "BTC": { "USD": 75899.55, "EUR": 65393.99, "BGN": null, ... }, "base": "BTC", "timestamp": 1779480604069 }
```

`BitcoinPriceManager::update_prices` parses that into `YadioResponse { btc: HashMap<String, f64> }`. serde cannot deserialize `null` into `f64`, so the **whole** response is rejected — mapped to `ServiceError::MessageSerializationError` (Display: "Error serializing message"). A single unpriced currency (BGN) takes down BTC prices for **all** 100+ currencies, leaving clients on stale data.

## Fix

Parse leniently as `HashMap<String, Option<f64>>` and keep only currencies with a usable rate — dropping `null` and any non-finite / non-positive value — before storing in `BITCOIN_PRICES` and publishing to Nostr. The unpriced currency is simply absent until Yadio prices it again; everything else updates normally.

Minimal and targeted: only `src/bitcoin_price.rs` (the `/exrates/BTC` cached path). The per-take `/convert` path in `get_market_quote` queries a single currency and is unaffected.

## Tests

- New regression test `test_yadio_response_with_null_rate_is_parsed_and_filtered` using the exact payload shape (USD, EUR, `BGN: null`): the lenient parse accepts it and the filter drops only BGN.
- Existing deserialization tests updated for the `Option<f64>` shape; the malformed-JSON / null-map / string-value error tests still reject as before.
- `cargo build`, `cargo test` (bitcoin_price: 11 pass), `cargo clippy --all-targets --all-features`, `cargo fmt` all clean.

## Note

This is the immediate hotfix for the Yadio single point of failure. The in-progress multi-source price module (spec `docs/PRICE_PROVIDERS.md`, Phase 0 in #747) removes the dependency on Yadio alone and its Yadio adapter will carry the same lenient parsing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing exchange rate data. The price system now gracefully processes incomplete currency rate information, filtering out invalid values while maintaining stability and accuracy of available rates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/MostroP2P/mostro/pull/748?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/bitcoin_price.rs
+++ b/src/bitcoin_price.rs
@@ -13,8 +13,13 @@ use tracing::{error, info};
 
 #[derive(Debug, Deserialize)]
 struct YadioResponse {
+    // Yadio reports `null` for currencies it currently has no rate for
+    // (observed: `"BGN": null`). A strict `HashMap<String, f64>` makes serde
+    // reject the *entire* response on the first null, taking every currency
+    // down with it. Parse leniently as `Option<f64>` and drop the bad
+    // entries in `update_prices`.
     #[serde(rename = "BTC")]
-    btc: HashMap<String, f64>,
+    btc: HashMap<String, Option<f64>>,
 }
 
 static BITCOIN_PRICES: Lazy<RwLock<HashMap<String, f64>>> =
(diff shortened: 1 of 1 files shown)


Catrya (approved):
tACK

arkanoider (approved):
tACK!

PR #752, opened 2026-05-26
fix: surface InvalidOrderId to clients as CantDo(NotFound)

### LINKED ISSUE #751: Admin actions silently drop InvalidOrderId errors instead of replying with CantDo(NotFound)
## Summary
When an admin action (`AdminSettle`, `AdminCancel`) receives an `order_id` that is not present in the database, the daemon returns `MostroError::MostroInternalErr(ServiceError::InvalidOrderId)`, which is mapped to `warning_msg(...)` in `app.rs::manage_errors`. `warning_msg`
only logs locally via `tracing::warn!` and never enqueues a DM back to the sender. The client therefore never learns the action failed, and
ends up waiting for a response that never comes (timing out).

This is the root cause of MostroP2P/mostro-cli#170 on the daemon side (the CLI side is being fixed independently).

## Current behavior

1. Admin sends `Action::AdminSettle` with a non-existent `order_id`.
2. `admin_settle_action` calls `get_order(...)` which returns `Err(MostroInternalErr(ServiceError::InvalidOrderId))` (see `mostro/src/util.rs:1169-1182`).
3. `manage_errors` (`mostro/src/app.rs:88-107`) matches the `MostroInternalErr` arm, which calls `warning_msg(action, e)` and returns without sending any DM.
4. Backend logs show: `Error in AdminSettle with context Order id not present in database - inner message No message`
5. The admin client receives nothing.

## Expected behavior

`InvalidOrderId` is a user-correctable input error, not an internal fault — the client should be told. The daemon should send a
`CantDo(NotFound)` (or a more specific reason if added) DM back to the sender so the admin sees that the order id was invalid.

A broader sweep is probably worthwhile: any `ServiceError` variant that is actually caused by user input (invalid id, malformed input, etc.)
should probably be promoted to a `CantDo` rather than left as an internal error.

## Suggested fix

Either:
- In `get_order`, return `MostroCantDo(CantDoReason::NotFound)` instead of `MostroInternalErr(ServiceError::InvalidOrderId)`; or
- In `manage_errors`, treat the specific `ServiceError::InvalidOrderId variant by enqueueing a `CantDo(NotFound)` DM in addition to logging.

The first option is more surgical and matches the semantics of `CantDoReason` (machine-readable, intended to be surfaced to users).

## Affected actions

Confirmed: `AdminSettle`, `AdminCancel`. Likely the same pattern affects any handler that calls `get_order` and returns its error unmodified.

## Additional context

Related CLI issue: MostroP2P/mostro-cli#170

AndreaDiazCorreia:
## Summary

Closes #751.

Admin actions (`AdminSettle`, `AdminCancel`) — and any handler that called `get_order` — used to return `MostroInternalErr(ServiceError::InvalidOrderId)` when the message lacked an `id` or the order was not present in the database. That error was routed to `manage_errors`, which only calls `warning_msg` and never enqueues a DM back to the sender. The client therefore never learned the action failed and timed out.

This change promotes the user-facing path of `InvalidOrderId` to `MostroCantDo(CantDoReason::NotFound)`, so the sender receives a `CantDo` DM. It matches the existing `CantDo` pattern already used in `last_trade_index_action` and `orders_action`, and is the option recommended in the issue.

## Changes

- `util.rs::get_order`: return `CantDo(NotFound)` when the message lacks an `id` and when the order is not present in the database.
- `app/dispute.rs::dispute_action`: return `CantDo(NotFound)` when the message lacks an `order_id`.
- `app/admin_take_dispute.rs::admin_take_dispute_action`:
  - Preserve `DbAccessError` semantics on real database failures (previously silently mapped to `InvalidOrderId`).
  - Return `CantDo(NotFound)` when the dispute's referenced order is missing, so the admin gets feedback.
- Unit tests for the three `get_order` branches: missing id, order not found, order found.

## Affected actions

`AdminSettle`, `AdminCancel`, `Dispute`, `AdminTakeDispute`, and any other handler that resolves an order through `get_order` (`FiatSent`, `TakeBuy`, `TakeSell`, `AddInvoice`, `Cancel`, `Release`, `RateUser`, `TradePubkey`).

## Test plan

- [x] `cargo build --tests -p mostro` — clean.
- [x] `cargo test -p mostro --bin mostrod` — 334 passed, 0 failed (3 new tests added).
- [x] `cargo clippy -p mostro --bin mostrod --tests` — no new warnings.
- [x] Manual: send `AdminSettle` with a non-existent `order_id` and verify the admin receives a `CantDo(NotFound)` DM instead of a timeout.

## Related

- CLI side: MostroP2P/mostro-cli#170

### DIFF
--- a/src/app/admin_take_dispute.rs
+++ b/src/app/admin_take_dispute.rs
@@ -180,11 +180,11 @@ pub async fn admin_take_dispute_action(
     // Get order from db using the dispute order id
     let order = if let Some(order) = Order::by_id(pool, dispute.order_id)
         .await
-        .map_err(|_| MostroInternalErr(ServiceError::InvalidOrderId))?
+        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?
     {
         order
     } else {
-        return Err(MostroInternalErr(ServiceError::InvalidOrderId));
+        return Err(MostroCantDo(CantDoReason::NotFound));
     };
 
     // Update dispute fields
--- a/src/app/dispute.rs
+++ b/src/app/dispute.rs
@@ -159,7 +159,7 @@ pub async fn dispute_action(
     let order_id = if let Some(order_id) = msg.get_inner_message_kind().id {
         order_id
     } else {
-        return Err(MostroInternalErr(ServiceError::InvalidOrderId));
+        return Err(MostroCantDo(CantDoReason::NotFound));
     };
     // Check dispute for this order id is yet present.
     if find_dispute_by_order_id(pool, order_id).await.is_ok() {
--- a/src/util.rs
+++ b/src/util.rs
@@ -1170,14 +1170,14 @@ pub async fn get_order(msg: &Message, pool: &Pool<Sqlite>) -> Result<Order, Most
     let order_msg = msg.get_inner_message_kind();
     let order_id = order_msg
         .id
-        .ok_or(MostroInternalErr(ServiceError::InvalidOrderId))?;
+        .ok_or(MostroCantDo(CantDoReason::NotFound))?;
     let order = Order::by_id(pool, order_id)
         .await
         .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
     if let Some(order) = order {
         Ok(order)
     } else {
-        Err(MostroInternalErr(ServiceError::InvalidOrderId))
+        Err(MostroCantDo(CantDoReason::NotFound))
     }
 }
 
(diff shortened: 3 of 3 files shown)


codaMW:
> Reviewed the three changed files and the test cases. The fix is correct and the approach is well-chosen.
>
> The core change in `get_order` is exactly right, both the missing-id case and the not-found case are genuinely user-facing conditions, not internal server errors. Routing them through `MostroCantDo(CantDoReason::NotFound)` means the sender receives a `CantDo` DM instead of silently timing out. The distinction preserved in `admin_take_dispute.rs` is also important: real database failures correctly stay as `DbAccessError` (internal), while a missing order correctly becomes `NotFound` (user-facing). That separation shows good understanding of the error model.
>
> The three unit tests cover the full decision tree of `get_order`  missing id, id present but order absent, and the happy path. The use of `setup_orders_pool()` with in-memory SQLite and the actual migration SQL is the right pattern it tests against the real schema rather than a hand-rolled approximation.
>
> One observation worth noting: the `admin_take_dispute` fix also silently upgraded what was previously a blanket `InvalidOrderId` mapping on the dispute's order lookup to a proper `DbAccessError` for real DB failures. That's a correctness improvement worth calling out explicitly in the PR description since it changes error semantics for infrastructure failures, not just the not-found case.
>
> cACK: logic is sound, tests are solid, change is minimal and targeted.

---

Issue #601, opened 2026-02-17
Lost recovery after restore-session when invoice payment failed in settled-hold-invoice state


Catrya:
If an order reaches the `settled-hold-invoice` status but the Lightning payment to the buyer invoice fails, Mostro sends an `add-invoice` action request so the user can provide a new invoice.

However, if the user performs a `restore-session` at this point, Mostro only returns the order with status `settled-hold-invoice`, does not resend the `add-invoice` action or any signal indicating that the previous payment attempt failed.

Because of this, the app has no way to know that:
- the invoice payment failed, and
- the buyer must provide a new invoice.

As a result, the order appears as successfully settled, while in reality the buyer has not been paid and has no way to receive their sats anymore.

arkanoider:
@mostronator can you propose a fix for this? 😄

BraCR10:
Hello team! How is this issue going? Can I work on this one, or is someone already working on it?

BraCR10:
> Hello [@BraCR10](https://github.com/BraCR10), Have you already taken this issue? Thank you

Hi @codaMW, feel free to suggest a solution

grunch:
@codaMW this is a good one if you want to work on it

arkanoider:
> ### Example: Restore-Session Response Format
> Here is an example of how the restore-session response should look **before** and **after** the fix.
> 
> #### Before (Current Behavior) — Buyer Loses Context
> {
>   "version": 1,
>   "request_id": 42,
>   "action": "RestoreSession",
>   "content": {
>     "orders": [
>       {
>         "id": "order-uuid-123",
>         "status": "settled-hold-invoice",
>         "amount": 100000,
>         "fiat_amount": 50,
>         "fiat_code": "USD",
>         "payment_method": "bank transfer",
>         "premium": 0,
>         "buyer_pubkey": "npub1buyer...",
>         "seller_pubkey": "npub1seller..."
>       }
>     ]
>   }
> }
> **Problem:** The buyer sees `status: settled-hold-invoice` and assumes the trade completed successfully. They have no indication that the Lightning payment failed and that they must provide a new invoice.
> 
> #### After (Fixed Behavior) — Buyer Informed of Retry Need
> {
>   "version": 1,
>   "request_id": 42,
>   "action": "RestoreSession",
>   "content": {
>     "orders": [
>       {
>         "id": "order-uuid-123",
>         "status": "settled-hold-invoice",
>         "amount": 100000,
>         "fiat_amount": 50,
[...]

PR #754, opened 2026-05-28
fix(restore-session): re-send AddInvoice for failed payments on session restore

codaMW:
## Problem

When a Lightning payment fails after a hold invoice is settled, Mostro sets `failed_payment = true` on the order and sends the buyer an `Action::AddInvoice` message asking them to provide a new invoice. This works correctly on first attempt.

However, if the buyer disconnects before responding and later performs a `restore-session` to recover their pending orders, Mostro returned the order list correctly but **never re-sent the `AddInvoice` message**. The buyer could see the order in `settled-hold-invoice` status but had no prompt to act on it and no way to know a new invoice was required. The order would be permanently stuck.

This is a real money-at-risk bug: the seller's hold invoice has already been settled, the buyer's sats are allocated, but the payment cannot complete because the buyer has no invoice to submit.

## Root Cause

The `restore-session` handler (`src/app/restore_session.rs`) only sends back the list of active orders and disputes. It had no logic to detect orders requiring active buyer participation, specifically orders in `settled-hold-invoice` status with `failed_payment = true`, and re-trigger the necessary prompt.

The original `AddInvoice` message is ephemeral (in-memory queue). Once the buyer disconnects it is gone. `restore-session` was the only reconnect mechanism and it did not reconcile this gap.

## Fix

After `send_restore_session_response` sends the order list back to the user, a reconciliation step is added:

1. Call `find_failed_payment_for_master_key`  a new targeted SQL query that fetches only orders where `failed_payment = true AND status = 'settled-hold-invoice' AND master_buyer_pubkey = ?` for the reconnecting user
2. For each matching order, re-enqueue `Action::AddInvoice` to the buyer

The fix is intentionally minimal:

- Zero changes to the existing restore flow, its function signatures, or the spawn/pool threading
- Only adds code **after** the existing response is already sent
- Falls through silently if no failed orders exist (the common case)
- Uses `get_db_pool()` which references the same underlying pool as the rest of the daemon

## Safety

The existing payment safety model is completely unchanged. Atomic state transitions and serial scheduler execution already prevent double payments even if `AddInvoice` is sent multiple times. This fix only adds a notification. It does not touch any payment state.

## Manual Testing

Tested on a local regtest stack: **bitcoind v25 + LND v0.18.5 + nostr-rs-relay + Mostro daemon (this branch) + mostro-cli**.

**Steps:**

1. Started all services locally (bitcoind regtest, LND funded with 10 BTC, local Nostr relay on port 7777, Mostro daemon)
2. Created a sell order via mostro-cli
3. Manually set the order to the failure state in Mostro's SQLite DB:

```sql
UPDATE orders
SET status = 'settled-hold-invoice',
    failed_payment = 1,
    master_buyer_pubkey = '<test_pubkey>',
    buyer_pubkey = '<test_pubkey>'
WHERE id = '<order_id>';
```

4. Ran `restore-session` via mostro-cli as the buyer identity
5. Observed Mostro daemon logs

**Daemon logs confirming the fix works:**

```
INFO mostrod::app::restore_session: Starting background restore session for master key: 3dccb248...
INFO mostrod::db: Background restore session completed with 2 orders, 0 disputes
INFO mostrod::app::restore_session: Restore session response sent to user 3dccb248...
INFO mostrod::app::restore_session: Re-sent AddInvoice for order e8d334b3-4d74-4d59-bd7e-957e3ee35ad0 on restore-session (failed payment)
INFO mostrod::util: Sending message ... "action":"add-invoice" ... to 3dccb248...
```

**Verified existing restore-session is unaffected:** A normal pending order restored correctly with no regression.

**CLI note:** The CLI showed `Received response with mismatched action. Expected: RestoreSession, Got: AddInvoice`. The AddInvoice arrived correctly at the client but the CLI does not currently handle receiving AddInvoice interleaved with a restore-session response. This is a pre-existing CLI-side limitation tracked separately in mostro-cli#170, not introduced by this PR.

## Files Changed

- `src/app/restore_session.rs` Added `master_key` parameter threading and AddInvoice re-send logic after restore response
- `src/db.rs` Added `find_failed_payment_for_master_key` helper function with three unit tests

## Unit Tests Added

Three tests for `find_failed_payment_for_master_key` in `src/db.rs`:

- `test_find_failed_payment_for_master_key_returns_matching`  correct order is found
- `test_find_failed_payment_for_master_key_ignores_different_key` other users' orders not returned
- `test_find_failed_payment_for_master_key_ignores_non_failed` orders with `failed_payment = false` not returned

All 330+ existing tests continue to pass.

## Pull Request Checklist

- [x] Manually tested the fix on a local regtest stack
- [x] Verified existing restore-session flow is unaffected
- [x] Unit tests added for the new DB function
- [x] Build clean, clippy clean, fmt clean
- [x] GPG signed commit

Disclosure: I consulted Claude to understand the codebase structure but the solution, testing, and verification were done by myself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session restoration now detects orders with failed payments, logs lookup errors or missing buyer keys, and re-enqueues their invoices in FIFO order (skipping invalid orders).
* **New Features**
  * Added DB lookup and queue routing to locate and retry failed-payment orders tied to a restoring account; preserves response/order ordering.
* **Tests**
  * Added tests for detecting and filtering failed-payment orders during restoration.

### DIFF
--- a/src/app/restore_session.rs
+++ b/src/app/restore_session.rs
@@ -1,5 +1,7 @@
 use crate::app::context::AppContext;
-use crate::{db::RestoreSessionManager, util::enqueue_restore_session_msg};
+use crate::config::settings::get_db_pool;
+use crate::db::{find_failed_payment_for_master_key, RestoreSessionManager};
+use crate::util::{enqueue_order_msg_on_restore_queue, enqueue_restore_session_msg};
 use mostro_core::prelude::*;
 use nostr_sdk::prelude::*;
 
@@ -42,14 +44,18 @@ pub async fn restore_session_action(
 
     // Start a background task to handle the results
     tokio::spawn(async move {
-        handle_restore_session_results(manager, trade_key).await;
+        handle_restore_session_results(manager, trade_key, master_key).await;
     });
 
     Ok(())
 }
 
 /// Handle restore session results in the background
-async fn handle_restore_session_results(mut manager: RestoreSessionManager, trade_key: String) {
+async fn handle_restore_session_results(
+    mut manager: RestoreSessionManager,
+    trade_key: String,
+    master_key: String,
+) {
     // Wait for the result with a timeout
     let timeout = tokio::time::Duration::from_secs(60 * 60); // 1 hour timeout
 
@@ -58,6 +64,7 @@ async fn handle_restore_session_results(mut manager: RestoreSessionManager, trad
             // Send the restore session response
             if let Err(e) = send_restore_session_response(
                 &trade_key,
+                &master_key,
                 result.restore_orders,
                 result.restore_disputes,
             )
@@ -82,6 +89,7 @@ async fn handle_restore_session_results(mut manager: RestoreSessionManager, trad
 /// Send restore session response to the user
 async fn send_restore_session_response(
     trade_key: &str,
+    master_key: &str,
     orders: Vec<RestoredOrdersInfo>,
     disputes: Vec<RestoredDisputesInfo>,
 ) -> Result<(), MostroError> {
@@ -101,6 +109,58 @@ async fn send_restore_session_response(
 
     // No key in the log line (AGENTS.md: scrub Nostr keys from logs).
     tracing::info!("Restore session response sent to user");
+
+    // Re-send AddInvoice for any orders stuck in settled-hold-invoice with a failed payment.
+    // Uses get_db_pool() since pool is not available in this function's scope.
+    //
+    // `find_failed_payment_for_master_key` filters on `master_buyer_pubkey =
+    // master_key`, so every order returned belongs to this restoring user as
+    // the BUYER. The correct AddInvoice recipient is therefore the order's
+    // own buyer trade key (`order.buyer_pubkey` / get_buyer_pubkey()) — the
+    // key the client actually listens on for order DMs. Sending to the
+    // master/identity key instead would (a) never reach the client, since
+    // order messages are only delivered to trade keys, and (b) publish the
+    // identity key as a gift-wrap recipient on Nostr, linking it to this
+    // order and breaking trade-key unlinkability.
+    let pool = get_db_pool();
+    match find_failed_payment_for_master_key(&pool, master_key).await {
+        Ok(failed_orders) => {
+            for order in failed_orders {
+                let buyer_trade_pubkey = match order.get_buyer_pubkey() {
+                    Ok(pk) => pk,
+                    Err(_) => {
+                        tracing::warn!(
+                            "Order {} has no valid buyer_pubkey (trade key); skipping AddInvoice re-send",
+                            order.id
+                        );
+                        continue;
+                    }
+                };
+                // Route through the restore-session queue so this AddInvoice
+                // is sent AFTER the restore-session response (same queue =
+                // FIFO), not before it.
+                enqueue_order_msg_on_restore_queue(
+                    Some(order.id),
+                    Action::AddInvoice,
+                    Some(Payload::Order(SmallOrder::from(order.clone()))),
+                    buyer_trade_pubkey,
+                    order.trade_index_buyer,
+                )
+                .await;
+                // No key in the log line (AGENTS.md: scrub Nostr keys from logs).
+                tracing::info!(
+                    "Re-sent AddInvoice for order {} to buyer trade key (failed payment)",
+                    order.id
+                );
+            }
+        }
+        Err(e) => {
+            tracing::error!(
+                "Failed to query failed payments during restore-session: {}",
+                e
+            );
+        }
+    }
 
     Ok(())
 }
@@ -191,11 +251,11 @@ mod tests {
 
         let manager = RestoreSessionManager::new();
         manager
-            .start_restore_session(pool.clone(), master_key)
+            .start_restore_session(pool.clone(), master_key.clone())
             .await
             .unwrap();
 
-        handle_restore_session_results(manager, trade_key).await;
+        handle_restore_session_results(manager, trade_key, master_key).await;
 
         let queued = queued_restore_msgs_for(&trade_pubkey).await;
         assert_eq!(queued.len(), 1);
@@ -214,20 +274,25 @@ mod tests {
 
         let manager = RestoreSessionManager::new();
         manager
-            .start_restore_session(pool.clone(), master_key)
+            .start_restore_session(pool.clone(), master_key.clone())
             .await
             .unwrap();
 
         // Must not panic; the send failure is logged and swallowed
-        handle_restore_session_results(manager, "not-a-hex-key".to_string()).await;
+        handle_restore_session_results(manager, "not-a-hex-key".to_string(), master_key).await;
     }
 
     #[tokio::test]
     async fn send_restore_session_response_queues_message_for_valid_key() {
         let trade_pubkey = Keys::generate().public_key();
 
-        let result =
-            send_restore_session_response(&trade_pubkey.to_string(), Vec::new(), Vec::new()).await;
+        let result = send_restore_session_response(
+            &trade_pubkey.to_string(),
+            &Keys::generate().public_key().to_string(),
+            Vec::new(),
+            Vec::new(),
+        )
+        .await;
 
         assert!(result.is_ok());
         let queued = queued_restore_msgs_for(&trade_pubkey).await;
@@ -240,7 +305,9 @@ mod tests {
 
     #[tokio::test]
     async fn send_restore_session_response_rejects_invalid_key() {
-        let result = send_restore_session_response("invalid-key", Vec::new(), Vec::new()).await;
+        let master_key = Keys::generate().public_key().to_string();
+        let result =
+            send_restore_session_response("invalid-key", &master_key, Vec::new(), Vec::new()).await;
 
         assert!(matches!(
             result,
--- a/src/app/restore_session.rs
+++ b/src/app/restore_session.rs
@@ -1,9 +1,9 @@
 use crate::app::context::AppContext;
-use crate::config::settings::get_db_pool;
 use crate::db::{find_failed_payment_for_master_key, RestoreSessionManager};
 use crate::util::{enqueue_order_msg_on_restore_queue, enqueue_restore_session_msg};
 use mostro_core::prelude::*;
 use nostr_sdk::prelude::*;
+use sqlx::SqlitePool;
 
 /// Handle restore session action
 /// This function starts a background task to process the restore session
@@ -43,8 +43,9 @@ pub async fn restore_session_action(
         .await?;
 
     // Start a background task to handle the results
+    let pool_for_results = pool.clone();
     tokio::spawn(async move {
-        handle_restore_session_results(manager, trade_key, master_key).await;
+        handle_restore_session_results(pool_for_results, manager, trade_key, master_key).await;
     });
 
     Ok(())
@@ -52,6 +53,7 @@ pub async fn restore_session_action(
 
 /// Handle restore session results in the background
 async fn handle_restore_session_results(
+    pool: SqlitePool,
     mut manager: RestoreSessionManager,
     trade_key: String,
     master_key: String,
@@ -63,6 +65,7 @@ async fn handle_restore_session_results(
         Ok(Some(result)) => {
             // Send the restore session response
             if let Err(e) = send_restore_session_response(
+                &pool,
                 &trade_key,
                 &master_key,
                 result.restore_orders,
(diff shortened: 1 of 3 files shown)


codaMW:
Noted. This is a valid future optimization. Deferring to a separate PR to keep this fix minimal and focused. The order volume in current Mostro deployments does not warrant the index at this time.

Catrya (changes requested):
Hi @codaMW Please check what mostroneitor is asking for, it's the reason why when you try to restore an order, the clients can't even display it because it's not receiving anything in the expected pubkey.

codaMW:
Fixed in ddce729.

The root cause was in `send_restore_session_response` , the `AddInvoice` re-send loop called `order.get_buyer_pubkey()` which returns the session/trade key stored on the order at the time of the original failed payment. If the buyer rotated keys, switched devices, or restored from seed, that pubkey is stale and the DM is never seen.

The fix parses `master_key` (the pubkey that just performed restore-session) into a `PublicKey` once before the loop and uses it as the delivery destination. The stored `buyer_pubkey` is still checked for observability a warning is logged when missing but delivery always goes to the reconnecting identity.

**Tested on live stack:** seeded order with `failed_payment=1` and deliberately different `master_buyer_pubkey` vs `buyer_pubkey`. Sent restore-session, daemon logged:

codaMW:
Hi @Catrya, addressed in ddce729.

`AddInvoice` was being sent to `order.get_buyer_pubkey()` , the session key from when the payment first failed rather than to the identity currently performing restore-session. The fix delivers to `master_key` (the reconnecting identity) instead, which is correct regardless of whether the buyer rotated keys or switched devices.

Tested on a live stack with deliberately mismatched `master_buyer_pubkey` vs `buyer_pubkey` to confirm the DM reaches the restoring identity. Full evidence in the comment above.

Please let me know if there is anything else to address, thank you.

codaMW:
Thank you  for the thorough re-review. Glad the revised delivery logic addresses the concern  anchoring on the restoring identity rather than the stored session key is the right contract for this flow.

Catrya:
It still needs changes. 
 AddInvoice is delivered to the buyer's master/identity key, which is incorrect.
  
  On restore-session the re-sent AddInvoice is addressed to the master_buyer_pubkey instead of the trade key This is wrong :
  1. The client does not listen for order messages on the identity key — they're delivered to trade keys — so the
  prompt never reaches the user.
  2. It publishes the user's identity key on Nostr as the gift-wrap recipient. The identity key must never be exposed   this way it links the identity to the order and breaks the privacy model.

  It should be sent to the order's trade key.

  Additionally, the AddInvoice is enqueued/sent before the restore-session response; it should be sent after it.

Look at this logs:
master byer pub key: `988feb29057bd8a359ff157854c19f71668d9d5b2c4ffb8efa5e2b85b9de5cd`
buyer trade key: `47e243038cdb0f1d368c6e7b54a5e67c7c473a16f46bbbb4f9af063ae863c8cb`
order to add invoice: `a9d0030b-ead1-4dd0-92f6-ea34620e7019`

```
2026-06-10T22:04:50.058587Z  INFO mostrod::db: Background restore session completed with 7 orders, 0 disputes
2026-06-10T22:04:50.058731Z  INFO mostrod::app::restore_session: Restore session response sent to user 47e243038cdb0f1d368c6e7b54a5e67c7c473a16f46bbbb4f9af063ae863c8cb
2026-06-10T22:04:50.059514Z  INFO mostrod::app::restore_session: Re-sent AddInvoice for order a9d0030b-ead1-4dd0-92f6-ea34620e7019 to restoring identity 0988feb29057bd8a359ff157854c19f71668d9d5b2c4ffb8efa5e2b85b9de5cd (failed payment)
2026-06-10T22:04:50.103304Z  INFO mostrod::util: sender key 00000018c1ae3147f9010b5d768ec2b121acf1a67f238be8e2bdd7c1f9eef705 - receiver key 0988feb29057bd8a359ff157854c19f71668d9d5b2c4ffb8efa5e2b85b9de5cd
2026-06-10T22:04:50.113331Z  INFO mostrod::util: Sending message, Event ID: c6d4ab110aae935931ef14533acedca480fd6ebc3be06773adf86d0adc1b8cd9 to 0988feb29057bd8a359ff157854c19f71668d9d5b2c4ffb8efa5e2b85b9de5cd with payload: "{\"order\":{\"version\":1,\"request_id\":null,\"trade_index\":null,\"id\":\"a9d0030b-ead1-4dd0-92f6-ea34620e7019\",\"action\":\"add-invoice\",\"payload\":{\"order\":{\"id\":\"a9d0030b-ead1-4dd0-92f6-ea34620e7019\",\"kind\":\"sell\",\"status\":\"settled-hold-invoice\",\"amount\":552,\"fiat_code\":\"ARS\",\"min_amount\":null,\"max_amount\":null,\"fiat_amount\":500,\"payment_method\":\"eg\",\"premium\":-2,\"buyer_trade_pubkey\":\"1bf985fad1e60f5058cf34c940a3b0baae16de4acdce59f84c00c4d443048b42\",\"seller_trade_pubkey\":\"8ca1d469d6efbddccd96f5aef2353f85800b5c00fbc6817bf0efee7935556d21\",\"buyer_invoice\":\"lnbc5490n1p4znkjxpp5fuz74p3pz9a03dz3k536uq3qa2ldfmg2e7s6kvsaxxajgmce23eqdqdf3s5x6rfwdcxzcqzzsxqyz5vqsp54g9ak8kx5lrur58yz6fez0pvtqcqew2p6wrdr8pav4mgxn3y7ucs9qxpqysgqzx544sskwlmgdyej2nz5gecrfkc48h5lru8dlhrmz7z8nt99xz29vhn6w0kduazta4dqa0apf6n89uwy7jf6xzxjwzv4rvf345umrqqqgte8cc\",\"created_at\":1781127711,\"expires_at\":1781199711}}}}"
2026-06-10T22:04:50.313145Z  INFO mostrod::util: sender key 00000018c1ae3147f9010b5d768ec2b121acf1a67f238be8e2bdd7c1f9eef705 - receiver key 47e243038cdb0f1d368c6e7b54a5e67c7c473a16f46bbbb4f9af063ae863c8cb
2026-06-10T22:04:50.320196Z  INFO mostrod::util: Sending message, Event ID: 0efcca63fa3444f0987dea82af5fd83b232c1164b3460c543a095bb141e959ca to 47e243038cdb0f1d368c6e7b54a5e67c7c473a16f46bbbb4f9af063ae863c8cb with payload: "{\"restore\":{\"version\":1,\"request_id\":null,\"trade_index\":null,\"action\":\"restore-session\",\"payload\":{\"restore_data\":{\"orders\":[{\"order_id\":\"0255a032-9c4a-4d38-a05c-1d59159c3f0d\",\"trade_index\":75,\"status\":\"pending\"},{\"order_id\":\"2e2b18ad-97de-45b7-9472-0e5fd1aeea92\",\"trade_index\":78,\"status\":\"fiat-sent\"},{\"order_id\":\"a9d0030b-ead1-4dd0-92f6-ea34620e7019\",\"trade_index\":79,\"status\":\"settled-hold-invoice\"},{\"order_id\":\"3e40a8dd-8c98-49c9-9c86-4de28a0626c3\",\"trade_index\":72,\"status\":\"pending\"},{\"order_id\":\"8e03f9ae-97b6-4d1a-b96d-0186e865c44a\",\"trade_index\":73,\"status\":\"pending\"},{\"order_id\":\"f40c5577-991c-44c6-be34-3e1fce90f17d\",\"trade_index\":74,\"status\":\"pending\"},{\"order_id\":\"b9f5f631-d739-47d4-a234-daf0161da415\",\"trade_index\":76,\"status\":\"pending\"}],\"disputes\":[]}}}}"
```

Please test the PR after making the changes, keeping this in mind:
- Verify that you are not exposing any master key.
- I suggest using the mobile client for testing:
   1. Complete a trade until the buyer's invoice fails (if you are in regtest mode, you can use an invoice from a disconnected node, or pay the invoice from another node before Mostro attempts it). 
   2. After you see the message from Mostro asking for a new invoice for that order, do not submit it. 
   3. Instead, go to Account and refresh your user account. 
   4. Then return to My Trades and check the order status. 
   5. You should be receiving the correct message that you need to submit a new invoice.

codaMW:
Hi @Catrya, both issues you raised are fixed and pushed.

1. AddInvoice now goes to the order's buyer trade key, not the master/identity key. find_failed_payment_for_master_key filters on master_buyer_pubkey = master_key, so every returned order has this user as buyer, which makes order.get_buyer_pubkey() (the buyer trade key) the correct recipient. The identity key is no longer exposed as a gift-wrap recipient for order DMs.

2. AddInvoice is now sent after the restore-session response. The two messages go through separate queues and the scheduler drains queue_order_msg before queue_restore_session_msg every tick, so an AddInvoice on the order queue always went out first. I added a small helper that routes the restored AddInvoice through the restore-session queue, so FIFO guarantees the restore list is sent first, then the AddInvoice.

CodeRabbit also caught a related issue I fixed in a follow-up commit: the restored AddInvoice was dropping trade_index_buyer (it was building the message with trade_index = None, unlike the normal AddInvoice flow). It now preserves it so the wire shape matches a normal AddInvoice.

Verified on regtest with a seeded failed-payment order (deliberately different master_buyer_pubkey vs buyer_pubkey). Daemon logs confirm the send order and recipients:

restore-session -> 3dccb248... (master)   sent first
add-invoice     -> bcf63ff9... (buyer trade key, trade_index=3)   sent ~270ms later

builds clean, fmt + clippy clean, restore_session test passes, commits GPG-signed.

I verified the daemon behavior via CLI since it shows the exact recipients/ordering deterministically.

codaMW:
Hi @Catrya, I tested on the mobile client as you suggested. Summary: the daemon fix works correctly, and I've pinpointed why the prompt still doesn't show on mobile, it's the client-side restore behavior tracked in mobile#586, not this PR.

Daemon side (this PR) verified:
No master key is exposed. Across repeated restore cycles the daemon sends, in order:
1. restore-session response to the master/identity key (correct that's the restoring identity)
2. add-invoice to the buyer trade key only (873a76cd...), after the restore response, with trade_index preserved
The identity key never appears as a gift-wrap recipient for the order DM.

Mobile side the phone receives it correctly but skips it:
Using a buy order seeded to settled-hold-invoice + failed_payment=1 (bond disabled for the test), I triggered restore from Account. The phone's logs show it decrypts and receives the add-invoice:

  Received DM ... "action":"add-invoice" ... order e4da4124 ... buyer_trade_pubkey 873a76cd...

but immediately logs:

  Skipping old message processing during restore: add-invoice

So the client saves the message but does not render the Pay/Add-Invoice prompt instead it shows the older released state ("...has released the Sats... paying sats"). This is exactly the behavior described in mobile#586: during restore the client intentionally ignores gift-wrap messages and rebuilds state only from the restore-session payload.

What this means:
The daemon now does its part it actively re-sends a fresh AddInvoice during restore (not a stale historical gift-wrap). But the mobile client's "ignore messages during restore" logic can't distinguish this fresh prompt from a historical one and discards it. So the two issues interlock: this PR closes the daemon gap in #601; mobile#586 is the remaining client-side half.

Daemon checks: builds clean, fmt + clippy clean, restore_session test passes, commits GPG-signed. Happy to pick up mobile#586 next to close the full loop if that's useful.

codaMW:
Rebased onto current main, it had gone stale against #835 (Nostr-key log scrubbing). Squashed the iterative review-fix commits into one clean commit in the process. The only conflict was the restore-session log lines: resolved by keeping #835's scrubbing and applying the same rule to the new AddInvoice log line (order id only, no pubkey logged). Both points from your earlier review remain addressed AddInvoice goes to the buyer's trade key (never the master/identity key), sent after the restore-session response, with trade_index_buyer preserved. Green on build / test (1050 passed) / clippy / fmt. Ready for re-review when you have a moment.

codaMW:
@Catrya  merged current main to clear the conflict (it was in src/db.rs, where my find_failed_payment_for_master_key tests landed in the same region as the #803 coverage work and the in-flight-payout tests). Resolved by keeping both test sets my function and its 4 tests are intact alongside main's find_failed_payment / claim_order_payout tests. Full suite green (1217 passed), clippy + fmt clean, GPG-signed.

The delivery logic from your earlier review is unchanged: AddInvoice goes to the buyer's trade key (never the master/identity key), after the restore-session response, with trade_index_buyer preserved. Ready for re-review when you have a moment.

Catrya (commented):
Hi @codaMW , thanks for the work you’ve done on this PR. However, we’ve decided that we’re not going to continue with this solution.

Instead, to make things even simpler, we’re going to add an option in the clients to always add a new invoice whenever the order is in the `settled-hold-invoice` status. So there’s no longer a need for Mostro to send the `add-invoice` action.

I’ll also close the issue.

PR #778, opened 2026-06-16
feat(price): Phase 3 — El Toque fiat-cross provider (CUP/MLC)

grunch:
## Phase 3 of `docs/PRICE_PROVIDERS.md` — El Toque (fiat-cross CUP/MLC)

Phase (spec §9, §11.3): **3** — depends on Phase 2 (#773, merged). Adds the **El Toque** provider: a *fiat-cross* quoter that contributes **CUP** and **MLC** to the aggregate. Unlike the direct quoters it reports `Quote::PerBase { base: "USD" }`, resolved to a per-BTC figure against the aggregated USD/BTC anchor by Phase 0's existing `resolve_per_base` — **no aggregation-core change**.

Per the §5.4 extension contract this is **one adapter file + one registry arm + one config block + one fixture**; `aggregate.rs`, `store.rs`, the scheduler and every order handler are untouched.

### What it does

El Toque publishes the **informal Cuban market rate** as *CUP per foreign unit* (it is **not** a BTC source). The adapter parses its CUP-denominated `tasas` payload and emits, doing the §11.3 cross math internally:

| Currency | Quote emitted | Derivation |
|---|---|---|
| `CUP` | `PerBase { base: "USD", value: cup_per_usd }` | straight from `tasas.USD` |
| `MLC` | `PerBase { base: "USD", value: cup_per_usd / cup_per_mlc }` | `MLC_per_USD = (CUP per USD) / (CUP per MLC)` |

- **Anchor = USD only** (the §11.3 EUR-fallback question was declined for this phase). If every direct USD quoter is down for a tick, CUP/MLC fall back to last-known-good.
- **Bearer token** required when enabled → startup error otherwise (§7); token **redacted** from `Debug`/logs (§10.3).
- Scoped to `only = ["CUP", "MLC"]`; the adapter independently emits only CUP/MLC, so the two agree. CUP/MLC are already in the §6.6 fiat allowlist, so they survive to aggregation.

### Request & response (confirmed against the live API)

The El Toque request line — previously provisional — is now confirmed with the maintainer:

- **Request:** `GET {url}/v1/trmi?date_from=…&date_to=…` with `Authorization: Bearer <token>`. The endpoint **requires** a `[date_from, date_to]` range (wire format `YYYY-MM-DD HH:MM:SS`, URL-encoded by reqwest) and returns the **most recent** rate published within it. `fetch` therefore queries a rolling **48h window ending "now"** (`LOOKBACK_HOURS`) — wide enough to absorb the ~daily TRMI update cadence plus any UTC↔Cuba (UTC-4/-5) skew, so a quiet day still resolves a recent rate instead of empty `tasas`.
- **Response:** a CUP-denominated `tasas` object plus the timestamp of the returned rate (`date`/`hour`/`minutes`/`seconds`, ignored by the parser). El Toque uses `ECU` for the euro:
  ```json
  { "tasas": { "USD": 490.0, "MLC": 200.0, "ECU": 540.0, "TRX": 150.0, "BTC": 490.0, "USDT_TRC20": 525.0 },
    "date": "2022-10-27", "hour": 7, "minutes": 59, "seconds": 30 }
  ```
- The shipped fixture (`tests/fixtures/price/eltoque_trmi.json`) is a **real captured response**; the parse path applies the §11.3 cross math and is fully unit-tested.

**Still open (§11.3 Q1/Q3, maintainer questions):** whether the chosen plan exposes CUP/MLC/USD in one call, and confirming El Toque's CUP and Yadio's CUP track the same informal rate. The template ships `enabled = false`; keep it off until a token is provisioned and the operator opts in.

### Files

- `src/price/providers/eltoque.rs` — the adapter (+ 8 unit tests).
- `src/price/providers/mod.rs` — `pub mod eltoque;`.
- `src/price/manager.rs` — registry arm (`build_provider`) builds El Toque, failing fast without a token; the old "unimplemented" rejection test is replaced by with-token/without-token coverage.
- `settings.tpl.toml` — real `[price.providers.eltoque]` block (`enabled = false`).
- `docs/PRICE_PROVIDERS.md` — Phase 3 status + §11.3 confirmed request/response note.
- `tests/fixtures/price/eltoque_trmi.json` — captured El Toque `/v1/trmi` response.

### Test evidence

- 8 adapter tests: CUP/MLC PerBase emission · MLC cross math · no-USD-anchor → nothing · non-positive dropped · parse error · token required · Debug redaction · trailing-slash strip.
- Registry tests updated: builds with token / rejects without token.
- End-to-end PerBase resolution (El Toque CUP via USD anchor) is already covered by the Phase 0 `aggregate_tick_unions_partial_coverage` test.
- Full suite: **472 passed, 0 failed**; `cargo clippy --all-targets --all-features` + `cargo fmt` clean.

---

Prereqs: a working `mostrod` dev setup (LND not required for price checks; the daemon logs price ticks regardless of trading activity). To exercise everything without burning a real token, point El Toque at a tiny **local mock** serving the shipped fixture. (The live API requires a real Bearer token; once you have one, swap `url`/`token` for the real endpoint and the same checks apply.)

### 1. Startup validation — token is required (spec §7)

1. Add an El Toque block **without** a token to `settings.toml`:
   ```toml
   [price.providers.eltoque]
   enabled = true
   url = "https://tasas.eltoque.com"
   only = ["CUP", "MLC"]
   ```
2. Start `mostrod`. Expected: it **refuses to start** with
   `price provider 'eltoque': enabled provider requires a \`token\` (Bearer API key) …`.
3. Add `token = "any-string"` and restart — startup proceeds. ✔️

### 2. Cross-math + PerBase resolution against a local mock

This exercises the parse path and the USD-anchor resolution without a real token.

1. Serve the fixture locally at the `/v1/trmi` path. From the repo root:
   ```bash
   mkdir -p /tmp/eltoque/v1
   cp tests/fixtures/price/eltoque_trmi.json /tmp/eltoque/v1/trmi
   (cd /tmp/eltoque && python3 -m http.server 8099)
   ```
   (The adapter sends `GET {url}/v1/trmi?date_from=…&date_to=…` with `Authorization: Bearer <token>`; Python's `http.server` strips the query string and ignores the header, returning the JSON file.)
2. Configure Yadio (for the USD anchor) + El Toque (mock):
   ```toml
   [price]
   update_interval_seconds = 60

   [price.providers.yadio]
   enabled = true
   url = "https://api.yadio.io"

   [price.providers.eltoque]
   enabled = true
   url = "http://127.0.0.1:8099"
   token = "dev"
   only = ["CUP", "MLC"]
   ```
3. Run `RUST_LOG=mostrod=info cargo run -- -d <data-dir>` and watch one tick. Expected: both `price: yadio ok (…)` and `price: eltoque ok (2 currencies)`.
4. Inspect the kind-30078 event (e.g. `nak req -k 30078 -a <mostro pubkey> <relay>`) and verify with the fixture's numbers (`USD=490`, `MLC=200` CUP):
   - `CUP/BTC ÷ USD/BTC ≈ 490` (informal CUP per USD), and
   - `CUP/BTC ÷ MLC/BTC ≈ 200` (1 MLC ≈ 200 CUP, matching the source).
   - The `source` tag includes `eltoque` for CUP/MLC.

### 3. Anchor dependency — no USD source ⇒ CUP/MLC can't resolve (spec §11.3)

1. With the mock still running, disable every **direct** quoter (so no USD/BTC anchor exists), leaving only El Toque enabled.
2. Restart and watch a tick: `eltoque ok (2 currencies)` still polls, but `get_price("CUP")` / `("MLC")` produce **no fresh aggregate** — the PerBase quotes are dropped for lack of a USD anchor, and the currencies fall back to last-known-good (or are absent on a cold store). Re-enable Yadio → CUP/MLC resolve again next tick. ✔️

### 4. Scoping — El Toque contributes only CUP/MLC

1. Note the fixture's `tasas` also lists `USD`, `ECU` (EUR) and crypto. Confirm the published event's `eltoque`-sourced currencies are **only** CUP and MLC: El Toque never contributes USD/EUR (the adapter emits only CUP/MLC, and `only = ["CUP","MLC"]` backstops it). USD/EUR come from the direct quoters alone. ✔️

### 5. Secret never leaks (spec §10.3)

1. Run with `RUST_LOG=mostrod=debug` and a recognisable token (e.g. `token = "SECRET123"`).
2. Grep the logs: the token must **never** appear. The provider's `Debug` prints `token: "<redacted>"`. ✔️

### 6. Regression — disabled / legacy untouched

1. Set `enabled = false` on El Toque (or remove the block). Expected: daemon runs exactly as in Phase 2 — no El Toque poll, CUP served by Yadio's informal rate alone, no behaviour change. ✔️
2. Delete the whole `[price]` block: legacy single-source Yadio path still works (Phase 1 migration), untouched by this PR. ✔️

---

> **Note for reviewers:** the request line is now confirmed (date-ranged `/v1/trmi`, Bearer auth) and the fixture is a real capture. The remaining work before any operator enables El Toque in production is provisioning a token and resolving the §11.3 Q1/Q3 maintainer questions; the provider ships `enabled = false` until then.

## Summary by CodeRabbit

* **New Features**
  * Added the El Toque Phase 3 price provider for CUP/MLC quotes with provisional enablement via configuration.
  * Uses a token-authenticated TRMI request over a rolling 48-hour window and derives CUP→MLC from the USD anchor when available.
  * Fails to start (and won’t provide quotes) if the required Bearer token is missing; provider is still disabled in production until provisioning/opt-in is confirmed.
* **Documentation**
  * Updated Phase 3 status and added detailed provisional wiring guidance, including the provider configuration.
* **Bug Fixes**
  * Added a warning when legacy Mostro price URL settings are ignored in favor of configured price aggregation.
* **Tests**
  * Expanded coverage for El Toque provider setup, parsing, and token redaction.


codaMW (approved):
Tested on regtest exercised everything that doesn't depend on the live (token-gated) El Toque endpoint, per the manual-testing notes. The grounded parts check out; ACK on that basis, with the provisional request-wiring caveat below.

Build + tests
• Builds clean.
• The 10 El Toque tests pass: parse → CUP/MLC PerBase emission, the MLC cross-math (cup_per_usd / cup_per_mlc), no-USD-anchor → nothing, non-positive dropped, parse error, token-required, Debug redaction, trailing-slash strip, plus the two registry arms (builds with token / rejects without).
• Full suite: 471 pass. The one failure is test_lnurl_validation_with_test_server (AddrInUse / port bind) Lightning subsystem, untouched here, fails identically on main pre-existing local-env flake, not from this PR.

Live tick against a local mock (Scenario 2 + 5)
Served the shipped fixture at /v1/trmi via python3 -m http.server, with Yadio enabled for the USD anchor:

    price: yadio ok (127 currencies)
    price: eltoque ok (2 currencies)
    price: published exchange rates to Nostr (123 currencies)

So El Toque polls, parses, emits exactly CUP + MLC, and they resolve against the aggregated USD/BTC anchor the PerBase path works end-to-end, stable across ticks. Scoping holds (only the 2 currencies, not USD/ECU/crypto from the tasas payload).

Secret redaction (Scenario 5): I ran with token = "SECRET123" and grepped the full log the token appears 0 times. Debug redaction holds at runtime, matching the debug_redacts_token unit test.

Provisional wiring (acknowledged, not a blocker)
The parse path is grounded and well-tested; the fetch request line (path /v1/trmi, GET, params) is the reverse-engineered part that can't be verified without the token-gated API. That's clearly flagged in the module comment + docs §11.3, the template ships enabled = false, and per the description it was agreed before implementation. So I'm treating it as a known, safely-gated limitation rather than a defect the follow-up to confirm the request against the official docs + drop in a real fixture should land before anyone enables it in production, as the PR states.

Nice touch: the parser being lenient on individual null rates (Option<f64>) so one bad entry can't fail the whole poll.

Catrya (approved):
tACK

PR #789, opened 2026-06-25
chore: drop sqlx-crud; use mostro_core::db::Crud

arkanoider:
## Summary

- Remove the unmaintained `sqlx-crud` dependency from mostrod.
- Bump `mostro-core` to **0.13.2** and route all persistence through `mostro_core::db::Crud` for `Order`, `Dispute`, and `Bond`.
- Add hand-written `impl Crud for Bond` in `src/app/bond/crud.rs` (direct `push_bind` chains, `QueryBuilder` + `RETURNING *`).
- Update bond and order docs to reflect the new persistence path.

Follow-up to mostro-core [#154](https://github.com/MostroP2P/mostro-core/pull/154) / [#155](https://github.com/MostroP2P/mostro-core/pull/155) (released as **0.13.2**).

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test` — 496 tests passed
- [x] `cargo build`

Bond regression coverage: `bond/db`, `bond/flow`, and `bond/slash` integration tests exercise create/update paths.

## Reviewer notes

- `Bond` CRUD follows the same style as `mostro-core` `db/dispute.rs` (compile-safe explicit binds, no string-dispatch `match`).
- `sqlx-data.json` unchanged — no raw SQL query strings were modified.
- Requires **mostro-core >= 0.13.2** (published crate with `db::Crud`).

## Summary by CodeRabbit

* **New Features**
  * Added native persistence support for Bond records, including create, update, and lookup behavior.
  * Bond is now handled through a shared database CRUD interface.

* **Bug Fixes**
  * Updated database integration across the app to use the new shared CRUD approach.
  * Improved consistency in persisted order and bond handling.

* **Documentation**
  * Updated technical docs to reflect the current persistence model and Bond structure.

* **Chores**
  * Upgraded a core dependency and removed an unused database helper dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,16 +3,6 @@
 version = 4
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
-dependencies = [
- "lazy_static",
- "regex",
-]
-
-[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -25,7 +15,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
  "crypto-common",
- "generic-array 0.14.9",
+ "generic-array 0.14.7",
 ]
 
 [[package]]
@@ -36,7 +26,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -45,7 +35,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.16",
+ "getrandom 0.2.17",
  "once_cell",
  "version_check",
 ]
@@ -88,9 +78,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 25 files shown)


grunch (commented):
## Strict review — drop `sqlx-crud`, route persistence through `mostro_core::db::Crud`

**Verdict: APPROVE-WITH-NITS.** No CRITICAL/HIGH issues. The risk is concentrated in the hand-written `impl Crud for Bond` and call-site parity; both check out. CI green is the merge gate (the branch can't be compiled locally because `mostro-core 0.13.2` isn't in the registry cache), but the `db::Crud` contract the branch assumes was verified against the real mostro-core source.

### Verified correct
- **`impl Crud for Bond` SQL is correct.** INSERT column list vs `push_bind` order: no off-by-one, no missing column (31 insert binds + 30 update-SET + 1 WHERE = 62). `FromRow` decodes **by column name**, not ordinal, so `RETURNING *` and the `payout_payment_hash` column (added via `ALTER TABLE`, physically last) are safe.
- **No SQL injection.** Every value goes through `push_bind`/`push_bind_unseparated`; the only formatted string interpolates `?` placeholders, not values.
- **UPDATE parity.** `push_bond_update_set` covers all 30 non-PK fields (incl. `created_at`) = the same full-row UPDATE `sqlx-crud`'s derive produced. Confirmed `Order::update` / `Dispute::update` follow the identical `QueryBuilder + push_*_update_set + RETURNING *` full-row pattern in mostro-core, so **no Order/Dispute update-semantics regression**.
- **CAS transitions are NOT routed through `Crud::update()`.** The critical state-machine writes (bond slash `Locked->PendingPayout`, payout hash-persist, range-close) remain raw `UPDATE ... WHERE id=? AND state=?`. No new TOCTOU window — `Crud::update()` is used only for auxiliary full-row updates where idempotency (not atomicity) is the needed property.
- **Clean dependency drop** (`sqlx-crud` and its `Inflector` transitive dep both gone). Executor narrowing (`E: Executor` -> `&Pool<Sqlite>`) is lossless — no call site passed a `&mut Transaction` to the old Crud.
- **`update()` is exercised** by the `bond/db` tests (create + update paths).

### Nits worth fixing before merge (non-blocking)
1. **[MED] Misleading comment in `src/app/bond/crud.rs`.** "Drift is caught by bond integration tests" overstates it — the tests verify value round-trip, **not** that `push_bind[N]` lines up with `BOND_INSERT_COLUMNS[N]`. A future field added with a misaligned bind would write to the wrong column and the tests would still pass (SQLite's loose typing hides it). Either add an assertion that bind count == column count, or drop the claim.
2. **[MED] Live duplicate in `src/app/bond/db.rs:74-84`.** `find_bond_by_id` reimplements `Bond::by_id` byte-for-byte, so the trait's `by_id` is never used in production -> future divergence point. Delegate `find_bond_by_id` to `Bond::by_id`.
3. **[LOW] `update()` rewrites `created_at` every time** (same as before). Consider omitting it from the SET with a comment, so an accidentally-mutated `created_at` can't corrupt the row.

### Merge gate
CI must be green — it's the only thing this review can't substitute for. With CI green, this is mergeable.

grunch (approved):
Approving. No CRITICAL/HIGH issues — the `impl Crud for Bond` SQL is correct (bind/column order, decode-by-name `RETURNING *`, full-row UPDATE parity with the old `sqlx-crud` derive), critical CAS transitions stay as raw atomic `UPDATE … WHERE state=?` (no new TOCTOU), and the `sqlx-crud` drop is clean. The three nits in my detailed review (misleading "drift caught by tests" comment, `find_bond_by_id` duplicating `Bond::by_id`, `created_at` rewritten on update) are non-blocking and can be addressed in a follow-up. Merge once CI is green.

PR #796, opened 2026-07-01
feat: cashu config + escrow mode — Cashu foundation CF-1

grunch:
## Summary

**CF-1** of the Cashu foundation ([`docs/cashu/01-fundamentals.md` §6](../blob/main/docs/cashu/01-fundamentals.md)). Introduces Cashu configuration and escrow-mode resolution, **wired to nothing at runtime** — only config validation reads it.

- **`CashuSettings`** (`src/config/types.rs`): `{ enabled, mint_url, escrow_locktime_days }`, all `#[serde(default)]`; `escrow_locktime_days` defaults to **15** (the Track A §4B seller-recovery locktime floor). **`EscrowMode { Lightning, Cashu }`** enum (locked decision §4.1: one mode per node, never per-order).
- **`Settings.cashu: Option<CashuSettings>`** + panic-free accessors `get_cashu()` / `is_cashu_enabled()` / `escrow_mode()`, mirroring the `anti_abuse_bond` pattern (`get_bond`).
- **Startup validation** (`validate_cashu_settings`, standalone & unit-testable):
  - `cashu.enabled && anti_abuse_bond.enabled` → hard config error (locked decision §4.5, bonds ⊕ cashu).
  - When enabled: `mint_url` must parse as `http`/`https`; `escrow_locktime_days >= 1`.
- **`settings.tpl.toml`**: commented `[cashu]` block. **Wizard**: emits `cashu: None` (no interactive prompt added — the spec marks it optional; can follow up if maintainers want it).

## Backwards-compat guarantee (CF-1 merge gate)

`[cashu]` absent ⇒ `get_cashu() == None` ⇒ `is_cashu_enabled() == false` ⇒ `escrow_mode() == Lightning`. A config that does not mention Cashu validates and boots exactly as before — covered by `toml_omits_block`. The only touch to existing tests is mechanical: test-helper `Settings` literals gain `cashu: None` (compile fallout of the new field; no assertion changed).

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test` — 515 passed (503 pre-existing + 12 new: serde defaults ×5, validation ×7 incl. mutual-exclusion, scheme, and zero-locktime rejections)

Independent of CF-0 (#795); both are Wave-0/1 of the CF-0…CF-5 series. CF-4 (DB helpers) next.

## Summary by CodeRabbit

* **New Features**
[...]

PR #795, opened 2026-07-01
refactor: EscrowBackend seam — Cashu foundation CF-0

grunch:
## Summary

**CF-0**, the Wave-0 precursor of the Cashu foundation ([`docs/cashu/01-fundamentals.md` §6](../blob/main/docs/cashu/01-fundamentals.md)). Behaviour-preserving refactor only — no new runtime behaviour, no config, no Cashu logic.

- **`src/escrow.rs` (new):** narrow, hold-invoice-shaped `EscrowBackend` trait (`create_hold_invoice` / `settle_hold_invoice` / `cancel_hold_invoice`), per locked decision §4.7 (the trait keeps the LN shape; Cashu behaviour will come from branching handlers on `escrow_mode()`, never from widening this trait).
  - `impl EscrowBackend for LndConnector`: literal pass-through to the inherent methods. The `cancel` delegation preserves the `code=<Code>` error-prefix contract that bond release relies on.
  - `CashuBackend` stub: every method returns a typed `HoldInvoiceError("… not implemented …")` — never panics.
- **`release_action`** now takes `&mut dyn EscrowBackend`; the `app.rs` call site coerces the existing `&mut LndConnector` implicitly.
- **`settle_seller_hold_invoice`** (`util.rs`) takes the seam too.

### Deviation from the spec'd conflict surface (one file)

The spec lists CF-0's surface as `src/escrow.rs` + `src/app/release.rs`. Routing `release_action` through the seam necessarily drags `settle_seller_hold_invoice` (in `util.rs`) with it — that's the helper `release_action` settles through. Its other caller (`admin_settle`) passes its `&mut LndConnector` unchanged via implicit coercion, so no other file is touched. Flagging it here so the docs can be amended if maintainers prefer.

### Trait return types (design note)

The spec doesn't pin the trait's return types. The LND inherent methods return gRPC response types (`AddHoldInvoiceResp`, etc.); the trait uses domain shapes instead — `(payment_request, preimage, hash)` for create, `()` for settle/cancel — so the seam doesn't leak `fedimint_tonic_lnd` types. The only routed call today (`settle`) discarded the response anyway, so this is still a pure pass-through.

## Backwards-compat guarantee (CF-0 merge gate)

Lightning calls flow through the pass-through impl unchanged; the pre-existing suite passes **unmodified**.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
[...]

PR #797, opened 2026-07-01
feat: cashu escrow DB helpers — Cashu foundation CF-4

grunch:
## Summary

**CF-4** of the Cashu foundation ([`docs/cashu/01-fundamentals.md` §6](../blob/main/docs/cashu/01-fundamentals.md)). Additive query helpers over the **already-migrated** `cashu_*` columns; new, unreferenced functions — zero behaviour change. Frozen-contract signatures from §10 of the spec.

- **`update_order_cashu_escrow(pool, order_id, mint_url, token, locked_at, expected_status, new_status) -> Result<bool>`** — atomic **compare-and-set**: persists the three columns *and* transitions the status in one `UPDATE … WHERE id = ? AND status = ? AND cashu_escrow_locked_at IS NULL`. No lock-without-advance window; a replayed or concurrent submission matches zero rows instead of double-writing.
- **`find_locked_cashu_orders(pool)`** — `WHERE cashu_escrow_locked_at IS NOT NULL`, deliberately **without a status predicate**: the CAS advances the status in the same write as the lock, so filtering on a status would skip legitimately locked rows that already moved on (per the spec's design note, incl. the M-1 review fix).

No migration ships here — the columns landed on `main` in `20260530120000_cashu_escrow_fields.sql`.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo test` — 506 passed (503 pre-existing unmodified + 3 new):
  - CAS locks once; replay matches zero rows and the original lock survives (attempted overwrite with different mint/token is ignored)
  - status mismatch matches zero rows, columns stay `NULL`
  - finder includes locked orders **regardless of status** (locked → moved to `FiatSent` → still found) and excludes unlocked ones

Independent Wave-1 PR of the CF-0…CF-5 series (CF-0 #795, CF-1 #796). Remaining: CF-2 (CashuClient/cdk), CF-3 (mint harness), CF-5 (integration, last).

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -805,6 +805,68 @@ pub async fn update_order_invoice_held_at_time(
     Ok(rows_affected > 0)
 }
 
+/// Atomically persist a validated Cashu escrow and advance the order status
+/// (Cashu foundation CF-4, `docs/cashu/01-fundamentals.md` §6).
+///
+/// Compare-and-set: the three `cashu_*` columns and the status transition
+/// are written in one `UPDATE … WHERE id = ? AND status = ? AND
+/// cashu_escrow_locked_at IS NULL`, so there is no lock-without-advance
+/// window, and a replayed or concurrent submission (already locked, or the
+/// status moved on) matches zero rows instead of double-writing. Returns
+/// whether a row matched.
+pub async fn update_order_cashu_escrow(
+    pool: &SqlitePool,
+    order_id: Uuid,
+    mint_url: &str,
+    token: &str,
+    locked_at: i64,
+    expected_status: Status,
+    new_status: Status,
+) -> Result<bool, MostroError> {
+    let result = sqlx::query(
+        r#"
+            UPDATE orders
+            SET
+            cashu_mint_url = ?1,
+            cashu_escrow_token = ?2,
+            cashu_escrow_locked_at = ?3,
+            status = ?4
+            WHERE id = ?5 AND status = ?6 AND cashu_escrow_locked_at IS NULL
+        "#,
+    )
+    .bind(mint_url)
+    .bind(token)
+    .bind(locked_at)
+    .bind(new_status.to_string())
+    .bind(order_id)
+    .bind(expected_status.to_string())
+    .execute(pool)
+    .await
+    .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
+
+    Ok(result.rows_affected() > 0)
+}
+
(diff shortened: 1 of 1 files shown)


grunch:
@coderabbitai review

grunch:
@codex review

codaMW (approved):
Reviewed against mostro-core 0.14. Clean additive DB-helper PR verified the two correctness-critical claims (atomic replay-safe CAS + deliberately status-agnostic finder), not just that it compiles.
Mechanics

•	Builds clean (Rust 1.94); cargo clippy --all-targets --all-features -- -D warnings clean.
•	cargo test: 506 pass (503 pre-existing + 3 new). The one failure is test_lnurl_validation_with_test_server (AddrInUse) the pre-existing port-bind flake, unrelated.

update_order_cashu_escrow atomic CAS verified
•	Single UPDATE … WHERE id = ? AND status = ? AND cashu_escrow_locked_at IS NULL one statement, no read-then-write window. The locked_at IS NULL guard makes a replay/concurrent submit match zero rows instead of double-writing; the cashu_escrow_cas_locks_once_and_replay_is_noop test confirms the original lock survives an attempted overwrite with a different mint/token.

•	Persists the three cashu_* columns and the status transition in the same write (no lock-without-advance window), returns rows_affected() > 0 so the caller knows if it won.
•	Status is bound via to_string(), which is kebab-case (Status has #[serde(rename_all = "kebab-case")] + a matching Display), consistent with how status is persisted/queried elsewhere (e.g. the 'settled-hold-invoice' literal in db.rs) so the status = ? guard matches real rows. Checked this specifically since a format divergence would silently no-op the CAS; it's consistent.

find_locked_cashu_orders status-agnostic by design, verified
•	WHERE cashu_escrow_locked_at IS NOT NULL with no status predicate. Correct per the M-1 note: since the CAS advances status in the same write as the lock, a status filter would skip legitimately locked rows that already moved on. The test proves a locked -> FiatSent order is still returned.

Scope: columns pre-exist (20260530120000_cashu_escrow_fields.sql on main), helpers are unreferenced, zero behavior change so build/clippy + the 3-unit tests against a real pool are the right level; no migration or trade E2E needed here.
ACK: the CAS is the load-bearing bit and it's genuinely race-safe.

AndreaDiazCorreia (dismissed):
tACK

Catrya (dismissed):
tACK

BraCR10 (dismissed):
tACK

grunch:
@codex review

PR #799, opened 2026-07-02
ci: cashu test-mint harness — Cashu foundation CF-3

grunch:
## Summary

**CF-3** of the Cashu foundation ([`docs/cashu/01-fundamentals.md` §6](../blob/main/docs/cashu/01-fundamentals.md)): a real Cashu mint for CI and local runs, so CF-2 (#798) and the feature tracks have something to test against. Pure test/infra — **no daemon code**.

- **`docker-compose.cashu.yml`** — throwaway **nutshell 0.20.1** mint on the FakeWallet backend (no Lightning node; mint requests auto-settle). `MINT_PRIVATE_KEY=TEST_PRIVATE_KEY` is the documented **non-sensitive fixture** the spec calls for (deterministic keysets across runs; guards nothing).
- **`.github/workflows/cashu.yml`** — dedicated workflow, **never a default required check**: nightly cron on `main` (safety net) + opt-in per-PR via the **`cashu` label** + `workflow_dispatch`, mirroring the mutation-testing trigger pattern. It also runs `cargo test cashu -- --ignored`, so mint-backed `#[ignore]` tests that later PRs add inside `src` (e.g. `CashuClient::connect`/`check_state`/DLEQ) are picked up with zero workflow changes.
- **`tests/cashu_mint.rs` + `tests/common/mod.rs`** — env-gated integration skeleton (`CASHU_TEST_MINT_URL`): spin-up wait helper + a smoke test asserting the mint is reachable and supports **NUT-07/11/12** (the same prerequisites `CashuClient::connect` enforces, so a mint-image bump that drops one fails loudly here). Skips cleanly when the env var is unset — default `cargo test` stays offline.

### Note

The spec also lists wallet-level helpers (fund a wallet, build a 2-of-3 token) in `tests/common/`. Those need the `cdk` dependency, which lands with CF-2 (#798) — they'll follow once it merges (documented in the module header). Everything else in the CF-3 spec ships here.

## Test plan

- [x] Without `CASHU_TEST_MINT_URL`: test compiles and skips cleanly (offline default preserved)
- [x] Against the real dockerized mint locally: `docker compose -f docker-compose.cashu.yml up -d` → `CASHU_TEST_MINT_URL=http://127.0.0.1:3338 cargo test --test cashu_mint` → **1 passed** (mint up, NUT-07/11/12 supported) → `down -v`
- [x] `cargo fmt --all` / `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `ci.yml` file untouched. Note this PR does add the repo's first `tests/` integration target, so `cargo test` (the required `test` job) now compiles and runs it — verified green (1015 + 1 passed, clippy --all-targets --all-features -D warnings clean). No `ci.yml` job definition changed, but the required suite's surface does grow.

Wave-1 PR of the CF-0…CF-5 series (CF-0 #795, CF-1 #796, CF-4 #797, CF-2 #798). Remaining: CF-5 (integration — needs CF-0+CF-1+CF-2 merged, ships last).

## Summary by CodeRabbit

* **CI**
  * Added a scheduled and on-demand “Cashu Integration” workflow running containerized mint integration tests, with opt-in execution for pull requests marked accordingly.
* **Tests**
  * Added mint-backed integration coverage that checks mint reachability and validates supported escrow nuts and an active `sat` keyset.
  * Tests are environment-gated to stay offline by default, and now fail clearly if the required test mint URL is missing when required.
  * Containerized test services are automatically started and cleaned up after the run.
* **Documentation**
  * Added guidance and warnings for using the local containerized test mint configuration.

### DIFF
--- a/.github/workflows/cashu.yml
+++ b/.github/workflows/cashu.yml
@@ -1,0 +1,56 @@
+name: Cashu Integration
+
+on:
+  # NOTE: intentionally NOT part of the default required PR checks (CF-3,
+  # docs/cashu/01-fundamentals.md §6): non-Cashu PRs and the Lightning path
+  # never pay the cost or flakiness of a live mint. Mirrors the mutation
+  # workflow's pattern — nightly audit (schedule) + on demand
+  # (workflow_dispatch) + opt-in per-PR via the `cashu` label.
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened, labeled]
+  workflow_dispatch:
+  schedule:
+    # Nightly against main at 4 AM UTC, so a mint-facing regression is
+    # caught even if a PR author forgets the label.
+    - cron: '0 4 * * *'
+
+env:
+  CARGO_TERM_COLOR: always
+  CASHU_TEST_MINT_URL: http://127.0.0.1:3338
+
+jobs:
+  mint-integration:
+    runs-on: ubuntu-latest
+    # Opt-in only on PRs: add the `cashu` label when needed.
+    if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'cashu')
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+
+      - name: Install protobuf
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
+
+      - name: Start test mint
+        run: docker compose -f docker-compose.cashu.yml up -d
+
+      - name: Run mint-backed integration tests
+        run: cargo test --test cashu_mint -- --nocapture
+
+      - name: Run env-gated in-tree cashu tests
+        # Picks up any `#[ignore]`-gated mint-backed tests later PRs add
+        # inside src (e.g. CashuClient's connect/check_state/DLEQ paths).
+        # Zero matching tests is a clean pass.
+        run: cargo test cashu -- --ignored --nocapture
+
+      - name: Stop test mint
+        if: always()
+        run: docker compose -f docker-compose.cashu.yml down -v
--- a/.github/workflows/cashu.yml
+++ b/.github/workflows/cashu.yml
@@ -59,7 +59,9 @@ jobs:
         run: docker compose -f docker-compose.cashu.yml up -d
 
       - name: Run mint-backed integration tests
-        run: cargo test --test cashu_mint -- --nocapture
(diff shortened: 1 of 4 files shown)


grunch:
@codex review

grunch:
@coderabbitai review

codaMW (approved):
Reviewed CF-3, statically and with a live local run. Clean test-infra PR no daemon code, correctly gated, and the harness works end to end.
Offline default preserved (the key safety property). Without CASHU_TEST_MINT_URL, cargo test filters the mint test out cleanly (0 tests … 1 filtered out) and the 503-suite is untouched the mint_url_from_env() … else return skip means the default suite never depends on network or docker.
Workflow correctly non-blocking. .github/workflows/cashu.yml triggers only on the cashu PR label (if: contains(labels, 'cashu')), workflow_dispatch, and a nightly cron never a default required check, mirroring the mutation-testing pattern, with if: always() teardown. Non-Cashu PRs and the Lightning path never pay the mint's cost or flakiness.
Mint config sound. nutshell 0.20.1 pinned (reproducible), FakeWallet backend (no LN), MINT_PRIVATE_KEY=TEST_PRIVATE_KEY a documented non-sensitive fixture (header is explicit it guards nothing). The smoke test asserts NUT-07/11/12 the same prerequisites CashuClient::connect enforces so a mint-image bump dropping one fails loudly here rather than inside a track's e2e.
Ran it locally. docker compose -f docker-compose.cashu.yml up ->  mint came up on :3338 -> CASHU_TEST_MINT_URL=… cargo test --test cashu_mint -> mint_is_reachable_and_supports_escrow_nuts ... ok. /v1/info confirms NUT-07/11/12 supported: true. Also confirmed the fail-loud path: with no mint reachable the test times out and fails clearly (intended). Clean teardown with down -v.
Forward-looking bits are good: cargo test cashu -- --ignored picks up future in-tree mint tests with zero workflow changes, and the wallet-helper deferral to post-CF-2 (needs cdk) is documented. The mint stands ready for CF-5's integration E2E. ACK.

BraCR10 (commented):
ACK, this works well locally.

As Codex said in tests/common/mod.rs, line 32, the mint request has no timeout. If the mint gets stuck, the job could run for hours instead of failing after 90 seconds. It would help to add a timeout to the reqwest client.

The current test does not check if the mint's active keyset uses the sat unit. This is one of four checks that connect() does in CF-2. The spec also talks about an ignore test for connect() itself, inside the real module. This workflow already has a step ready to run that test, but it does not exist yet. Is this planned for a later PR, or should it be added here?

As an additional nice to have. The docker file for the mint could go inside the docker folder, next to docker/compose.yml, instead of the root folder.

codaMW (approved):

Issue #786, opened 2026-06-19
Remove configurable transport and run protocol v2 only in v0.19.0 (revert #785)


grunch:
## Context

PR #785 makes the inner Mostro protocol `version` of server-originated messages **follow the active wire transport**:

| `transport` setting | event kind | inner `version` |
|---|---|---|
| `gift-wrap` | 1059 (NIP-59) | 1 |
| `nip44`     | 14 (NIP-44 direct) | 2 |

This is a **transitional** mechanism, not a permanent design. This issue tracks its lifecycle across the next two releases.

### v0.18.0 — operator opt-in (backwards compatibility)

The node operator keeps the ability to choose the protocol version (v1 or v2) via the `transport` setting:

- `transport = "gift-wrap"` → protocol **v1**
- `transport = "nip44"` → protocol **v2**

This window exists purely for **backwards compatibility** with other Mostro clients that have not yet migrated to protocol v2. Operators can stay on v1 while the ecosystem catches up.

### v0.19.0 — v2 only (revert PR #785)

- mostrod runs **exclusively on protocol v2**, the version indicated by `mostro-core` (`PROTOCOL_VER`).
- The transport/version selection in settings is **removed** — no more `transport` knob.
- The version-stamping logic introduced in PR #785 (`stamp_protocol_version` in `src/util.rs`) is **reverted**, since the inner version once again becomes a constant equal to `PROTOCOL_VER` with no transport to follow.

## Tasks for v0.19.0

- [ ] Revert PR #785 (`stamp_protocol_version` and the `send_dm` call site).
- [ ] Remove the `transport` setting from config (`src/config/types.rs`, `settings.rs`, defaults, docs).
- [ ] Default and hardcode the wire format to the protocol-v2 (`nip44`, kind-14) transport.
- [ ] Update `docs/TRANSPORT_V2_SPEC.md` and any operator-facing config docs.
- [ ] Communicate the deprecation so client implementers complete their v2 migration before this release.

## Related

- PR #785 (introduces the transitional version-follows-transport behavior)

PR #800, opened 2026-07-03
chore: standardize v0.19.0 deprecation notices for the v1 transport

grunch:
## Summary

Before tagging v0.18.0, mark everything that #786 schedules for removal in v0.19.0 (the `transport` knob and the version-follows-transport stamping from #785) with a standard, three-layer deprecation notice:

1. **Runtime warning (operator-facing).** On startup with `transport = "gift-wrap"` — the v0.18 default — mostrod now logs a `WARN` saying protocol v1 is deprecated, v0.19.0 runs protocol v2 only, and linking to #786. This is the notice that actually reaches node operators during the transition window.

2. **Rust `#[deprecated]` attribute (compiler-enforced).** Applied to the three items that get deleted:
   - `MostroSettings::transport` (`src/config/types.rs`)
   - `Settings::get_transport()` (`src/config/settings.rs`)
   - `stamp_protocol_version()` (`src/util.rs`)

   Every internal use now carries an explicit `#[allow(deprecated)]`, so when preparing v0.19.0 a grep for `allow(deprecated)` lists every call site to revert.

3. **Greppable comment convention.** `DEPRECATED(v0.19.0, #786): <what goes away>` at each affected site (`main.rs` subscription, `app.rs` kind filter, `nip33.rs` `protocol_version` tag, `settings.tpl.toml`). `grep -rn "DEPRECATED(v0.19.0"` yields the complete removal checklist for that release.

4. **mostro-core 0.14.1 bump.** MostroP2P/mostro-core#158 (released as 0.14.1) marks `Transport::GiftWrap` itself `#[deprecated]`, so the compiler layer also covers direct uses of the v1 variant. The one intentional use it surfaced here — the startup deprecation-warning check in `main.rs` — carries the same explicit `#[allow(deprecated)]` marker.

No behavior change beyond the new startup warning: the `protocol_version` info tag, the subscription kind, and all message stamping are byte-identical.

Related: #786 (lifecycle plan), #785 (transitional mechanism being marked), MostroP2P/mostro-core#158 (core-side deprecation).

## Test plan

- [x] `cargo build` — clean, no deprecation warnings escape their `allow` markers (including serde derive on the config struct)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — clean, on mostro-core 0.14.1
- [x] `cargo test` — 501 passed, 0 failed
- [x] `cargo fmt`
- [ ] Manual: start mostrod with default settings and confirm the gift-wrap deprecation `WARN` appears once at startup

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clearer runtime warnings when the soon-to-be-removed transport option is selected, including a deprecation notice and transition to protocol v2-only behavior.
  * Updated protocol version tagging to avoid deprecated access patterns.

* **Documentation**
  * Expanded configuration inline comments to clarify the v0.19.0 removal timeline and reference the related issue.

* **Chores**
  * Marked the transport-related setting/accessors as deprecated and suppressed resulting deprecation warnings across relevant code paths and tests.
  * Bumped the `mostro-core` dependency patch version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1843,9 +1843,9 @@ dependencies = [
 
 [[package]]
 name = "mostro-core"
-version = "0.14.0"
+version = "0.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4d73bbd0d6b8b283ec477d11330617bd55d3b1e40dc07db8a084e74cc22f01c2"
+checksum = "30391197a9de16ce45b7ee5a92d5b3f74c3d12da4d13d442aff00c93c2424038"
 dependencies = [
  "bitcoin",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ reqwest = { version = "0.12.1", default-features = false, features = [
   "json",
   "rustls-tls",
 ] }
-mostro-core = { version = "0.14", features = ["sqlx"] }
+mostro-core = { version = "0.14.1", features = ["sqlx"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 async-trait = "0.1.83"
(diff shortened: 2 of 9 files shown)


grunch:
@codex review

Issue #814, opened 2026-07-20
[MEDIUM] lnurl::resolv_ln_address returns Ok("") on failure and does unchecked amount*1000


grunch:
### Severity: MEDIUM

`lnurl::resolv_ln_address` reports success on genuine failures, and multiplies an amount without an overflow/validity check.

### Where
`src/lnurl` — `resolv_ln_address` (LN address / LNURL-pay resolution).

### Problems
1. **Silent failure:** on several error paths the function returns `Ok("")` (an empty invoice string) instead of an `Err`. Callers that only check `is_ok()` will proceed with an empty BOLT11, pushing the failure downstream where it's harder to diagnose.
2. **Unchecked arithmetic:** the millisat conversion does `amount * 1000` without validating the input range, risking overflow / nonsensical amounts for adversarial or corrupt inputs.

### Suggested fix
- Return a typed `Err` on every failure path; never hand back `Ok("")`.
- Use checked arithmetic (`checked_mul`) and validate the amount range before conversion.

_Found during the deep review accompanying #803._

grunch:
Adding a concrete caller gap found while reviewing #820 locally, since it sharpens problem 1 above.

The `Ok("")` contract is not just "harder to diagnose downstream" — there is one caller that acts on the empty string:

```rust
// src/app/release.rs:499
let payment_request = if let Ok(addr) = ln_addr {
    resolv_ln_address(&addr.to_string(), amount, None)
        .await
        .map_err(|_| MostroInternalErr(ServiceError::LnAddressParseError))?
} else {
    payment_request
};
// ...
let payment_task = ln_client_payment.send_payment(&payment_request, amount as i64, tx);
```

`do_payment` never checks for empty, so a rejected LNURL resolution goes straight into `send_payment("")`. The dev-fee path does check (`resolve_dev_fee_invoice`, `src/app/dev_fee.rs:939`) and returns `LnAddressParseError` — so the two callers of the same function disagree on whether `Ok("")` means success.

#820 adds two more paths that return `Ok("")` (explicit `{"status":"ERROR"}` handling on both requests), so the number of ways to reach that `send_payment("")` grew. It's still pre-existing behavior — `tag != "payRequest"` already got there — and not a blocker for that PR, but it moves this issue's priority up a little for me.

Whoever picks this up: the typed-`Err` refactor in the suggested fix resolves both callers at once, and is preferable to bolting an `is_empty()` guard onto `do_payment`.

[...]

PR #820, opened 2026-07-20
Add LUD-12 comment to dev fee LNURL-pay calls

ToRyVand:
## Summary

Closes #694.

Dev fee payments carried no identifying information, making it impossible to trace which order or Mostro node originated a given payment on the receiving end. This adds an optional [LUD-12](https://github.com/lnurl/luds/blob/luds/12.md) `comment` (`mostro-dev-fee order=<id> node=<npub>`) to the LNURL-pay callback, gated by the server-advertised `commentAllowed` limit — servers that don't support it are left untouched.

## Bug found and fixed while implementing this

The existing callback-building code concatenated `?amount={amount_msat}` onto the raw `callback` string returned by the LNURL server, before parsing it as a URL:

```rust
let callback = format!("{callback}?amount={amount_msat}");
```

Real LNURL servers can return a `callback` that already carries its own query string (e.g. `https://host/cb?id=abc`, common with LNbits and others). In that case the second `?` is not a delimiter — everything after it, including the literal `?`, becomes part of the previous parameter's value. So `amount` silently never reaches the server as its own parameter; the previous key absorbs it instead (verified by actually running this against the exact `url` crate version pinned in this repo). This predates this PR, but since the same line was being rewritten to attach the comment, it's fixed here too rather than left next to new code. Extracted `build_callback_url`, a pure helper that adds `amount` and `comment` via `query_pairs_mut` (never by string concatenation), with a regression test (`build_callback_url_preserves_existing_query_params`) covering exactly this case.

## Other changes

`resolve_dev_fee_invoice` now receives `keys: &Keys` instead of calling `crate::util::get_keys()` internally (which re-parses the nsec — an EC key derivation — on every call). `AppContext::keys()` already caches this and is documented as the preferred alternative (`src/app/context.rs`). Threaded through `scheduler.rs -> run_dev_fee_cycle -> process_new_dev_fee_payments -> resolve_dev_fee_invoice`; `ctx` was already available at the scheduler spawn site, just not passed down. This also removes the only fallible path around building the comment, so there's no error to silently swallow.

## Response to review

@grunch's review caught real gaps — all addressed in `860bcac`:

- **`cargo fmt` failure (Critical)** — tree is now rustfmt-clean; `cargo fmt --all -- --check` passes.
- **Silent truncation (Major)** — `fit_comment` now returns `None` instead of cutting a comment mid-pubkey when it exceeds `commentAllowed`; a `warn!` fires so a dropped comment is visible in logs instead of silently producing a mutilated-but-plausible identifier.
- **No scheme validation / SSRF (Major)** — `extract_lnurl` and `build_callback_url` now reject any scheme other than `http`/`https`, closing the path a buyer-supplied lightning address could use to point mostrod at `javascript:`/`file:`/internal addresses.
- **Duplicate `amount`/`comment` params (Minor)** — any pre-existing `amount`/`comment` pair on `callback` is dropped before appending ours, so there's no ambiguity about which value the server honors.
- **LNURL error reason discarded (Major)** — both `resolv_ln_address` requests (address resolution and callback) now check for `{"status":"ERROR","reason":...}` and log the server's actual reason instead of silently returning an empty invoice.
- **Hex pubkey instead of npub (Minor)** — the comment now renders the node identifier as `npub1...` via `to_bech32()`, both more idiomatic for a nostr project and 63 chars shorter.
- **Untested comment string (Minor)** — extracted `dev_fee_comment` into its own pure function with a dedicated test asserting the exact `mostro-dev-fee order=... node=...` shape, so the format can't silently drift again.
- **`use super::Keys` (Trivial)** — test module now imports `Keys` directly from `nostr_sdk`.

Left open, both flagged as optional in the review rather than blocking: retrying once without the comment on LNURL rejection, and a settings toggle to opt out of sending the comment entirely. Happy to file follow-ups if useful.

## Test plan

```
$ cargo fmt --all -- --check
(clean)

$ cargo clippy --all-targets --all-features -- -D warnings
cargo clippy: No issues found

$ cargo test
test result: FAILED. 1031 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

The 1 failure (`lightning::invoice::tests::test_lnurl_validation_with_test_server`) is pre-existing and unrelated — it binds a mock server to a fixed port (8080) that's already occupied in my environment; confirmed neither this commit nor any other commit on this branch touches `src/lightning/invoice.rs`.

New/changed test coverage (`src/lnurl.rs` and `src/app/dev_fee.rs`):
- [x] `amount` added as its own query parameter (base case)
- [x] Regression: callback with a pre-existing query string — `amount` is added as its own pair instead of merging into the existing one
- [x] `comment` attached when the server allows it (`commentAllowed > 0`)
- [x] `comment` omitted when the server doesn't advertise support
- [x] `comment` omitted (not truncated) when it exceeds the server-advertised length
- [x] `extract_lnurl`/`build_callback_url` reject non-`http(s)` schemes
- [x] Pre-existing `amount`/`comment` query params are dropped before appending ours
- [x] `dev_fee_comment` produces the exact `mostro-dev-fee order=... node=npub1...` shape

## Summary by CodeRabbit

* **New Features**
  * Added optional LUD-12 user comments for LNURL-pay resolution when permitted by the server.
  * Dev-fee payment comments now include the order ID and node public key.
  * LNURL callback URLs preserve existing query parameters while adding amount (and comments when allowed).
* **Bug Fixes**
  * Enforced HTTP(S)-only LNURL targets and HTTP(S)-only callback URLs to prevent invalid/non-web schemes.
  * Added graceful handling for LNURL error responses by returning an empty result.
* **Tests**
  * Updated/added coverage for comment truncation rules and callback query parameter preservation.

### DIFF
--- a/src/app/dev_fee.rs
+++ b/src/app/dev_fee.rs
@@ -71,6 +71,7 @@ use mostro_core::error::MostroError;
 use mostro_core::error::MostroError::MostroInternalErr;
 use mostro_core::error::ServiceError;
 use mostro_core::order::Order;
+use nostr_sdk::Keys;
 use sqlx::SqlitePool;
 use std::collections::HashSet;
 use tokio::sync::mpsc::channel;
@@ -87,13 +88,14 @@ pub async fn run_dev_fee_cycle(
     pool: &SqlitePool,
     ln_client: &mut LndConnector,
     confirmed: &mut HashSet<uuid::Uuid>,
+    keys: &Keys,
 ) {
     info!("Checking for unpaid development fees");
 
     cleanup_stale_pending_markers(pool).await;
     verify_confirmed_orders(pool, ln_client, confirmed).await;
     recover_partial_payments(pool, ln_client, confirmed).await;
-    process_new_dev_fee_payments(pool, ln_client, confirmed).await;
+    process_new_dev_fee_payments(pool, ln_client, confirmed, keys).await;
 }
 
 // ── Phase 1: Stale PENDING cleanup ──────────────────────────────────────
@@ -392,6 +394,7 @@ async fn process_new_dev_fee_payments(
     pool: &SqlitePool,
     ln_client: &mut LndConnector,
     confirmed: &mut HashSet<uuid::Uuid>,
+    keys: &Keys,
 ) {
     let unpaid_orders = match find_unpaid_dev_fees(pool).await {
         Ok(orders) => orders,
@@ -434,7 +437,7 @@ async fn process_new_dev_fee_payments(
 
         let (payment_request, payment_hash_hex) = match tokio::time::timeout(
             std::time::Duration::from_secs(20),
-            resolve_dev_fee_invoice(&order),
+            resolve_dev_fee_invoice(&order, keys),
         )
         .await
         {
@@ -883,7 +886,10 @@ fn parse_pending_timestamp(marker: &str) -> Option<u64> {
 ///
 /// # Timeouts
 /// - LNURL resolution: 15 seconds
-pub async fn resolve_dev_fee_invoice(order: &Order) -> Result<(String, String), MostroError> {
+pub async fn resolve_dev_fee_invoice(
+    order: &Order,
+    keys: &Keys,
+) -> Result<(String, String), MostroError> {
     info!(
         "Resolving dev fee invoice for order {} - amount: {} sats to {}",
         order.id, order.dev_fee, DEV_FEE_LIGHTNING_ADDRESS
@@ -893,9 +899,21 @@ pub async fn resolve_dev_fee_invoice(order: &Order) -> Result<(String, String), 
         return Err(MostroInternalErr(ServiceError::WrongAmountError));
     }
 
+    // LUD-12 comment so the receiving end can trace which order/node a dev
+    // fee payment came from.
+    let comment = format!(
+        "mostro-dev-fee order={} node={}",
+        order.id,
+        keys.public_key()
+    );
+
     let payment_request = tokio::time::timeout(
         std::time::Duration::from_secs(15),
-        resolv_ln_address(DEV_FEE_LIGHTNING_ADDRESS, order.dev_fee as u64),
+        resolv_ln_address(
+            DEV_FEE_LIGHTNING_ADDRESS,
+            order.dev_fee as u64,
+            Some(comment.as_str()),
+        ),
     )
     .await
     .map_err(|_| {
@@ -1496,6 +1514,7 @@ mod tests {
     use super::{
         handle_payment_timeout, process_new_dev_fee_payments, recover_partial_payments,
         resolve_dev_fee_invoice, run_dev_fee_cycle, send_dev_fee_payment, verify_confirmed_orders,
+        Keys,
     };
     use crate::lightning::LndConnector;
 
@@ -1549,7 +1568,7 @@ mod tests {
         let pool = setup_orders_db().await;
         let mut ln = dead_lnd().await;
         let mut confirmed = HashSet::new();
-        run_dev_fee_cycle(&pool, &mut ln, &mut confirmed).await;
+        run_dev_fee_cycle(&pool, &mut ln, &mut confirmed, &Keys::generate()).await;
         assert!(confirmed.is_empty());
     }
 
@@ -1820,7 +1839,7 @@ mod tests {
 
         let mut ln = dead_lnd().await;
         let mut confirmed = HashSet::new();
-        process_new_dev_fee_payments(&pool, &mut ln, &mut confirmed).await;
+        process_new_dev_fee_payments(&pool, &mut ln, &mut confirmed, &Keys::generate()).await;
 
         let row: (i32, Option<String>) =
             sqlx::query_as("SELECT dev_fee_paid, dev_fee_payment_hash FROM orders WHERE id = ?")
@@ -1842,7 +1861,7 @@ mod tests {
             .unwrap();
         let mut ln = dead_lnd().await;
         let mut confirmed = HashSet::new();
-        process_new_dev_fee_payments(&pool, &mut ln, &mut confirmed).await;
+        process_new_dev_fee_payments(&pool, &mut ln, &mut confirmed, &Keys::generate()).await;
         assert!(confirmed.is_empty());
     }
 
@@ -1955,7 +1974,7 @@ mod tests {
                 .fetch_one(&pool)
                 .await
                 .unwrap();
-        assert!(resolve_dev_fee_invoice(&order).await.is_err());
+        assert!(resolve_dev_fee_invoice(&order, &Keys::generate()).await.is_err());
     }
 
     #[tokio::test]
--- a/src/app/dev_fee.rs
+++ b/src/app/dev_fee.rs
@@ -71,7 +71,8 @@ use mostro_core::error::MostroError;
 use mostro_core::error::MostroError::MostroInternalErr;
 use mostro_core::error::ServiceError;
 use mostro_core::order::Order;
-use nostr_sdk::Keys;
+use nostr_sdk::prelude::ToBech32;
+use nostr_sdk::{Keys, PublicKey};
 use sqlx::SqlitePool;
 use std::collections::HashSet;
 use tokio::sync::mpsc::channel;
@@ -879,6 +880,14 @@ fn parse_pending_timestamp(marker: &str) -> Option<u64> {
 
 // ── Invoice resolution & payment (moved from release.rs) ───────────────
 
+/// LUD-12 comment identifying which order/node a dev fee payment came
+/// from. `node` is rendered as `npub…` rather than hex: it's the
+/// self-describing form in a nostr project, and it's shorter.
+fn dev_fee_comment(order_id: &uuid::Uuid, node: &PublicKey) -> String {
+    let node_id = node.to_bech32().unwrap_or_else(|_| node.to_string());
+    format!("mostro-dev-fee order={order_id} node={node_id}")
+}
+
 /// Resolve a dev fee LNURL invoice for the given order.
 ///
 /// Contacts the dev fee lightning address, obtains a fresh invoice,
@@ -901,11 +910,7 @@ pub async fn resolve_dev_fee_invoice(
 
     // LUD-12 comment so the receiving end can trace which order/node a dev
     // fee payment came from.
-    let comment = format!(
-        "mostro-dev-fee order={} node={}",
-        order.id,
-        keys.public_key()
-    );
+    let comment = dev_fee_comment(&order.id, &keys.public_key());
 
     let payment_request = tokio::time::timeout(
         std::time::Duration::from_secs(15),
@@ -1512,11 +1517,12 @@ mod tests {
     // ── LND-dependent phases against a lazily-connected dead client ──
 
     use super::{
-        handle_payment_timeout, process_new_dev_fee_payments, recover_partial_payments,
-        resolve_dev_fee_invoice, run_dev_fee_cycle, send_dev_fee_payment, verify_confirmed_orders,
-        Keys,
+        dev_fee_comment, handle_payment_timeout, process_new_dev_fee_payments,
+        recover_partial_payments, resolve_dev_fee_invoice, run_dev_fee_cycle, send_dev_fee_payment,
+        verify_confirmed_orders,
     };
     use crate::lightning::LndConnector;
+    use nostr_sdk::Keys;
 
     /// Real `LndConnector` against a dead endpoint: `connect` is lazy,
     /// so it always builds; every RPC fails in ~1ms with a transport
@@ -1974,7 +1980,28 @@ mod tests {
                 .fetch_one(&pool)
                 .await
                 .unwrap();
-        assert!(resolve_dev_fee_invoice(&order, &Keys::generate()).await.is_err());
+        assert!(resolve_dev_fee_invoice(&order, &Keys::generate())
+            .await
+            .is_err());
+    }
+
+    #[test]
+    fn dev_fee_comment_has_expected_shape() {
+        use nostr_sdk::prelude::ToBech32;
+
+        let order_id = uuid::Uuid::new_v4();
+        let keys = Keys::generate();
+        let comment = dev_fee_comment(&order_id, &keys.public_key());
+        assert_eq!(
+            comment,
+            format!(
+                "mostro-dev-fee order={} node={}",
+                order_id,
+                keys.public_key().to_bech32().unwrap()
+            )
+        );
+        assert!(comment.starts_with("mostro-dev-fee order="));
+        assert!(comment.contains(" node=npub1"));
     }
 
     #[tokio::test]
--- a/src/lnurl.rs
+++ b/src/lnurl.rs
@@ -3,6 +3,7 @@ use mostro_core::prelude::*;
 use once_cell::sync::Lazy;
 use reqwest::Client;
 use serde_json::Value;
+use tracing::{error, warn};
 
 pub static HTTP_CLIENT: Lazy<Client> = Lazy::new(|| {
     Client::builder()
@@ -21,6 +22,11 @@ pub static HTTP_CLIENT: Lazy<Client> = Lazy::new(|| {
 /// # Returns
 /// * `Ok(String)` - The extracted LNURL
 /// * `Err(MostroError)` - If the address is invalid or cannot be resolved
+///
+/// Validates the scheme is `http`/`https` before returning: a bech32-decoded
+/// LNURL is attacker-controlled input, and every caller of this function
+/// (`ln_exists`, `resolv_ln_address`) does an unguarded GET against the
+/// result, so this is the one place that check has to hold for all of them.
 async fn extract_lnurl(address: &str) -> Result<String, MostroError> {
     let url = if address.to_lowercase().starts_with("lnurl") {
         let lnurl = LnUrl::decode(address.to_string())
@@ -39,6 +45,11 @@ async fn extract_lnurl(address: &str) -> Result<String, MostroError> {
         };
         format!("{base_url}/.well-known/lnurlp/{user}")
     };
+    let parsed = reqwest::Url::parse(&url)
+        .map_err(|_| MostroInternalErr(ServiceError::LnAddressParseError))?;
+    if !crate::util::is_http_or_https(&parsed) {
+        return Err(MostroInternalErr(ServiceError::LnAddressParseError));
+    }
     Ok(url)
 }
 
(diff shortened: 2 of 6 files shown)


grunch:
@ToRyVand thanks for your contribution, can you please solve conflicts?

ToRyVand:
@grunch conflicts resolved (rebased onto `main`). Ready for another look, thanks!

arkanoider:
uTACK for me, good job @ToRyVand , let's wait @grunch opinion

src/app/dev_fee.rs
```diff
@@ -1955,7 +1974,7 @@ mod tests {
                 .fetch_one(&pool)
                 .await
                 .unwrap();
-        assert!(resolve_dev_fee_invoice(&order).await.is_err());
+        assert!(resolve_dev_fee_invoice(&order, &Keys::generate()).await.is_err());
```
grunch: _⚠️ Potential issue_ | _🔴 Critical_

**`cargo fmt` was not run — the `fmt` CI job fails on this line.**

The PR description lists `$ cargo fmt` in the test plan, but the tree is not rustfmt-clean:

```console
$ cargo fmt --all -- --check
Diff in src/app/dev_fee.rs:1974:
-        assert!(resolve_dev_fee_invoice(&order, &Keys::generate()).await.is_err());
+        assert!(resolve_dev_fee_invoice(&order, &Keys::generate())
+            .await
+            .is_err());
```

`.github/workflows/ci.yml:16` runs `cargo fmt --all -- --check`, and the `test` job declares `needs: [fmt, clippy]` — so this blocks the whole pipeline. This is the only formatting violation in the diff.

```suggestion
        assert!(resolve_dev_fee_invoice(&order, &Keys::generate())
            .await
            .is_err());
```

src/lnurl.rs
```diff
@@ -69,7 +69,46 @@ pub async fn ln_exists(address: &str) -> Result<(), MostroError> {
...
 }
 
-pub async fn resolv_ln_address(address: &str, amount: u64) -> Result<String, MostroError> {
+/// LUD-12: returns the comment truncated to `max_len` chars, or `None` if
+/// there's no comment to send or the server advertises no support for one
+/// (`max_len == 0`).
+fn fit_comment(comment: Option<&str>, max_len: usize) -> Option<String> {
+    let comment = comment.filter(|_| max_len > 0)?;
+    Some(comment.chars().take(max_len).collect())
+}
```
grunch: _🛠️ Refactor suggestion_ | _🟠 Major_

**Silent truncation produces a plausible-looking but wrong identifier — defeating the purpose of the feature.**

The comment is 127 chars. If `commentAllowed` is ever below that (a different dev-fee address, or the provider lowering its limit), `chars().take(max_len)` cuts through the middle of the node pubkey and emits something that still *looks* like a valid hex key. Whoever reconciles payments on the receiving end has no way to tell a complete identifier from a mutilated one, and nothing is logged.

For a feature whose entire purpose is traceability, a partial trace is worse than none: omit rather than truncate.

```suggestion
/// LUD-12: returns the comment when it fits within `max_len` chars, or
/// `None` if there's no comment to send, the server advertises no support
/// for one (`max_len == 0`), or the comment would have to be truncated —
/// a half-written order id or pubkey is a worse trace than no trace.
fn fit_comment(comment: Option<&str>, max_len: usize) -> Option<String> {
    let comment = comment.filter(|c| max_len > 0 && c.chars().count() <= max_len)?;
    Some(comment.to_string())
}
```

Note this inverts `fit_comment_truncates_to_server_limit` (line 285), which would become `fit_comment_none_when_too_long`. If you prefer to keep truncating, then please at least `warn!` when it happens — the current behaviour is undetectable in production either way.

src/lnurl.rs
```diff
@@ -69,7 +69,46 @@ pub async fn ln_exists(address: &str) -> Result<(), MostroError> {
...
+/// silently mangles the query if `callback` already carries one (a real LNURL
+/// server behavior, e.g. `https://host/cb?id=abc`).
+fn build_callback_url(
+    callback: &str,
+    amount_msat: u64,
+    comment: Option<&str>,
+    comment_allowed: usize,
+) -> Result<reqwest::Url, MostroError> {
+    let mut url = reqwest::Url::parse(callback)
+        .map_err(|_| MostroInternalErr(ServiceError::LnAddressParseError))?;
```
grunch, line 114: _🛡️ Security_ | _🟠 Major_

**No scheme validation on an attacker-controlled URL.**

`callback` comes straight out of a remote LNURL server's JSON. Verified against the pinned `url 2.5.8` — any scheme parses and gets the query appended:

```
build_callback_url("ftp://h/cb", ...)          => Ok("ftp://h/cb?amount=100000")
build_callback_url("javascript:alert(1)", ...) => Ok("javascript:alert(1)?amount=100000")
```

This matters because the address is not always trusted: in `src/app/release.rs:499`, `do_payment` feeds `resolv_ln_address` a **buyer-supplied** lightning address (`order.buyer_invoice`). A hostile buyer can stand up their own LNURL server and return `callback: "http://127.0.0.1:8332/..."` or `http://169.254.169.254/latest/meta-data/`, and mostrod will issue the GET. Only `pr` is read back, so it's blind SSRF — but internal port scanning and GET-triggered side effects are real.

This weakness pre-dates the PR. Flagging it because `build_callback_url` is the choke point this PR introduces, and is now the only place in the codebase that parses this value — the check is four lines and belongs here:

```suggestion
    let mut url = reqwest::Url::parse(callback)
        .map_err(|_| MostroInternalErr(ServiceError::LnAddressParseError))?;
    if !matches!(url.scheme(), "http" | "https") {
        return Err(MostroInternalErr(ServiceError::LnAddressParseError));
    }
```

Blocking private/link-local ranges is a larger change and better tracked as a separate issue.

src/lnurl.rs
```diff
@@ -69,7 +69,46 @@ pub async fn ln_exists(address: &str) -> Result<(), MostroError> {
...
+    url.query_pairs_mut()
+        .append_pair("amount", &amount_msat.to_string());
```
grunch, line 126: _🧹 Nitpick_ | _🟡 Minor_

**A callback that already carries `amount` yields a duplicate parameter.**

Verified: `build_callback_url("https://h/cb?amount=5", 100_000, ...)` produces `https://h/cb?amount=5&amount=100000`. Which one the server honours is undefined.

Not exploitable for overpayment — `src/lightning/mod.rs:269-278` compares the decoded invoice amount against the requested one and aborts with `"Wrong amount"`. But the result is another opaque, permanently-retrying dev-fee failure (see the comment on line 141). Consider dropping any pre-existing `amount`/`comment` keys before appending, so the value you compute is the one that's sent.

src/lnurl.rs
```diff
@@ -99,7 +138,9 @@ pub async fn resolv_ln_address(address: &str, amount: u64) -> Result<String, Mos
             return Ok("".to_string());
         }
         let callback = body["callback"].as_str().unwrap_or("");
-        let callback = format!("{callback}?amount={amount_msat}");
+        let comment_allowed = body["commentAllowed"].as_u64().unwrap_or(0) as usize;
```
grunch, line 192: _⚠️ Potential issue_ | _🟠 Major_

**New failure mode lands in a path that discards the server's error reason, and then retries forever.**

Two problems compound here.

**(a) The reason is thrown away.** LNURL errors are returned as **HTTP 200** with `{"status":"ERROR","reason":"..."}`. That body has no `pr`, so `body["pr"].as_str().unwrap_or("")` (line 157) yields `""` and the function returns `Ok("")`. `src/app/dev_fee.rs:934` then logs the generic *"returned empty invoice"* and the server's `reason` is never surfaced. This PR introduces a brand-new class of rejection into that path — comment too long, comment unsupported, comment rejected — all of which will now be invisible.

Related: `as_u64()` returns `None` if a server encodes `commentAllowed` as `255.0` or as a string, silently collapsing to `0` and dropping the comment with no trace.

**(b) It retries forever with identical input.** On resolution failure, `src/app/dev_fee.rs:450` calls `release_pending_claim(...)` and `continue`s, so the next cycle retries — with the same comment — producing a permanent failure loop and no diagnostic pointing at the comment.

Minimum fix, surfacing the reason:

```rust
if body["status"].as_str() == Some("ERROR") {
    let reason = body["reason"].as_str().unwrap_or("unknown");
    error!("LNURL callback rejected: {reason}");
}
```

Worth considering as well: retry once *without* the comment before giving up. The dev fee is real money and shouldn't be blocked indefinitely by an optional field.

src/app/dev_fee.rs
```diff
@@ -893,9 +899,21 @@ pub async fn resolve_dev_fee_invoice(order: &Order) -> Result<(String, String),
...
     }
 
+    // LUD-12 comment so the receiving end can trace which order/node a dev
+    // fee payment came from.
+    let comment = format!(
+        "mostro-dev-fee order={} node={}",
+        order.id,
+        keys.public_key()
+    );
+
```
grunch, line 914: _🧹 Nitpick_ | _🟡 Minor_

Three small things on the comment string itself:

1. **The PR description doesn't match the code.** The summary says the comment is `order=<id> node=<pubkey>`; the code emits `mostro-dev-fee order=… node=…`. The prefix is useful — keep it and update the description.

2. **`keys.public_key()` renders hex, not npub.** Confirmed output: `node=96f7d5278ba0892c6379c00fb085d8b5e876e0ca53dfc4125fe8197582800b33`. In a nostr project, an operator-facing trace identifier reads better as a self-describing `npub1…`; consider `.to_bech32()`. (It also shaves 63 chars off the 127-char total.)

3. **No opt-out.** Every operator's node pubkey and order ids are now unconditionally disclosed to a third-party custodial provider. Both are already public on relays, so the marginal exposure is *correlation* rather than secrets — but it's a choice some operators will want to make. A settings toggle would be reasonable.

src/lnurl.rs
```diff
@@ -170,6 +211,81 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn fit_comment_truncates_to_server_limit() {
+        assert_eq!(
+            fit_comment(Some("order=1 node=abc"), 7),
+            Some("order=1".to_string())
+        );
     }
```
grunch, line 379: _🧪 Test coverage_ | _🟡 Minor_

**The string this PR exists to produce is never asserted.**

All eight new tests exercise the generic helpers (`fit_comment`, `build_callback_url`). Nothing asserts that `resolve_dev_fee_invoice` builds `mostro-dev-fee order=… node=…`, and nothing covers the wiring from `resolv_ln_address(comment)` through to the `comment` pair on the callback URL.

Concretely: you could delete the `format!` at `src/app/dev_fee.rs:903-908` entirely and all eight tests would still pass.

Extracting the format into a pure function makes it testable without a live LNURL server:

```rust
fn dev_fee_comment(order_id: &Uuid, node: &PublicKey) -> String {
    format!("mostro-dev-fee order={order_id} node={node}")
}
```

Then assert both the shape and that it stays within a sane budget — the 127-char figure is what makes issue #694's traceability claim hold.

src/app/dev_fee.rs
```diff
@@ -1496,6 +1514,7 @@ mod tests {
     use super::{
         handle_payment_timeout, process_new_dev_fee_payments, recover_partial_payments,
         resolve_dev_fee_invoice, run_dev_fee_cycle, send_dev_fee_payment, verify_confirmed_orders,
+        Keys,
```
grunch: _🧹 Nitpick_ | _🟢 Trivial_

Reaching a re-exported `use` item through `super::`. It compiles, but importing the crate directly in the test module is clearer about where `Keys` actually comes from:

```rust
use nostr_sdk::Keys;
```

grunch (commented):
**Actionable comments posted: 8**

<details>
<summary>🗂️ Review context</summary>

Reviewed `16cf6e6..0af1019` against `main` (branch is up to date, 0 commits behind). Findings below were verified by execution, not inspection: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test lnurl:: dev_fee::`, a live `curl` of the production dev-fee LNURL endpoint, and throwaway probes of `build_callback_url` against the pinned `url 2.5.8`.

</details>

## Verdict

The design is sound and the incidental bug find is genuine — but this cannot merge as-is: **CI will fail on `cargo fmt`**, and the feature ships with zero observability and no test on the one string the PR exists to produce.

## What's right

- **The pre-existing bug is real and correctly diagnosed.** `?` is a legal character *inside* a query string (RFC 3986: `query = *( pchar / "/" / "?" )`), so `format!("{callback}?amount={amount_msat}")` on a callback that already carried a query did cause `amount` to be absorbed into the preceding parameter's value. Good catch, good regression test, and correct to fix it here rather than leave it adjacent to new code.
- **Characters vs. bytes: you got this right.** LUD-12 states *"The value of `commentAllowed` should be the number of **characters** accepted for the `comment` query parameter."* `chars().take(max_len)` matches the spec; a fair number of implementations count bytes and get this wrong.
- Passing `&Keys` instead of re-deriving the EC key via `get_keys()` on every cycle is a legitimate improvement, cleanly threaded.
- `cargo clippy --all-targets --all-features -- -D warnings` is clean. All 14 `lnurl::` and 36 `dev_fee::` tests pass.

## Sizing note on the comment

The comment is **127 characters**:

```
mostro-dev-fee order=58bb28ce-de34-4f3c-9bcc-c73bf1756e13 node=96f7d5278ba0892c6379c00fb085d8b5e876e0ca53dfc4125fe8197582800b33
```

The production recipient currently accommodates it:

```console
$ curl -s https://walletofsatoshi.com/.well-known/lnurlp/pivotaldeborah52
{"callback":"https://livingroomofsatoshi.com/api/v1/lnurl/payreq/...","commentAllowed":255,...}
```

So this works today. The concern raised inline is the *failure* behaviour when it ever stops fitting, which is currently silent and produces a truncated-but-plausible-looking pubkey.

## ⚠️ Repository-level finding (outside this PR)

`.github/workflows/ci.yml` triggers on `on: push` only — no `pull_request`. The `fmt`, `clippy` and `test` jobs therefore ran on the contributor's fork and **are not reported as checks on this PR** (`gh pr checks 820` shows only `mint-integration`, `mutation-baseline`, `mutation-pr` — all skipped — plus CodeRabbit). That is exactly why the formatting failure below reached review unnoticed. Worth a follow-up issue: `on: [push, pull_request]`.

grunch (approved):
Approving. Reviewed locally on the branch — build, lints and the full suite, plus a couple of things I verified by running rather than reading.

## Verification

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets --all-features -- -D warnings` | clean |
| `cargo test --bins` | **1032 passed, 0 failed** |

The failure noted in the PR description (`lightning::invoice::tests::test_lnurl_validation_with_test_server`) passes here — it was the fixed port 8080 being occupied on your machine, not the code. The suite is fully green. (Filed as part of #839: that harness should bind port 0 instead.)

## What I checked beyond reading the diff

**The `?amount=` bug and its fix are real.** Ran `build_callback_url` against the `url` version pinned in this repo:

```
https://pay.example.com/cb?id=abc&amount=100000&comment=mostro-dev-fee+order%3D550e8400-...+node%3Dnpub1abc
```

`amount` lands as its own pair, and `build_callback_url_preserves_existing_query_params` covers exactly the case that used to break. Good catch, and right call fixing it here instead of leaving it next to new code.

**The feature works against the real endpoint.** Queried the live dev-fee address (`walletofsatoshi.com/.well-known/lnurlp/pivotaldeborah52`): it advertises `commentAllowed: 255`, and the comment is ~126 chars, so it fits with room to spare. No silent degradation in production.

One accuracy note for the record: that endpoint's callback is `https://livingroomofsatoshi.com/api/v1/lnurl/payreq/<uuid>`, with **no** query string — so the `?amount=` bug never affected dev-fee payments. Where it bit is the buyer payout path (`do_payment`, with buyer-supplied addresses such as LNbits-style callbacks). The fix is valuable, just on a different code path than the description implies.

## Findings, all filed as follow-ups

Nothing blocking. The design is sound, the extracted pure functions are well tested, and the scheme allow-list closes a real vector.

- #837 — **MEDIUM**, host-based SSRF. This is the "known gap, tracked separately" your `build_callback_url` doc comment promises; there was no issue behind it, so now there is. I confirmed it is reachable from user input: `user@169.254.169.254` and `user@localhost` both pass `LightningAddress::from_str` and produce an outbound GET. Pre-existing and explicitly out of scope for this PR.
- #838 — **LOW**, `reason length: N bytes` isn't actionable for an operator. CodeRabbit's log-injection concern was right, but sanitize-and-truncate is the middle ground rather than discarding the reason entirely.
- #839 — **LOW**, the two new `status: "ERROR"` branches and the `commentAllowed` wiring have no test coverage. The pure helpers are well covered; it's the plumbing inside `resolv_ln_address` that isn't.
- #840 — **LOW**, minor cleanups: the comment-fit rule is duplicated between `fit_comment` and the `warn!` branch, the warning can repeat per scheduler tick, and spaces are form-urlencoded as `+`.
- Commented on #814 rather than opening a duplicate — it already covers `resolv_ln_address` returning `Ok("")`, and this PR adds two more paths that reach it. Added the concrete detail that `do_payment` (`src/app/release.rs:499`) never guards against the empty string, while `resolve_dev_fee_invoice` does.

## One thing to fix before merge

The PR description still says the LNURL error reason is logged — *"log the server's actual reason instead of silently returning an empty invoice"*. Commit `57f1172` changed that to logging only the byte length. Worth updating the body so it matches the code that ships.

Nice work on this one — the incidental bug find and the regression test around it are the best part of the PR.

ToRyVand:
## Merged — technical summary

Merged by @grunch in `a99799e`. Recap of what shipped, for the record:

**Problem.** The dev-fee LNURL-pay callback only sent `amount` — no way for a receiving node to tell which Mostro instance or trade a payment came from.

**Solution.** LUD-12 comment support: `resolv_ln_address` reads the server's `commentAllowed` limit and attaches `"mostro-dev-fee order=<id> node=<pubkey>"` (truncated to fit), silently omitted when the server doesn't support it.

**Review round (2026-07-21, 8 comments, "cannot merge as-is") — all addressed in `860bcac`:**
- `cargo fmt` failing CI (Critical)
- `fit_comment` silently truncating mid-pubkey instead of omitting (Major)
- No scheme validation on the LNURL callback URL — SSRF surface via buyer-supplied lightning address (Major)
- Duplicate `amount`/`comment` query params possible (Minor) — fixed with `Url::query_pairs_mut()` instead of string concatenation, which also fixed a pre-existing bug where servers returning a callback with its own query string silently corrupted `amount`
- LNURL server's `reason` on `{"status":"ERROR"}` discarded (Major)
- Node id logged as hex instead of npub (Minor)
- Untested log string, import nitpick (Minor)

**Proactive follow-up (`57f1172`):** CodeRabbit's re-review flagged the same `reason` field being logged verbatim via `error!()` — log-injection risk (CWE-117) on the same buyer-controlled path. Fixed by logging byte length instead of raw text rather than waiting for another review round, since it overlapped with what grunch had already scrutinized.

**Final:** 4 commits, 6 files, +274/-16. `cargo fmt`/`clippy -D warnings`/`cargo test` clean throughout.

Closes #694.

PR #821, opened 2026-07-20
fix(lightning): don't panic the daemon on malformed preimage/hash (#804)

### LINKED ISSUE #804: [CRITICAL] Malformed hex in preimage/hash panics and kills the whole daemon (lightning/mod.rs)
### Severity: CRITICAL

A single malformed row in the `orders` table can panic and take down the **entire** daemon.

### Where
`src/lightning/mod.rs` (~lines 150 and 170), in the settle / cancel path that decodes the stored `preimage` and payment `hash` from hex:

```rust
let preimage = hex::decode(preimage).expect(...);   // ~:150
let hash     = hex::decode(hash).expect(...);        // ~:170
```

### Why it's critical
- The main event loop in `src/app.rs` processes messages **sequentially on a single task**, and `main()` does not wrap message handling in a panic boundary / `catch_unwind`.
- Any order whose `preimage`/`hash` column is not valid hex (corruption, a partial write, a bad migration, a manual DB edit, a future bug that writes the wrong column) will hit `.expect()` the next time that order is settled/cancelled and **panic the whole process** — not just that order.
- This turns a one-row data problem into a full denial-of-service for every user of the instance.

### Repro (conceptually)
1. Have an order reach the point where the hold invoice is settled/cancelled.
2. Its `preimage` or `hash` column contains a non-hex value.
3. The handler calls `hex::decode(...).expect(...)` → panic → daemon down.

### Suggested fix
- Replace both `.expect()` calls with typed error handling (`?` into a `MostroError`/`ServiceError`), so a bad row fails **that** operation and is logged, not the process.
- Consider validating/normalizing `preimage`/`hash` on write so the invariant is enforced at the boundary.
- Optionally wrap per-message handling in the event loop so one panicking handler can't kill the daemon.

_Found during the deep review accompanying #803 (coverage push). Reporting for a focused, tested fix._

grunch:
Closes #804.

## Problem

`settle_hold_invoice` and `cancel_hold_invoice` decoded the stored `preimage` / `hash` columns with `hex::from_hex(..).expect(..)`.

The main event loop in `src/app.rs` processes messages sequentially on a single task, and `main()` does not wrap message handling in a panic boundary. So **a single malformed row** — corruption, a partial write, a bad migration, a manual DB edit — would take down the **entire daemon** the next time that order or bond was settled/cancelled. A one-row data problem became a full denial of service for every user of the instance.

## Fix

Both call sites now go through a `decode_hash32` helper that returns `ServiceError::HoldInvoiceError`. The bad row fails **that** operation and is logged; the process survives.

Additionally:

- **Validation at the boundary**: the helper enforces the 32-byte invariant, so a wrong-length column produces a clear error here instead of an opaque gRPC rejection from LND.
- **No secret leakage**: the error names the offending column but never includes the value — the preimage is what claims the HTLC, and errors end up in logs.

**No signature changes**: both functions already returned `Result<_, MostroError>`, and all 11 call sites (`util.rs`, `app/cancel.rs`, `admin_cancel.rs`, `scheduler.rs`, `bond/flow.rs`, `bond/slash.rs`) already propagated with `?` or logged the error.

## Follow-up included after review

Codex flagged (P2) that `job_cancel_orders` only logged a `cancel_hold_invoice` failure and then fell through to mark the funds as returned and republish/cancel the order — persisting the order out of `find_order_by_seconds`'s waiting-state eligibility window, so no later tick re-evaluates it. The order gets cleared while the hold invoice is still encumbered.

That swallow predates this PR (any transient LND error already hit it); the `.expect()` merely masked the malformed-hash trigger by killing the process instead. Since a bad `hash` column now returns an error, the path is reachable, so it is fixed here: on failure we `continue`, leaving the order eligible for the next tick — the same trade-off the bond slash/release branch below already makes. The "funds returned" log now only fires on success.

## Audited scope

Grep over all of `src/`: no other `.expect()` / `.unwrap()` on hex decoding remains in production code (the only one left, in `cashu/mod.rs`, is under `#[cfg(test)]`). All other `cancel_hold_invoice` / `settle_hold_invoice` call sites were checked for the same log-and-proceed shape: `app/cancel.rs`, `admin_cancel.rs` and `util.rs` propagate with `?`; `bond/flow.rs` and `bond/slash.rs` classify the error and leave the bond `Locked` / unmarked for retry. `scheduler.rs` was the only offender.

## Test plan

- [x] 6 new unit tests: valid lowercase and uppercase hex, non-hex, odd length, empty string, wrong length (31 and 33 bytes).
- [x] An assert helper verifies the error names the column and **never** contains the value.
- [x] Full suite: 535 passed, 0 failed.
- [x] `cargo fmt --all` + `cargo clippy --all-targets -- -D warnings` clean.

Caveat on coverage: `job_cancel_orders` has no tests (it builds a real `LndConnector`), so the scheduler change is verified by inspection and compilation, not by a test exercising it.

### DIFF
--- a/src/lightning/mod.rs
+++ b/src/lightning/mod.rs
@@ -55,6 +55,41 @@ pub fn routing_fee_cap_sats(amount: i64) -> i64 {
     max_fee as i64
 }
 
+/// Length in bytes of a Lightning payment preimage and of the payment
+/// hash derived from it (both are SHA-256 sized).
+const HASH_LEN: usize = 32;
+
+/// Decode a hex-encoded 32-byte preimage or payment hash — as stored in
+/// the `orders` / `bonds` tables — into the raw bytes LND expects.
+///
+/// This must never panic. The main event loop in `src/app.rs` processes
+/// messages sequentially on a single task with no panic boundary, so an
+/// `.expect()` here would turn a single malformed row (corruption, a
+/// partial write, a manual DB edit) into a full-daemon outage for every
+/// user of the instance. Returning a typed error keeps the blast radius
+/// at the one operation that touched the bad row.
+///
+/// `field` names the column for the log line. The value itself is never
+/// included in the error: the preimage is the secret that claims the
+/// HTLC, and errors end up in logs.
+fn decode_hash32(field: &str, value: &str) -> Result<Vec<u8>, MostroError> {
+    let bytes = Vec::<u8>::from_hex(value).map_err(|e| {
+        MostroInternalErr(ServiceError::HoldInvoiceError(format!(
+            "invalid {field}: not valid hex ({e})"
+        )))
+    })?;
+
+    if bytes.len() != HASH_LEN {
+        return Err(MostroInternalErr(ServiceError::HoldInvoiceError(format!(
+            "invalid {field}: expected {} bytes, got {}",
+            HASH_LEN,
+            bytes.len()
+        ))));
+    }
+
+    Ok(bytes)
+}
+
 impl LndConnector {
     pub async fn new() -> Result<Self, MostroError> {
         let ln_settings = Settings::get_ln();
@@ -147,7 +182,7 @@ impl LndConnector {
         &mut self,
         preimage: &str,
     ) -> Result<SettleInvoiceResp, MostroError> {
-        let preimage = FromHex::from_hex(preimage).expect("Wrong preimage");
+        let preimage = decode_hash32("preimage", preimage)?;
 
         let preimage_message = SettleInvoiceMsg { preimage };
         let settle = self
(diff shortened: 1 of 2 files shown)


grunch:
> **P2** — Don't clear orders after local hash decode failures

Good catch, confirmed and fixed in e89f9ce.

I verified the path Codex points at (`job_cancel_orders` in `src/scheduler.rs`): the `cancel_hold_invoice` error was only logged, and execution fell through to mark the funds as returned and republish/cancel the order. That persists the order out of `find_order_by_seconds`'s eligibility window (`status ∈ {WaitingBuyerInvoice, WaitingPayment}`), so **no later tick re-evaluates it** — the order is cleared with the hold invoice still encumbered, and the `info!` claims the funds went back to the seller.

One nuance on the origin: the swallow **predates** this PR — any transient LND error already triggered it. What the `.expect()` did was mask the malformed-hash trigger by killing the process before reaching that code. Now that a typed error is returned, the path becomes reachable, so it belongs in this PR.

The fix `continue`s on failure, leaving the order eligible for the next tick — the same trade-off the bond slash/release branch a few lines below already makes: stay consistent with the HTLC and retry, rather than persist a state that doesn't reflect it. The "funds returned" log now only fires on success.

I audited the other call sites for the same shape: `app/cancel.rs`, `admin_cancel.rs` and `util.rs` propagate with `?`; `bond/flow.rs` and `bond/slash.rs` classify the error and leave the bond `Locked` / unmarked as `Released` for retry. `scheduler.rs` was the only one with the log-and-proceed shape.

`cargo fmt` + `cargo clippy --all-targets -D warnings` clean, 535 tests green.

PR #824, opened 2026-07-20
ci: serve the coverage badge from our own Pages, drop Codecov

grunch:
Follow-up to #823.

## Why

Two problems with the Codecov badge I added in #823:

1. **codecov.io asks visitors to sign up** to view the report. That's poor form generally, and particularly at odds with a project whose premise is not depending on intermediaries.
2. **The badge read `unknown`** at the very top of the README. Codecov was never onboarded, so the upload step skipped itself exactly as designed — the net effect was advertising a broken integration on the front page.

## What changes

The report is already published on our own domain, so the badge data can come from there too:

- The workflow emits `coverage/badge.json` in shields.io's [`endpoint`](https://shields.io/badges/endpoint-badge) schema, next to the HTML.
- The README renders it via shields — already a dependency for the license, Rust and version badges, so nothing new is trusted.
- The number therefore always matches the report sitting beside it, and **no reader has to sign up for anything**.

The percentage is read from `--json --summary-only` (`.data[0].totals.lines.percent`), not scraped out of the human-readable table. Badge colour follows the usual shields thresholds so a regression is visible at a glance.

Also in here:

- `lcov.info` now ships alongside the HTML at `coverage/lcov.info` instead of being rendered and discarded, so editors and external tools can consume the raw data.
- The README points at the canonical `https://mostro.network/mostro/coverage/` rather than the `github.io` URL that redirects to it.
- The Codecov step, its token plumbing and the "Codecov tracks the trend" sentence are all removed.

## What we give up

Codecov's per-PR delta comment and historical trend graph. Neither exists today (nothing was ever uploaded), so nothing is actually lost. If the delta-on-PR feedback turns out to be wanted later, Codecov can be added back alongside this badge — they aren't mutually exclusive.

## Test plan

- [x] YAML parses.
- [x] Ran the badge script locally against real profile data.
- [x] Confirmed shields renders the generated value.

One bug found by actually running it rather than assuming: the script first produced `92,8%` — with a comma — because `printf`'s decimal separator follows `LC_NUMERIC` and my machine is localised. GitHub runners are on the C locale, so **CI would never have caught this**; it would only have surfaced for the next contributor who ran it locally. The step now exports `LC_ALL=C` instead of relying on the runner's locale.

## Summary by CodeRabbit

* **Documentation**
  * Updated the project’s coverage badge to use the Mostro-hosted report.
  * Added the new published coverage report URL and clarified coverage generation and badge updates.

* **Chores**
  * Coverage reports now generate badge data and publish directly to the hosted coverage site.
  * Removed automated Codecov uploads from the coverage workflow.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -39,11 +39,6 @@ jobs:
   coverage:
     name: Generate and publish report
     runs-on: ubuntu-latest
-    env:
-      # Surfaced as an env var so the upload step can be skipped when Codecov
-      # is not configured: the `secrets` context is not available in a
-      # step-level `if:`.
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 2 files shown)


grunch:
@codex review

PR #826, opened 2026-07-20
test(restore-session): mutation score 75%, plus a cargo-mutants CI job

### LINKED ISSUE #637: test: mutation testing for session restore module
## Context

Follow-up from #618. Implement mutation testing for session restore — ensures users can recover their trading sessions.

## Scope

- `src/app/restore_session.rs` — Session restoration logic

## Goal

- Run `cargo mutants --file src/app/restore_session.rs`
- Kill surviving mutants
- Target: **>70% mutation score**

## Priority Mutants to Kill

- Session validation logic
- Order state recovery
- Permission checks (who can restore what)
- Data integrity verification

## Acceptance Criteria

- [ ] Baseline mutation report
- [ ] Critical mutants killed
- [ ] Mutation score documented in PR

Ref: #618

ToRyVand:
Follow-up of #618 / closes #637.

**Rescoped.** This PR previously extracted `util::is_valid_hex_pubkey` from the two hex guards in `restore_session_action`. #848 removes those guards as unreachable, and it's right — `identity` and `sender` are `PublicKey` (mostro-core `nip59.rs:71,75`), so `.to_string()` is always 64 hex and the error arm can't be hit. The helper's only two call sites were those guards, so it went with them rather than fight #848 for the same lines. The two invalid-key tests it also added turned out to already exist on `main` from #803 (`restore_session.rs:242`, `:264`).

Rebased onto `main` at 0.18.5.

## Changes

- `.github/workflows/mutation.yml` — mutation-testing job, with concurrency capped so `cargo mutants` doesn't OOM the runner, and PR-diff filenames passed as an argv array instead of spliced through `make ARGS=` into a second shell.
- `Makefile` — a `mutants` target for running the same thing locally.
- `src/app/restore_session.rs` — `RESTORE_SESSION_TIMEOUT_SECS` names the inline `60 * 60`, with a test pinning the value, and the timeout log now formats from the constant instead of a hardcoded `"1 hour"` that drifted from it.
- `.gitignore` — local `.claude/` settings.

## Mutation report

`cargo mutants --file src/app/restore_session.rs` → **75% (9/12 caught)**, against the >70% target in #637.

The 3 survivors:

| Mutant | Why it survives |
|---|---|
| `17:5` replace `restore_session_action` with `Ok(())` | Killing it needs a full `AppContext` (real Sqlite pool + `nostr_sdk::Client` + `Settings` + `OrderMsgQueue`). Out of proportion here, and consistent with how `admin_add_solver_action` is left untested at that level. |
| `24:59` replace `\|\|` with `&&` | Inside the master-key hex guard. |
| `29:58` replace `\|\|` with `&&` | Inside the trade-key hex guard. |

Both `\|\|` survivors live in the guards #848 deletes. Once that merges they stop existing as mutants, and this file goes to **90% (9/10)** with no further test work.

## On the CodeRabbit review

- **Timeout log tied to the constant** — applied (`efd8a7c`).
- **Behaviour-level test for the timeout branch** — skipping. Driving the `Err(_)` branch needs a test-controlled timeout, which means making the duration injectable in production code purely so a test can reach it. The constant test pins the product decision (1 hour), and the mutant that changes it is caught; the wholesale-replacement mutant above is the honest remaining gap, already documented rather than papered over.

## Test plan

- [x] `cargo fmt --check` — clean
- [x] `cargo test --all` — 1214 passed, 3 ignored
- [x] `cargo mutants --file src/app/restore_session.rs` — 12 mutants, 9 caught, 3 missed (75%)

## Summary by CodeRabbit

- **Security**
  - Sensitive master-key and trade-key values are no longer exposed in restore-session logs or timeout messages.

- **Reliability**
  - Restore-session result waiting now uses a defined one-hour timeout, with clearer timeout reporting.

- **Maintenance**
  - Automated testing and validation workflows now run more safely and consistently.

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,0 +1,4 @@
+# Cap test-thread fan-out per mutant run. Without this, each cargo-mutants
+# job spawns a test binary with --test-threads = num_cpus, which multiplies
+# with CARGO_MUTANTS_JOBS (see `make mutation-test`) and can exhaust RAM.
+additional_cargo_test_args = ["--test-threads=4"]
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -1,4 +1,0 @@
-# Cap test-thread fan-out per mutant run. Without this, each cargo-mutants
-# job spawns a test binary with --test-threads = num_cpus, which multiplies
-# with CARGO_MUTANTS_JOBS (see `make mutation-test`) and can exhaust RAM.
-additional_cargo_test_args = ["--test-threads=4"]
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,132 @@
+# Line-coverage report: publishes the browsable `cargo llvm-cov` HTML to
+# GitHub Pages and uploads lcov to Codecov for the badge / trend.
+#
+# NOTE: intentionally NOT run on push or per-PR. `cargo llvm-cov` rebuilds the
+# whole tree with instrumentation and then runs the full suite (~1000 tests),
+# so wiring it to every push burns CI minutes to regenerate a report nobody
+# reads between merges. Same reasoning as `mutation.yml`: a weekly audit plus
+# on-demand runs. Trigger it by hand from the Actions tab after a merge that
+# you expect to move coverage.
+#
+# PUBLISHING PATH: branch-based (`gh-pages`), deliberately matching
+# `mutation.yml`. GitHub Pages has a single Source setting per repo, so a
+# repo cannot serve both a branch and an `actions/deploy-pages` artifact.
+# Using `deploy-pages` here would make the coverage artifact the *entire*
+# site and silently stop serving `/mutation-testing/`. Each report owns a
+# subdirectory instead: `/coverage/` and `/mutation-testing/`.
+name: Coverage
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Weekly, Sundays at 05:00 UTC — two hours after the mutation audit so the
+    # two long jobs don't contend for runners.
+    - cron: '0 5 * * 0'
+
+env:
+  CARGO_TERM_COLOR: always
+
+# `contents: write` is required to push the rendered report to `gh-pages`.
+permissions:
+  contents: write
+
+# Never let two runs push `gh-pages` at once; the later one wins.
+concurrency:
+  group: coverage-pages
+  cancel-in-progress: true
+
+jobs:
+  coverage:
+    name: Generate and publish report
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          # Nothing after this step needs the checkout token; the publish step
+          # gets its own. Keeps the credential out of the later cargo/llvm-cov
+          # steps and out of any artifact they produce.
+          persist-credentials: false
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          # llvm-cov needs the profdata/profraw tooling shipped here.
+          components: llvm-tools-preview
+
+      - name: Install protobuf
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y protobuf-compiler
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
+        with: { cache-on-failure: true }
+
+      - uses: taiki-e/install-action@cargo-llvm-cov
+
+      # Run the instrumented suite once, then render every report from the
+      # same profile data. Rendering separately would re-run all ~1000 tests.
+      #
+      # Default features only, matching `cargo test` in ci.yml. `--all-features`
+      # would pull in `startos`, a packaging feature with no test surface, and
+      # the published percentage would stop matching what contributors get from
+      # `cargo llvm-cov --summary-only` locally.
+      - name: Run instrumented tests
+        run: cargo llvm-cov --no-report
+
+      # `--output-dir DIR` writes into `DIR/html`, not `DIR`. Passing
+      # `target/llvm-cov/html` here would nest it one level deeper and the
+      # published directory would have no index.html at its root — a 404.
+      - name: Render HTML report
+        run: cargo llvm-cov report --html --output-dir target/llvm-cov
+
+      # Published next to the HTML so editors and external tools can consume
+      # the raw data without re-running the suite.
+      - name: Render lcov report
+        run: cargo llvm-cov report --lcov --output-path target/llvm-cov/html/lcov.info
+
+      - name: Coverage summary
+        run: cargo llvm-cov report --summary-only >> "$GITHUB_STEP_SUMMARY"
+
+      # Self-hosted badge data, so the README number comes from this repo's
+      # own Pages site instead of a third-party service that would make
+      # readers sign up to see it. shields.io renders it via its `endpoint`
+      # type; the repo already depends on shields for the license, Rust and
+      # version badges, so this adds no new dependency.
+      #
+      # The percentage is read from the machine-readable summary rather than
+      # scraped from the human table: `--json` with `--summary-only` is the
+      # supported combination and its shape is stable
+      # (`.data[0].totals.lines.percent`).
+      - name: Generate badge data
+        run: |
+          set -euo pipefail
+          # Force the C locale: printf's decimal separator follows LC_NUMERIC,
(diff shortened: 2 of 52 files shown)


ToRyVand:
## Status after 15 quiet days — Critical finding closed, and I found my own commit had broken mutation testing entirely

Bumping this because the last thing on the PR is CodeRabbit's Critical, which reads as unanswered. It isn't — but verifying that turned up something worse in my own diff.

### The shell-injection finding is fixed (`8c0e3f9`)

Filenames from `git diff --name-only` are attacker-controlled and were being joined into a string, round-tripped through `make ARGS=`, and re-parsed by a second shell. They are now built as a bash array and passed as `"${file_args[@]}"`, so a crafted filename can only land as a single `--file` value, never as extra argv tokens. The PR job calls `cargo mutants` directly; the Makefile carries a comment saying `$(ARGS)` is a plain-text splice safe only for hand-typed input.

One precision on the wording: the vector was **argument injection into cargo-mutants/cargo/rustc's flag surface**, not classic command substitution — unquoted expansion word-splits but does not re-parse `$()`/backticks. Same fix either way, but the Makefile comment documents the threat model, so it should be accurate.

### `c6d3e5c` — `.cargo/mutants.toml` was disabling every mutation run

`87b2b6f`, in this PR, added it as an OOM guard:

```toml
additional_cargo_test_args = ["--test-threads=4"]
```

cargo-mutants places those args **before** `cargo test`'s own `--`, so cargo rejects the flag instead of forwarding it to libtest:

```
*** cargo test --verbose --package=mostro@0.18.0 --test-threads=4
error: unexpected argument '--test-threads' found
*** result: Failure(1)
ERROR cargo test failed in an unmutated tree, so no mutants were tested
```

The baseline never passed, so **no mutant was ever tested** — through the Makefile target or the CI job, since cargo-mutants reads this file however it is invoked. A guard that silently disabled the thing it was guarding, which is the worse failure mode: it reads as protection while providing none.

No config or CLI mechanism in `cargo-mutants 27.1.0` forwards arguments past that `--`; `CARGO_MUTANTS_JOBS` is the cap that actually binds, and it was sufficient on its own. Removed the file. The baseline now runs to completion (1021 passed locally; the single failure is the known hardcoded-8080 `AddrInUse` flake that #849 fixes).

### Separate, pre-existing: the root `.mutants.toml` is never read

Not from this PR — `711b807` (#619) — but worth knowing while mutation config is in view. cargo-mutants reads `.cargo/mutants.toml`; `--config`/`--no-config` are both documented against that path. The root `.mutants.toml` is inert. Verified by replacing its contents with an invalid key: cargo-mutants ran happily and listed mutants in `src/main.rs`, which that file's own `exclude_globs` excludes.

So its `examine_globs`, `exclude_globs`, `timeout = 600`, `jobs = 4` and `test_tool_options` have never applied. Happy to open a separate issue rather than widen this PR — say which you'd prefer.

### Heads-up on merge order with #849

#849 touches the same `mutation-test` target: `CARGO_MUTANTS_JOBS=1` and no hard-coded LN port. The reason applies here too — workers share the host's TCP ports, and a test failing on `AddrInUse` **counts as a killed mutant**, inflating the score. Whichever lands second needs a trivial rebase on that target.

ToRyVand:
#848 removes the two hex guards in `restore_session.rs` as unreachable, and it's right: `identity` and `sender` are `PublicKey` (mostro-core `nip59.rs:71,75`), so `.to_string()` is always 64 hex and the error arm can't be hit.

This PR refactors those same guards into an `is_valid_hex_pubkey` helper and tests them — mutation coverage for code that shouldn't exist. Dropping that part, keeping the mutation CI job, the Makefile target and the named timeout constant. No reply needed.

Catrya (changes requested):
Request changes. First the good part: passing the PR-diff filenames as a bash
array instead of word-splitting a string is the right fix — those names are
attacker-controlled and `cargo mutants $file_args` splits them — and the comment
explaining why that job deliberately bypasses the Makefile target is well
placed. The branch also merges cleanly into main, compiles, and the 11
`restore_session` tests pass on the merged tree.

**`CARGO_MUTANTS_JOBS=2` does the opposite of what the commit says.** The commit
is titled "cap cargo-mutants concurrency to avoid OOM", but upstream treats
parallelism as opt-in: the book says cargo-mutants *can* run jobs in parallel and
suggests "starting at `-j2` or `-j3`" when you want more. So `2` isn't capping
anything — it's doubling the default.

Worse, the same page has a **Non-hermetic tests** section:

> If your test suite is non-hermetic — for example, if it talks to an external
> database — then running multiple jobs in parallel may cause test flakes.
> cargo-mutants is just running multiple copies of `cargo test` simultaneously:
> if that doesn't work in your tree, then you can't use this option.

This tree is non-hermetic: `test_lnurl_validation_with_test_server` binds a fixed
`127.0.0.1:8080` — `b31d3d0` here even calls it out as a known `AddrInUse` flake.
With two workers that test fails on contention, and a test failing for its own
reasons scores as a killed mutant, inflating the number this PR reports.

That's your own argument in #849, verbatim from its Makefile:

> Single worker on purpose: [...] parallel workers collide on the LNURL test's
> listener. That collision fails the test, and a test failing for its own reasons
> counts as a killed mutant — silently inflating the score this target exists to
> measure.

#849 sets `1`, this sets `2`, both are open, and they conflict: each adds a
`mutation-test` target and rewrites the same `mutation.yml` jobs. **These two
should be one PR.** Merging either first breaks the other, and in the wrong order
CI ends up running mutation testing with a port-flake-inflated score — the exact
problem #849 exists to document.

**`b31d3d0`'s conclusion is wrong.** The diagnosis is excellent — the config key
lands before `cargo test`'s `--`, which is why the baseline never passed — but
"no config-file or CLI mechanism in cargo-mutants 27.1.0 forwards arguments past
that `--`" isn't right. Upstream documents a second double-dash:

```sh
cargo mutants -- -- --test-threads 1 --nocapture

```
and warns about exactly the trap you hit: "this may interact poorly with using
additional_cargo_test_args in the configuration file, as the argument lists are
currently appended without specially handling the -- separator." The mechanism
exists; the value was in the wrong place. Dropping the OOM guard entirely and
doubling the worker count leaves the tree in the worst of both states.

The test is a tautology.

assert_eq!(RESTORE_SESSION_TIMEOUT_SECS, 3600);

It restates the constant three lines above it. It kills the mutant, but it
protects nothing — change the timeout to 30 minutes and you edit the constant and
the test in the same commit. You're honest about the other gap (the
whole-function replacement mutant, documented rather than papered over); this one
deserves the same honesty. The real improvement in this file is that the log no
longer says a hardcoded "1 hour" while the constant could drift — that part
stands on its own.

The 75% is stale. #848 is merged: those hex guards no longer exist on main
(10 lines gone from restore_session.rs), so the two || survivors you attribute
to it are already gone. The body predicts 90% "once that merges" — it did. Worth
re-measuring before this lands.

Nits: the log now prints "3600 seconds" instead of "1 hour", which is a step down
for whoever reads it; and .gitignore's .claude is unrelated to the subject of
the PR.

ToRyVand:
Closing in favour of #849, per @Catrya's review here: both PRs add a
`mutation-test` target and rewrite the same `mutation.yml` jobs, with
conflicting worker counts, so merging either first breaks the other.

What carried over into #849:

- the bash array for the PR job's `--file` flags — the real fix in this PR,
  and it stays as-is;
- the `mutation-test` target, with the Makefile's `ARGS` warning about
  attacker-controlled input;
- the named `RESTORE_SESSION_TIMEOUT_SECS` and its derived log message.

What did not:

- `CARGO_MUTANTS_JOBS=2`. @Catrya was right that it doubles rather than caps,
  and I could check the part she couldn't: cargo-mutants 27.1.0 keeps exactly
  one scratch dir on a 16-CPU host with no `-j`, so one mutant at a time is
  already the default. #849 pins `1` explicitly and says that is what it is.
- `assert_eq!(RESTORE_SESSION_TIMEOUT_SECS, 3600)`. It restated the constant
  three lines above it. Dropping it left `60 * 60` generating two mutants
  nothing kills, so the constant became the literal `3_600` — the product
  computed nothing and existed only to be flipped.
- the `.gitignore` `.claude` entry, unrelated to the subject.

The 75% figure here is superseded: #848 merged and removed the hex guards the
two `||` survivors were attributed to. #849 re-measures on the surface it
actually adds tests for — 17 mutants, 14 caught, 1 missed, 2 unviable.

PR #831, opened 2026-07-20
feat(cashu): restore/monitor in-flight locked escrows — Track A TA-3

grunch:
> **Stacked on #830 (TA-2) → #829 (TA-1) → #828 (CF-5).** Review/merge the stack in order; the base retargets as each lands. The diff shown is TA-3 only.

## What & why

On boot in Cashu mode, re-hydrate in-flight locked escrows from the DB — the Cashu analogue of the Lightning path's `find_held_invoices` resubscribe — so a restart never loses sight of an escrow that is locked but not yet released (`docs/cashu/02-track-a-lock.md` §6/§10). This is the optional, recommended monitor PR of Track A.

## Changes

- **`CashuClient::check_token_unspent`**: whether **every** proof of a stored escrow token is still unspent at the mint (NUT-07), fail-closed on a short state reply. Reuses the Y-derivation the lock path already uses; it does **not** re-run the full `verify_escrow_token` acceptance check (the token is immutable and was validated at lock — this only answers "is it still live?").
- **`cashu_restore::restore_cashu_escrows`**: list `find_locked_cashu_orders`, and for each re-check the token, logging a warning for any escrow that is spent/pending (redeemed or double-spent while the daemon was down). **Best-effort, log-only** — it never mutates order state (that belongs to the release/cancel/dispute tracks) and never blocks boot on a mint hiccup.
- Wired into the `main.rs` Cashu boot branch after the mint connects, before the scheduler starts.

## Tests

- Restore over an empty pool finds nothing and never contacts the mint (offline).
- The mint-backed spent/live detection is exercised by the CF-3 integration harness.

## Checklist

- [x] `cargo fmt --check`
- [x] `cargo test` — 1028 passed
- [x] Off-by-default: no behaviour change when Cashu is disabled

Depends on #830 (TA-2), #829 (TA-1), #828 (CF-5). Refs: `docs/cashu/02-track-a-lock.md` (TA-3).

---

Prereqs: the Cashu-mode `mostrod` setup from CF-5 (#828) — mint up (`docker compose -f docker-compose.cashu.yml up -d`), `[cashu] enabled = true`. This branch is stacked on TA-2 (contains CF-5 + TA-1 + TA-2 + TA-3). Run with `RUST_LOG=mostro=info`.

### 1. Clean start — nothing locked

1. With an empty (or no-locked-escrow) DB, start `mostrod` in Cashu mode.
2. Expected, during boot (after the mint connects, before the scheduler starts):
   ```
   cashu restore: no in-flight locked escrows
   ```
   ✔️

### 2. Re-hydrate a locked escrow across a restart

1. Get an order into the locked state — either run the end-to-end lock on the full stack (take + `AddCashuEscrow`), or seed a row directly:
   ```sql
   UPDATE orders
   SET cashu_mint_url = 'http://127.0.0.1:3338',
       cashu_escrow_token = '<a real token minted on the test mint>',
       cashu_escrow_locked_at = strftime('%s','now'),
       status = 'active'
   WHERE id = '<order-id>';
   ```
2. Restart `mostrod`. Expected at boot:
   ```
   cashu restore: re-hydrating 1 in-flight locked escrow(s)
   cashu restore: order <id> escrow is live (status active)
   ```
   ✔️
3. Now spend/redeem that token out-of-band at the mint, then restart again. Expected — the monitor flags it:
   ```
   cashu restore: order <id> escrow token is spent/pending at the mint (status active) — needs attention
   ```
   The restore never mutates order state — it only logs. ✔️

### 3. Unit test

```bash
cargo test --bin mostrod cashu_restore
```
Expected: the empty-pool restore path is a clean no-op that never contacts the mint. ✔️

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1052,10 +1052,11 @@ mod tests {
         }
 
         /// Every action that creates, advances, or settles an order — plus the
-        /// permanently-blocked buyer-invoice/bond actions and the not-yet-
-        /// implemented `AddCashuEscrow` (its CF-5 stub returns `InvalidAction`
-        /// too) — must be rejected with `CantDo(InvalidAction)` in Cashu
-        /// foundation mode. This is the DoD "no trade can complete yet" gate.
+        /// permanently-blocked buyer-invoice/bond actions — must be rejected
+        /// with `CantDo(InvalidAction)` in Cashu foundation mode. This is the
+        /// DoD "no trade can complete yet" gate. `AddCashuEscrow` is excluded:
+        /// Track A (TA-1) implements it, so it no longer routes to
+        /// `InvalidAction` — it runs the real lock handler.
         #[tokio::test]
         async fn blocks_every_order_lifecycle_action_with_invalid_action() {
             let _ =
@@ -1075,7 +1076,6 @@ mod tests {
                 Action::Cancel,
                 Action::Dispute,
                 Action::RateUser,
-                Action::AddCashuEscrow,
                 Action::AdminCancel,
                 Action::AdminSettle,
                 Action::AddBondInvoice,
--- a/src/app.rs
+++ b/src/app.rs
@@ -5,6 +5,7 @@
 pub mod context;
 
 // Submodules for different trading actions
+pub mod add_cashu_escrow; // Cashu escrow lock handler (Track A / CF-5 stub)
 pub mod add_invoice; // Handles invoice creation
 pub mod admin_add_solver; // Admin functionality to add dispute solvers
 pub mod admin_cancel; // Admin order cancellation
@@ -26,6 +27,7 @@ pub mod take_sell; // Taking sell orders
 pub mod trade_pubkey; // Trade pubkey action // Sync user trade index action
 
 // Import action handlers from submodules
+use crate::app::add_cashu_escrow::add_cashu_escrow_action;
 use crate::app::add_invoice::add_invoice_action;
 use crate::app::admin_add_solver::admin_add_solver_action;
 use crate::app::admin_cancel::admin_cancel_action;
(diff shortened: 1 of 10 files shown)


ToRyVand (commented):
Reviewed the TA-3 change itself. The monitor is well-scoped and the fail-closed NUT-07 check is right. Two things though: the branch as pushed can't be reviewed or merged as-is, and the restore query has a growth problem that will surface as soon as Track B lands.

## 1. The stack has diverged — this branch does not contain #830

`#831`'s head (`456c3f4`) is **not** a descendant of `#830`'s head (`f76d5fe`). The two branches carry *different commits for the same work*:

| Work | on #830 | on #831 |
|---|---|---|
| CF-5 | `abd9f3a` (#828, **merged to main**) | `0793689` (own copy) |
| TA-1 | `ec4a046` (#829, **merged to main**) | `4c0780a` (own copy) |
| TA-2 | `0baf13a` + `f76d5fe` (the atomic-claim fix) | `a16c745` (own copy, no CAS fix) |

So #830 has been rebased onto `main` and picked up the merged CF-5/TA-1, while #831 still carries the pre-merge duplicates. Three consequences:

- The header's *"The diff shown is TA-3 only"* isn't what GitHub renders — the +1198/−178 across 10 files includes duplicated TA-1/TA-2 content, which is why `add_cashu_escrow.rs` (+367), `util.rs` (+148) and `take_sell`/`take_buy` appear in a PR whose Changes list mentions none of them.
- #828 and #829 are **already merged**, so this would re-apply landed work.
- **It's also missing `f76d5fe`** — the atomic-claim fix on #830 that closes the concurrent-take race. TA-3's own copy of TA-2 predates it.

The real TA-3 change is small and reviewable: `git diff a16c745 456c3f4` is **117 lines across 3 files** (`cashu/mod.rs` +30, `cashu_restore.rs` +82, `main.rs` +5) — exactly the Changes list. That's what I reviewed below. A rebase onto current #830 (or onto `main` once #830 lands) would make the PR show that and nothing else.

Since #832 is stacked on this branch, it inherits the same divergence.

## 2. The restore set only ever grows, and boot waits on it

`find_locked_cashu_orders` selects `WHERE cashu_escrow_locked_at IS NOT NULL` with **no status filter and no LIMIT** — the test is even named `find_locked_cashu_orders_includes_locked_regardless_of_status`. And nothing ever clears that column; a test comment in the same file states it outright: *"nothing ever clears `cashu_escrow_locked_at`"*. The only writer is the lock CAS.

Today that's harmless — Track A has no release path, so every locked escrow genuinely is in flight. But once release lands (Track B), a completed trade keeps `cashu_escrow_locked_at` set with a legitimately **spent** token, so on every subsequent boot the monitor will:

- re-check every escrow the node has ever locked, for the life of the database, and
- log `escrow token is spent/pending at the mint … — needs attention` for each completed trade — a warning that means "healthy" in that case.

An unbounded, monotonically growing set of false alarms is the failure mode that trains operators to ignore the warning, which costs exactly the signal this PR exists to provide.

It also has a cost at boot. `restore_cashu_escrows(...).await` is awaited **inline** in the Cashu boot branch (`main.rs`), the loop is sequential, and each `check_token_unspent` is bounded by `MINT_REQUEST_TIMEOUT` (10 s) but the *aggregate* is not. With an unreachable mint the boot delay is `N × 10 s` — so *"never blocks boot on a mint hiccup"* holds per call but not overall. After a few hundred lifetime trades that's a long wait before the scheduler starts.

Three directions, any of which would do:
- filter the query to non-terminal statuses (cheapest, and it's where the intent already is — "in-flight");
- clear `cashu_escrow_locked_at` (or stamp a `cashu_escrow_released_at`) when the escrow resolves, which Track B needs a hook for anyway;
- run the checks concurrently (`futures::future::join_all`, as `PriceManager::update_all` already does) and/or spawn the restore instead of awaiting it, so a sick mint delays the monitor rather than the daemon.

The first is probably the right one to do here since it keeps TA-3 self-contained; the second is the real fix once release exists.

## 3. The TA-3 code itself reads well

- `check_token_unspent` deliberately skipping the full `verify_escrow_token` re-validation is correct and the reasoning is stated (token is immutable and was validated at lock; this only answers "still live?").
- Fail-closed on a short state reply (`states.states.len() != expected → false`) is the right default and is the kind of thing that's easy to get backwards.
- Log-only with no state mutation, and the explicit note that state transitions belong to release/cancel/dispute, keeps the boundary clean.
- The "locked but carries no escrow token" arm flagging a data anomaly rather than silently skipping is a good touch.

I reviewed the isolated `a16c745..456c3f4` commit rather than running the suite on the branch as pushed, since a tree that re-applies already-merged commits isn't the one that would land — happy to re-check after a rebase.

Contributor, not a maintainer — technical review only, no merge signal implied.

PR #835, opened 2026-07-21
fix(restore-session): scrub Nostr keys from log lines

### LINKED ISSUE #834: Nostr keys logged in cleartext in restore_session.rs (violates AGENTS.md log-scrubbing guideline)
## Problem

`AGENTS.md:48` states:

> Scrub logs that might leak invoices or Nostr keys; rotate secrets promptly if exposed.

`src/app/restore_session.rs` violates this in 3 places (checked against `main`@`57b8ff5`), logging raw Nostr keys via `tracing::info!`/`tracing::warn!`:

1. **`master_key`** — the user's persistent Nostr identity key, not a per-trade ephemeral one — in `restore_session_action`, [lines 29-32](https://github.com/MostroP2P/mostro/blob/57b8ff5/src/app/restore_session.rs#L29-L32):
   ```rust
   tracing::info!(
       "Starting background restore session for master key: {}",
       master_key
   );
   ```

2. **`trade_key`** in `send_restore_session_response`, [line 102](https://github.com/MostroP2P/mostro/blob/57b8ff5/src/app/restore_session.rs#L102):
   ```rust
   tracing::info!("Restore session response sent to user {}", trade_key,);
   ```

3. **`trade_key`** in `send_restore_session_timeout`, [line 115](https://github.com/MostroP2P/mostro/blob/57b8ff5/src/app/restore_session.rs#L115):
   ```rust
   tracing::warn!("Restore session timed out for user: {}", trade_key);
   ```

## Why it matters

`master_key` is the user's persistent Nostr identity — not a rotating per-trade key. If these `info`/`warn` logs reach an external aggregator or observability platform (routine in production), anyone with log access can correlate "this identity restored a session at this timestamp" in cleartext. That's a real privacy leak for a project whose core value proposition is trade privacy.

## Suggested fix

Drop the key from the log message (surrounding context — order id, timing — is usually enough for debugging), or if cross-log correlation is genuinely needed, log a short non-reversible prefix/hash instead of the full key.

## Found via

2 of the 3 instances were flagged by CodeRabbit's automated review on #826 (a mutation-testing PR that never touches these lines — confirmed via `git diff upstream/main...HEAD -- src/app/restore_session.rs`, so this is pre-existing on `main`, unrelated to that PR). The third (`master_key` at line 29) turned up grepping the same pattern across the rest of the file after the bot's finding.

ToRyVand:
## Problem

`AGENTS.md:48` says: *"Scrub logs that might leak invoices or Nostr keys; rotate secrets promptly if exposed."*

`src/app/restore_session.rs` logged `master_key`/`trade_key` verbatim in 3 places (`tracing::info!`/`tracing::warn!`). `master_key` in particular is the user's persistent Nostr identity, not a per-trade ephemeral key — every successful restore-session request hits this path in production (the hex-format validation guards always pass for real `PublicKey` values, so this isn't a rare branch). If these logs reach an external aggregator, anyone with log access can correlate "this identity restored a session at this timestamp" in cleartext.

## Solution

Dropped the key from all 3 log lines. The surrounding message (action + outcome) is still useful for debugging without the identifying value.

## Tests

No test asserted on log content before or after — the fix only removes an interpolated value from 3 `tracing` macro calls, no behavior change. Full suite still green: `cargo test` 1019/1020 passing (the 1 failure is a pre-existing, unrelated local-environment issue — a fixed-port test server colliding with something already bound to :8080, same as on other recent PRs).

`cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` both clean.

Closes #834

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed sensitive key information from restore-session logs.
  * Restore-session processing and error handling remain unchanged.

### DIFF
--- a/src/app/restore_session.rs
+++ b/src/app/restore_session.rs
@@ -26,10 +26,10 @@ pub async fn restore_session_action(
         return Err(MostroCantDo(CantDoReason::InvalidPubkey));
     }
 
-    tracing::info!(
-        "Starting background restore session for master key: {}",
-        master_key
-    );
+    // No key in the log line — master_key is the user's persistent Nostr
+    // identity, not a per-trade ephemeral one (AGENTS.md: scrub Nostr keys
+    // from logs).
+    tracing::info!("Starting background restore session");
 
     // Create a new manager for this specific restore session
     let manager = RestoreSessionManager::new();
@@ -99,7 +99,8 @@ async fn send_restore_session_response(
     )
     .await;
 
-    tracing::info!("Restore session response sent to user {}", trade_key,);
+    // No key in the log line (AGENTS.md: scrub Nostr keys from logs).
+    tracing::info!("Restore session response sent to user");
 
     Ok(())
 }
@@ -112,7 +113,8 @@ async fn send_restore_session_timeout(trade_key: &str) -> Result<(), MostroError
     // Send timeout message without payload since Text doesn't exist
     enqueue_restore_session_msg(None, trade_pubkey).await;
 
-    tracing::warn!("Restore session timed out for user: {}", trade_key);
+    // No key in the log line (AGENTS.md: scrub Nostr keys from logs).
+    tracing::warn!("Restore session timed out for user");
 
     Ok(())
 }


arkanoider:
@grunch what do you think? We can remove that printout of keys I suppose. It's a good practise also in the restore case imo.

grunch (approved):
LGTM

grunch:
thank you very much @ToRyVand good catch 😄

ToRyVand:
## Merged — technical summary

Merged by @grunch in `2626941`. Recap for the record:

**Problem.** `AGENTS.md` requires scrubbing logs that could leak invoices or Nostr keys. `restore_session.rs` logged `master_key`/`trade_key` verbatim in 3 places — `master_key` is the user's persistent Nostr identity, and the path runs on every successful restore-session request in production.

**How it was found.** Not from the backlog — surfaced while reviewing CodeRabbit's post-rebase re-review of #826. Verified the lines were pre-existing on `main` (untouched by #826's own diff) before filing #834 publicly; confirmed no `SECURITY.md`/disclosure policy exists in the repo and that the values are Nostr *public* keys, not secrets, so a public issue was the right channel.

**Fix.** Dropped the key from all 3 log sites, replacing with an inline comment citing the `AGENTS.md` rule at each site rather than one comment at the top of the file — keeps the reasoning visible exactly where someone might be tempted to re-add the value for debugging.

**Follow-up filed, not scope-crept in:** a review pass on this fix found the identical pattern unaddressed in 5 more files (`scheduler.rs`, `app.rs` ×2, `last_trade_index.rs`, `db.rs`). Rather than expand this PR, opened #836 proposing a structural fix (redaction layer / lint) instead of 5 more one-off patches — left unclaimed for maintainer input.

**Final:** 1 commit, 1 file, +8/-6. `cargo fmt`/`clippy -D warnings`/`cargo test` clean.

Closes #834.

PR #841, opened 2026-07-27
feat: add Nostr trusted-node price provider (#697)

ToRyVand:
## Summary

Closes #697. Adds a `nostr` price provider so `mostrod` can source BTC/fiat
quotes by subscribing to kind-30078 (NIP-33) rate events published by
trusted Mostro nodes, instead of hitting an HTTP price API — for operators
in regions where those APIs are network-blocked (the motivating case:
`api.yadio.io` is DNS-blocked for Venezuelan ISPs; PR #685 already solved
this for mobile clients, this closes the gap for the node operator).

It slots into the existing multi-source `PriceProvider` registry
(`src/price/`, `docs/PRICE_PROVIDERS.md`) as one more adapter — no changes
to the aggregation core, the store, the scheduler, or any order handler.

- `src/price/providers/nostr.rs` (new): `NostrProvider` queries the
  process-wide Nostr client (already connected to `[nostr]`'s relays — no
  new connections) for `kind=30078, authors=trusted_nodes, d=mostro-rates`
  once per tick, re-verifies each returned event's pubkey against
  `trusted_nodes` client-side, and takes the **freshest** valid event as
  the tick's source. With several `trusted_nodes` configured this is
  redundancy (if one is down/stale, a fresher one wins), not a statistical
  combine across nodes — kept deliberately simple, see `docs/PRICE_PROVIDERS.md`
  §11.7 for the rationale.
- `src/price/manager.rs`, `src/price/providers/mod.rs`: registry wiring
  (`build_provider` arm + module declaration).
- `settings.tpl.toml`: commented `[price.providers.nostr]` template block.
- `docs/PRICE_PROVIDERS.md`: new §11.7 appendix section (the code already
  referenced it) + a note on the Phase table (§8) that this is an
  out-of-band addition to the original publish-only 5-phase plan.
- `src/price/config.rs`: fixes `ProviderConfig::validate()` — the
  `trusted_nodes` exemption from requiring a `url` was previously generic
  across *any* provider id, so e.g. copy-pasting `trusted_nodes` onto
  `[price.providers.yadio]` would silently pass startup validation and only
  fail later with a confusing per-tick HTTP error against an empty URL. Now
  scoped to `nostr` only, with a regression test.

## Test plan

- [x] `cargo build` — clean.
- [x] `cargo test price::` — 117 passed (includes new `parse_content`,
      `select_freshest`, constructor, and registry-wiring tests).
- [x] `cargo clippy --all-targets --all-features` — no issues.
- [x] `cargo fmt --check` — clean.
- [x] **Live-relay evidence**: an `#[ignore]`d test
      (`price::providers::nostr::tests::live_relay_fetch_returns_real_rates`,
      not run in CI) exercises the real `fetch()` path against
      `wss://relay.mostro.network` using the two example `trusted_nodes`
      pubkeys from the issue. Run manually:
      `cargo test price::providers::nostr::tests::live_relay_fetch_returns_real_rates -- --ignored --exact`
      — returned **140 live currencies** (USD, EUR, CUP, MLC, ARS, …),
      confirming the whole path (client → filter → freshest-selection →
      parse) works end-to-end against production infrastructure, not just
      fixtures.

## Summary by CodeRabbit

* **New Features**
  * Added a Nostr-based price provider for BTC/fiat rates using kind `30078` “mostro-rates” events from configured trusted nodes.
  * Selects the freshest valid quote and ignores untrusted or outdated events.
  * Added `nostr` as a selectable price-provider option.
* **Documentation**
  * Added configuration guidance for trusted nodes.
* **Bug Fixes**
  * Improved provider configuration validation and fallback behavior.
* **Tests**
  * Expanded coverage for parsing, freshness, validation, and fallback scenarios.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2320,6 +2320,7 @@ dependencies = [
  "lnurl-rs",
  "mostro-core",
  "mutants",
+ "nostr-relay-builder",
  "nostr-sdk",
  "once_cell",
  "prost 0.14.4",
@@ -2436,6 +2437,23 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ade30de16869618919c6b5efc8258f47b654a98b51541eb77f85e8ec5e3c83a6"
 dependencies = [
  "nostr",
+]
+
+[[package]]
+name = "nostr-relay-builder"
+version = "0.44.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "08ed09e8363503c6f464087b530a1f43cf68be6292bd4928b5ea1141b03dc11c"
+dependencies = [
+ "async-utility",
+ "async-wsocket",
+ "atomic-destructor",
+ "hex",
+ "negentropy",
+ "nostr",
+ "nostr-database",
+ "tokio",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ tokio = { version = "1.47.1", features = ["full", "test-util", "macros"] }
 axum = "0.8.4"
 tower-http = { version = "0.6.6", features = ["cors"] }
 bech32 = "0.11.0"
+nostr-relay-builder = "0.44.1"
 
 [build-dependencies]
 tonic-prost-build = "0.14.1"
--- a/docs/PRICE_PROVIDERS.md
+++ b/docs/PRICE_PROVIDERS.md
@@ -480,6 +480,14 @@ only = ["CUP", "MLC"]    # El Toque is only meaningful for these (§6.6)
 
 Phases 3 and 4 both depend on Phase 2 and can land in either order.
 
+> **Out-of-band addition — Nostr *subscribe* (issue #697, §11.7).** The
+> five phases above only ever covered *publishing* rates to Nostr (Phase 5).
+> A separate, independently-landing addition — the `nostr` provider — makes
+> `mostrod` able to *consume* another Mostro node's published rates instead
+> of an HTTP API, for operators in regions where the price APIs are
+> network-blocked. It slots into the existing registry (§5.4) like any
+> other provider and does not otherwise depend on Phase 5.
+
 ---
 
 ## 9. Phase details
@@ -837,6 +845,52 @@ BTCPayServer solves the same problem; contrasting choices clarify ours:
   premium/fee is the only markup. We call this out so the spread concept
   is not reintroduced by accident when porting a provider idea from
   BTCPay.
+
+### 11.7 Nostr (subscribe, trusted-node relay mode) — issue #697
+
+Every source above is an HTTP API. This one is not: it sources quotes from
+**other Mostro nodes**, over Nostr, for operators whose network blocks the
+HTTP price APIs outright (the motivating case: `api.yadio.io` is DNS-blocked
+for Venezuelan ISPs) but who can still reach a Nostr relay.
+
+- **Not a new publish path.** Mostro nodes already publish their aggregated
+  rates as a kind-30078 NIP-33 event (`d = "mostro-rates"`, content
+  `{"BTC": {ccy: price}}`) — see `docs/NOSTR_EXCHANGE_RATES.md` and
+  `nip33::new_exchange_rates_event`. The `nostr` provider is a **consumer**
+  of that exact same event, from other operators' nodes instead of its own.
+- **Config (`[price.providers.nostr]`, §7):** no `url` — `trusted_nodes` (a
+  list of hex pubkeys) is the provider's only required field. Enabled
+  without at least one valid `trusted_nodes` pubkey is a startup error
+  (mirrors El Toque's missing-token fail-fast, §7).
+- **Relays are not reconfigured.** The provider reuses the process-wide
+  Nostr client already connected to the node's own `[nostr]` relays — it
+  does not open a second set of connections or accept a separate relay list.
+- **One-shot query per tick, not a persistent subscription.** `fetch()` is
+  called once per `update_interval_seconds` tick like every other provider
+  (spec §5.3); it issues a single `kind=30078, authors=trusted_nodes,
+  #d=mostro-rates` relay query bounded by a short fixed timeout, rather than
+  keeping an open `REQ` subscription running between ticks.
+- **Freshest wins — no cross-node combine.** With several `trusted_nodes`
+  configured, the provider does **not** run the §6.2 median+outlier combine
+  across their individual events; it takes the single event with the
+  highest `created_at` among the trusted, valid ones. Rationale (confirmed
+  with the maintainer): the issue frames multiple `trusted_nodes` as
+  *redundancy* (if one is down or stale, use another), not as independent
+  samples to statistically combine, and skipping a second aggregation layer
+  inside a single provider keeps the adapter's contract identical to every
+  HTTP provider (one `ProviderQuotes` map out of `fetch()`, no new state).
+  A compromised single trusted node can therefore still push a bad price
+  with no in-provider outlier guard; operators mitigate by choosing
+  `trusted_nodes` they actually trust, same as any other single-sourced
+  provider.
+- **Client-side pubkey verification.** Even though the relay-side `authors`
+  filter already restricts the query, the adapter re-checks each returned
+  event's `pubkey` against `trusted_nodes` before trusting its content —
+  the same verification `docs/NOSTR_EXCHANGE_RATES.md` requires of mobile
+  clients, applied here on the operator side.
+- **Currency codes are re-upper-cased** (§6.6) even though a Mostro
+  publisher already emits uppercase codes — a third-party trusted node is
+  outside this codebase's control.
 
 ---
 
--- a/settings.tpl.toml
+++ b/settings.tpl.toml
@@ -180,6 +180,17 @@ port = 50051
 # url = "https://tasas.eltoque.com"
 # # token = "xxxx"        # REQUIRED when enabled; provider refuses to start otherwise
 # only = ["CUP", "MLC"]
+#
+# # Subscribe to prices published (kind 30078) by trusted Mostro nodes over
+# # Nostr instead of an HTTP API — for operators in regions where price APIs
+# # are DNS/IP blocked (docs/PRICE_PROVIDERS.md §11.7). Reuses the relays
+# # already configured in [nostr]; no url needed. With several trusted_nodes,
+# # the freshest valid event wins — not a cross-node combine.
+# [price.providers.nostr]
+# enabled = false
+# trusted_nodes = [
+#     # hex pubkeys of Mostro nodes you trust to publish accurate rates
+# ]
 
 # Anti-abuse bond (issue #711). Opt-in, disabled by default. Uncomment to
 # require a Lightning hold-invoice bond from takers and/or makers. See
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -30,6 +30,14 @@ pub use types::{
 // They are shared across the application using Arc and Mutex/RwLock for thread safety
 pub static MOSTRO_CONFIG: OnceLock<Settings> = OnceLock::new();
 pub static NOSTR_CLIENT: OnceLock<Client> = OnceLock::new();
+/// Dedicated Nostr client for the price provider's kind-30078 queries.
+///
+/// Kept separate from [`NOSTR_CLIENT`] so `verify_subscriptions(true)` (and
+/// the REQ `limit` it enforces) applies only to price fetches — not to the
+/// daemon's long-lived `.limit(0)` inbox subscription in `main.rs`, where
+/// pre-EOSE filter verification would reject matching trade messages
+/// (hermeme, PR #841).
+pub static PRICE_NOSTR_CLIENT: OnceLock<Client> = OnceLock::new();
 pub static LN_STATUS: OnceLock<LnStatus> = OnceLock::new();
 pub static DB_POOL: OnceLock<Arc<sqlx::SqlitePool>> = OnceLock::new();
 
--- a/src/price/aggregate.rs
+++ b/src/price/aggregate.rs
@@ -2,7 +2,7 @@
 //! function is a deterministic transform over in-memory inputs, which is
 //! what makes the numeric heart of the feature exhaustively testable.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use super::provider::{ProviderId, ProviderQuotes, Quote};
 
@@ -164,7 +164,11 @@ pub fn aggregate_tick(
         }
     }
     let mut resolved: HashMap<String, Vec<(ProviderId, f64)>> = HashMap::new();
-    let mut resolved_nostr_anchor: HashMap<String, bool> = HashMap::new();
+    // Providers whose *resolved* candidate for a currency came from a
+    // Nostr-touched anchor. Taint is applied only if that provider later
+    // survives the target currency's outlier filter (step 3) — a discarded
+    // wild cross must not mark an otherwise-local aggregate dependent.
+    let mut nostr_anchored_resolvers: HashMap<String, HashSet<ProviderId>> = HashMap::new();
     for (id, currency, base, value) in &per_base {
         if let Some(anchor) = anchors.get(base) {
             let candidate = value * anchor;
@@ -174,7 +178,10 @@ pub fn aggregate_tick(
                     .or_default()
                     .push((*id, candidate));
                 if *anchor_uses_nostr.get(base).unwrap_or(&false) {
-                    resolved_nostr_anchor.insert(currency.clone(), true);
+                    nostr_anchored_resolvers
+                        .entry(currency.clone())
+                        .or_default()
+                        .insert(*id);
                 }
             }
         }
@@ -201,13 +208,16 @@ pub fn aggregate_tick(
                 .filter(|x| x.is_finite() && **x > 0.0)
                 .count()
                 .min(u8::MAX as usize) as u8;
+            let nostr_anchor_dependent = nostr_anchored_resolvers
+                .get(currency)
+                .is_some_and(|tainted| contributors.iter().any(|c| tainted.contains(c)));
             out.insert(
                 currency.clone(),
                 AggregateResult {
                     value,
                     sources,
                     contributors,
-                    nostr_anchor_dependent: *resolved_nostr_anchor.get(currency).unwrap_or(&false),
+                    nostr_anchor_dependent,
                 },
             );
         }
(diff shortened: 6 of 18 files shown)


ToRyVand:
Pushed \`b7d5e4c\` addressing both CodeRabbit nitpicks:
- \`RELAY_QUERY_TIMEOUT\` is no longer a fixed 8s constant — \`NostrProvider\` now derives its relay-query timeout from the shared \`provider_timeout_seconds\` (threaded through \`build_provider\`), so it stays in sync if that setting changes instead of risking the outer \`poll_budget\` timeout cutting it off first.
- Extracted the \`Filter\` construction into \`NostrProvider::build_filter()\`, now covered by a plain unit test asserting kind/authors/identifier — no longer only exercised by the \`#[ignore]\`d live-relay test.

\`cargo build\`, \`cargo test\` (11 passed, 1 ignored), \`cargo clippy --all-targets --all-features -- -D warnings\`, \`cargo fmt --check\`: all clean.

arkanoider:
### Maintainer follow-up: discard stale Nostr rate events

Pushed `dff055e3` onto this branch.

**Problem.** `get_price` / the store TTL only look at `as_of` (when *this* node last wrote the aggregate). On every successful Nostr poll we set `as_of = now`, even if the trusted-node kind-30078 event is days old. Relays often keep expired NIP-40 events, so a dead or idle trusted node could keep refreshing the cache clock with zombie rates — especially bad for the nostr-only / API-blocked case this provider targets.

**Fix.** `NostrProvider::select_freshest` now also requires `now - created_at ≤ [price].max_price_staleness_seconds` (same shared TTL as the store). If every trusted event is missing, untrusted, or too old, `fetch` returns `Err` so the tick counts as a provider failure (breaker / other sources / last-known-good) instead of an empty success that still looks “fresh”.

**Wiring.** `max_price_staleness_seconds` is passed into `NostrProvider::new` the same way `provider_timeout_seconds` already is — no new public `from_settings` API.

**Tests.** Unit coverage for discard-older, all-stale → none, and exact max-age boundary.

arkanoider:
@ToRyVand tACK and good job! I posted a fix to avoid staleness of nostr event that was not checked, take a look on your side and then for me it's green for merge.

ToRyVand:
Thanks @arkanoider — the TTL fix in `dff055e3` looks right, and it closes a real gap the original PR left open.

CodeRabbit caught one more edge in that same commit worth fixing before merge: `select_freshest`'s freshness check (`now.saturating_sub(created_at) <= max_age`) only rejects events that are *too old*. A future-dated `created_at` (forged or clock-skewed relay) saturates that subtraction to 0, passes the check trivially, and then wins `max_by_key` outright — the exact zombie-rate scenario the TTL fix was meant to close, just from the other direction.

Pushed `e52f93f`: rejects `created_at > now` before the age check, plus a test (`select_freshest_rejects_future_dated_events`) asserting a future-dated event loses to a legitimately current one. `cargo fmt --check`, `clippy -D warnings`, and the full `price::` test suite (123 tests) are all green.

Should be good for another look whenever you get a chance.

arkanoider:
@ToRyVand the bot seems ranting again, take a look and verify if it's a real issue, reading quickly they seem correct. I missed in my fixed the `Event::is_expired_at` thing. Take a look again please, and fix. We are more than happy to merge this new nostr currency source.

arkanoider:
@ToRyVand the bot are bad guys, still ranting on other points...

ToRyVand:
Pushed `d5e91de` addressing all 4 points from the strict re-review.

**2 & 4 (nostr.rs) — mechanical, fully closed:**
- `select_freshest` → `rank_candidates` (returns every valid candidate, newest first) + `pick_first_usable` (tries them in order until one parses to a non-empty rate map). A malformed/empty body from the newest trusted event no longer suppresses a good older one — 3 new tests cover this.
- `rank_candidates` now revalidates `kind`/`d` client-side instead of trusting the relay-side query filter, since `verify_subscriptions` defaults to `false` in the locked `nostr-relay-pool` — 2 new tests cover a wrong-kind and a wrong-`d` trusted event.

**3 (manager.rs) — Nostr made fallback-only:** `restrict_nostr_to_fallback` drops the Nostr provider's quotes for any currency another provider already covered that tick, before `aggregate_tick` runs. A trusted node's rate is itself already a combine of its own direct sources, so blending it in alongside this node's direct sources was double-counting correlated info. 2 new tests.

**1 (manager.rs) — republish guard, partial:** `republishable_rates` stops republishing any currency whose only surviving contributor is `Nostr` — a re-stamped `created_at`/`expiration` on a purely-relayed rate was erasing how old the underlying quote really is. This closes the common case, but I want to flag its limit honestly: it's a mitigation, not full provenance tracking. A currency that's genuinely Nostr-only forever (an operator fully cut off from direct APIs, which is this feature's actual target user) still won't propagate past one hop under this fix — by design, since we don't know its true age without a separate origin-timestamp tag. Closing that fully needs a NIP-40-adjacent protocol addition (an origin `created_at` distinct from the outer event's), which touches `docs/NOSTR_EXCHANGE_RATES.md` and is really a spec-level follow-up, not something I want to improvise into this PR under time pressure. Happy to open a tracking issue for it if you agree with that split.

Verification: `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`, full `price::` suite (134 tests, up from 123) all green.

arkanoider:
tACK for me! Let's wait a pass from a bot to complete it.

arkanoider:
## Local verification of ermeme blockers (head `d5e91deb`)

Checked out the same commit as the PR head (`d5e91debbd071fe65afe70cfb9cd8329e2cb227a`) and walked every point that was still marked **blocking** in ermeme's reviews (including the earlier NIP-40 item that was already closed on `c493844`).

### Verdict: all requested blockers are satisfied on this head

| # | Blocker (ermeme) | Status on `d5e91de` | Where |
|---|------------------|---------------------|--------|
| 0 | NIP-40 expired events still eligible | **Fixed** (already on `c493844`) | `rank_candidates` filters `!e.is_expired_at(&now)`; tests cover newest-expired vs older-valid and all-expired |
| 1 | Indefinite republish / trust-cycle freshness | **Fixed** (option A from the review) | `republishable_rates` drops currencies whose only surviving contributor is `Nostr` before `publish_rates_to_nostr` |
| 2 | Newest malformed/empty event suppresses a valid redundant node | **Fixed** | `rank_candidates` (newest-first) + `pick_first_usable`; tests for malformed, empty, and all-unusable |
| 3 | Remote aggregate counted as an independent local vote | **Fixed** (fallback-only) | `restrict_nostr_to_fallback` in `aggregate_tick` keeps Nostr only for currencies no other provider covered that tick |
| 4 | Kind / `d` not revalidated client-side | **Fixed** | `rank_candidates` requires `kind == 30078` and `d == mostro-rates`; tests for wrong-kind and wrong-`d` |

### Notes

- Blocker **1** is closed via the review's first alternative ("prevent republishing Nostr-derived-only aggregates as fresh observations"), not full origin-timestamp provenance. That matches the author's stated scope split; a protocol-level origin age tag remains a reasonable follow-up for multi-hop Nostr-only operators, but it is **not** required to clear this review item as written.
- Blocker **3**'s concrete double-count example is addressed by dropping overlapping Nostr quotes before aggregation (direct sources alone decide the value when they cover the currency).

### Verification run locally

- `cargo test price::` → **134 passed**, 0 failed, 1 ignored

tACK from my side on the correctness items above — the remaining open review from ermeme[bot] looks stale relative to `d5e91de`.

grunch:
@coderabbitai review

arkanoider:
@ToRyVand i think we are ready for merge, fix last coderabbit nipticks and we should be ok!

ToRyVand:
## CodeRabbit items addressed — plus one new finding worth your call

Pushed two commits on top of `d5e91de`, kept separate so the correctness change is reviewable on its own.

**`2de2902` — case-insensitive fallback coverage** (the actionable comment on `manager.rs:613`)

`restrict_nostr_to_fallback` compared raw currency keys while `aggregate_tick` upper-cases every code before grouping. Not every adapter normalises on the way out — `yadio.rs:50` forwards the API's codes verbatim — so a direct `usd` and a Nostr `USD` slipped past the filter and then landed in the same aggregate. I confirmed the added regression test fails on the raw comparison and passes with the fold.

**`4c58f07` — the two nitpicks**

The `source` tag is now derived from the currencies that survive `republishable_rates` rather than the tick-wide contributor list, so it can no longer name a provider absent from the published body. That made the `successes` parameter redundant, so it is gone. Also renamed the leftover `select_freshest` references to `rank_candidates`, matching the prefix the envelope tests already use.

Gates: `cargo test price::` 136 passed / 1 ignored (was 134), `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean.

---

### New finding — an unresolvable `PerBase` suppresses a usable Nostr quote

This surfaced while verifying the above, and I reproduced it on the pushed head.

`restrict_nostr_to_fallback` measures coverage from raw quote keys at `manager.rs:298`, *before* `aggregate_tick` runs at `manager.rs:300`. El Toque covers CUP/MLC only as USD-anchored `PerBase` quotes, and `aggregate.rs:93` drops a `PerBase` silently when its anchor is missing. So El Toque "covers" CUP, Nostr's usable direct `PerBtc` CUP is discarded, El Toque then fails to resolve, and CUP ends up with **no price at all**.

Repro — El Toque CUP as `PerBase { base: "USD", value: 300.0 }`, Nostr CUP as `PerBtc(30_000_000.0)`, no direct USD that tick — yields `aggregates = {}`.

The trigger is narrow: it needs a trusted rate event carrying CUP but not USD. But the failure mode is *no price* rather than a wrong one, in exactly the blocked-API deployment §11.7 targets.

I did not fix it here, because both obvious minimal fixes are wrong:

- Letting only `PerBtc` establish coverage re-introduces the Nostr/direct blending that this restriction exists to prevent, for every currency whose cross *does* resolve.
- Pre-aggregating the non-Nostr providers to derive coverage misjudges it in precisely the deployment at issue, where the USD anchor legitimately comes from Nostr itself.

A correct fix means determining coverage after anchor resolution — a real restructure and a new review surface on a PR that is otherwise ready. Happy to do it in this PR if you would rather not ship the gap, or to file it as a follow-up. Your call.

### Note on the accepted `as_of` deferral

Not re-opening this, just putting a number on it: `nostr.rs:98` sets the provider's `max_age` **to** `max_price_staleness_seconds`, and the store stamps `as_of = now` at `manager.rs:321`. So an event accepted at 1799s old is re-stamped fresh and served for another 1800s — roughly 2× the configured budget on the defaults. That is the provenance item already accepted as a follow-up; flagging the magnitude only.

arkanoider:
## Addressed ermeme's three remaining blockers (`a8a6aef2`)

Pushed a follow-up on top of `4c58f070` that closes the three items from [the latest strict re-review](https://github.com/MostroP2P/mostro/pull/841#pullrequestreview-4839023893).

### 1. Unresolved `PerBase` no longer suppresses a usable Nostr fallback

**Fix.** `restrict_nostr_to_fallback` now decides coverage **after** fiat-cross resolution via `currencies_covered_by_non_nostr`:

- Anchors are built from *all* direct `PerBtc` quotes (including Nostr), so a local El Toque CUP/USD that needs Nostr's USD still counts as covering CUP.
- A non-Nostr `PerBase` only covers a currency when `value × anchor` is finite and positive.
- Raw key presence alone is no longer enough — El Toque CUP without a USD anchor no longer drops Nostr's usable `PerBtc` CUP.

**Tests.** `restrict_nostr_to_fallback_keeps_nostr_when_per_base_cannot_resolve`, `restrict_nostr_to_fallback_drops_nostr_when_per_base_resolves_via_nostr_usd`.

### 2. Nostr-anchor taint tracked through cross resolution

**Fix.** `AggregateResult` gains `nostr_anchor_dependent`. `aggregate_tick` sets it when a `PerBase` resolves against an anchor whose surviving contributors include `Nostr`. `republishable_rates` now drops those currencies as well as pure Nostr-only ones — so El Toque CUP × Nostr USD is not re-stamped as a fresh `source=eltoque` observation.

`contributors` still names only the cross provider (anchor providers remain intermediates for the `source` tag).

**Tests.** `aggregate_tick_marks_cross_resolved_via_nostr_usd_as_anchor_dependent`, `republishable_rates_drops_nostr_anchor_dependent_cross`.

### 3. Local bound on relay event collection

**Fix.** `fetch` streams via `Client::stream_events` instead of unbounded `fetch_events`/`force_insert`. While receiving it:

- early-rejects non-trusted / wrong-kind / wrong-`d` frames (`is_plausible_candidate`);
- hard-caps retained candidates at `MAX_FETCHED_RATE_EVENTS` (64);
- also caps inspected frames under a flood of junk;
- still sets `Filter::limit` for compliant relays.

**Tests.** `collect_bounded_candidates_drops_junk_and_caps_trusted_envelope`.

### Verification

- `cargo fmt --check` / `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test price::` → **141 passed**, 1 ignored

Happy to adjust further if anything still looks off on a re-review of `a8a6aef2`.

arkanoider:
Hi @ToRyVand I created a commit to fix latest @hermemes bot rants, I will check them better tomorrow with my eyes. Please take a look and evaluate also on your side.

ToRyVand:
## Reviewed `a8a6aef2` — both aggregation fixes are correct

Checked both against the current head. Building anchors from *all* direct `PerBtc` including Nostr is the key move in `currencies_covered_by_non_nostr` — it covers the local-cross-on-relayed-USD case I couldn't close minimally when I raised it. And keeping the anchor's providers out of `contributors` was right: the `source` tag keeps meaning "whose value survived" instead of getting overloaded. Both have real tests. Nothing I'd change.

Three things I'd still touch.

### 1. The inspected-frame cutoff should just go (`nostr.rs:271-285`)

ermeme's starvation point is right, but I'd delete rather than refine. `inspected` counts across the *merged* pool stream, so one noncompliant relay can spend all 256 frames before an honest relay's trusted event is yielded, failing the tick.

And the cutoff isn't buying what it was added for: breaking out of this loop doesn't stop the pool driver — it keeps consuming until the timeout regardless (see 3). So it costs a starvation path and saves only some cheap loop iterations. The retained-candidate cap of 64 is the one doing real work.

```diff
         let mut events = Vec::new();
-        let mut inspected = 0usize;
         while let Some(event) = stream.next().await {
-            inspected += 1;
             if Self::is_plausible_candidate(&event, &self.trusted_nodes) {
                 events.push(event);
                 if events.len() >= MAX_FETCHED_RATE_EVENTS {
                     break;
                 }
             }
-            // Bound work even when the relay floods irrelevant frames.
-            if inspected >= MAX_FETCHED_RATE_EVENTS.saturating_mul(4) {
-                break;
-            }
         }
```

### 2. Taint is set before the final outlier filter (`aggregate.rs:168-181` vs `:197`)

Correct, low severity. `resolved_nostr_anchor` is inserted in step 2; the `combine` that can discard that cross as an outlier runs in step 3. So a currency whose final value is entirely local can still be marked `nostr_anchor_dependent` and dropped from republication. Fails safe, but it's a silent republication gap. Deriving the flag from the pairs surviving the outlier predicate (the `kept_contributors` pass already there) closes it, plus a test.

### 3. The relay-pool dedup is real, but I don't think it gates this PR

I read `nostr-relay-pool 0.44.1` rather than take the review at face value, and the mechanism is as described: unbounded `HashSet<EventId>` in `stream_events_targeted`, `task::spawn`ed driver, `let _ = tx.send(event).await` — dropping our receiver doesn't stop it.

The severity framing is what I'd push back on. What grows is the ID set; delivered events are dropped once the receiver is gone. And it's bounded and transient in a way a real OOM isn't:

- `provider_timeout_seconds` (default `10`) hard-caps how long the driver runs, and the set is freed when it ends — nothing accumulates across ticks.
- Every frame pays signature verification before reaching the dedup, throttling arrival.
- The hostile party has to be a relay in the operator's own `nostr_settings.relays` — a chosen trust boundary, not an open surface.

(No megabyte figure attached on purpose — I haven't benchmarked one.)

More to the point, it isn't a price-provider bug: any pooled query through this client has the same property, and fixing it properly means subscription-filter enforcement or a bounded relay-level receive path — a dependency-layer change with its own review surface.

### Proposed split

- **This PR:** (1) and (2) — one deletion, one small flag-placement fix with a test.
- **Follow-up issue:** the relay-pool dedup bound, against the Nostr client layer where the fix actually lives.

Does that split work? Happy to push (1) and (2) once you've done your own pass on `a8a6aef2` — didn't want to move the head out from under the review you said you'd do.

arkanoider:
When the bot will be happy on this @ToRyVand and we finally merge we deserve a beer! :smile:

arkanoider:
## Addressed the remaining stream-boundary blocker (`a745bebd`)

Reviewed [the latest CHANGES_REQUESTED](https://github.com/MostroP2P/mostro/pull/841#pullrequestreview-4839739517) against `nostr-relay-pool 0.44.1` — **the finding is real**, not a hallucination. `RelayPool::stream_events_targeted` does keep an unbounded `HashSet<EventId>` in a spawned driver that ignores a dropped consumer, so an application-only retained-candidate cap was not a sufficient bound.

Also incorporated the two follow-up edges ToRyVand called out on that head.

### 1. Bound at / before the relay-pool driver

**Fix.** `connect_nostr` now uses `mostro_nostr_client_options()`, which sets `verify_subscriptions(true)`. Non-matching frames are rejected inside the relay driver *before* they can enter the pool dedup set. `build_filter()` carries `limit(MAX_FETCHED_RATE_EVENTS)` so that ceiling is enforced there too (with `verify_subscriptions` the driver also rejects past-limit deliveries pre-EOSE).

App-level early-filter + retained-candidate cap remains as defense in depth.

### 2. Removed the inspected-frame cutoff

**Fix.** Dropped the merged-stream `inspected >= 256` break. One noisy relay could fill that global first-N window before an honest relay’s trusted event arrived (provider-level DoS). It also did not stop the pool driver. The retained-candidate cap of 64 stays.

### 3. Nostr-anchor taint only after outlier survival

**Fix.** `nostr_anchor_dependent` is now true only when a Nostr-anchored cross provider actually survives the target currency’s `kept_contributors` pass — a discarded wild cross no longer blocks republication of an otherwise-local aggregate.

**Test.** `aggregate_tick_nostr_anchor_taint_clears_when_cross_is_outlier`.

### Verification

- `cargo test price::` → **142 passed**, 1 ignored

Happy to take another pass on `a745bebd`.

arkanoider:
Agreed with [@hermemes](https://github.com/hermemes) on [this review](https://github.com/MostroP2P/mostro/pull/841#pullrequestreview-4852382343): `verify_subscriptions(true)` must not be process-wide.

**`0464dfc6`** scopes the bound to Nostr price queries only:

1. **Daemon client** (`connect_nostr` / `NOSTR_CLIENT`) keeps size limits but leaves `verify_subscriptions` **off**, so the long-lived `.limit(0)` inbox in `main.rs` is not charged pre-EOSE frames against a zero limit.
2. **Price client** (`PRICE_NOSTR_CLIENT` / `get_price_nostr_client` / `price_nostr_client_options`) enables `verify_subscriptions` and is used only by `NostrProvider::fetch`. Lazy-init on first price tick.
3. **Observable policy** (`MostroNostrClientPolicy`) so tests assert `verify_subscriptions` on/off directly — `ClientOptions` fields are not publicly readable.
4. **Local-relay regressions** via `nostr-relay-builder` `MockRelay`:
   - price client drops mismatched frames (`send_random_events`)
   - price client enforces REQ `limit` before pool forwarding
   - daemon `limit(0)` still delivers a matching live event after EOSE

Publishing kind-30078 rates still uses the daemon client (`get_nostr_client`); only the price *query* path uses the verified client.

ToRyVand:
Ran the last gate — `28bbe62` is `cargo fmt --all` and nothing else.

Three wrapping hunks in `src/util.rs`'s new relay tests, no behaviour touched. `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test price::` (142 passed, 1 ignored) are all clean on that head.

Nothing left for you to run — didn't want you spending a round trip on rustfmt.

On my side there's nothing outstanding: `a745beb` already covers both edges I raised on `a8a6aef2`, and `0464dfc` is the right call on scoping — the `.limit(0)` daemon inbox would have been charged pre-EOSE frames against a zero limit if verification had stayed process-wide. Nice catch keeping publishing on the daemon client and only the query path on the verified one.

Beer stands. 🍻

arkanoider:
We need the final sentence from @grunch !

grunch (approved):
LGTM! let's go!!!

Thanks @ToRyVand great job!

ToRyVand:
Merged as `a8e923bb` — thanks @grunch for the final call and @arkanoider for carrying a lot of this yourself.

Closing recap for whoever reads this later.

**What landed.** A `NostrProvider` in the existing price-provider registry: operators whose price APIs are network-blocked can source BTC/fiat quotes from the kind-30078 rate events Mostro nodes already publish, taking the freshest valid event from a configured `trusted_nodes` list. 12 commits, +1850/-44 across 18 files, 9 days, 14 reviews.

**The interesting part was everything after the first commit.** Five review rounds, two review bots and two maintainers, and most of the substance was correctness rather than style:

- **NIP-40 / freshness** — selection honoured the local TTL but not a rate event's own expiration tag, and a future-dated `created_at` saturated the staleness check to zero and always won. Both closed with boundary tests.
- **Correlated sourcing** — the relayed value is itself an aggregate of some other node's sources, so counting it as an independent vote double-counted the same information into the local median. It is now strictly fallback-only, and coverage is decided *after* fiat-cross resolution so an unresolvable `PerBase` cannot suppress a quote that would actually price the currency.
- **Provenance** — republishing a Nostr-derived rate re-stamped it fresh, which a trust cycle could ride indefinitely. Nostr-only and Nostr-anchored currencies are now excluded from republication, with the anchor taint read off the candidates that survive outlier filtering rather than stamped at resolution time.
- **Relay bounds** — the pool's dedup set is unbounded in a driver that ignores a dropped consumer. @arkanoider closed this properly at the relay layer with `verify_subscriptions` on a dedicated price client, keeping the daemon inbox's `.limit(0)` live-delivery behaviour intact. That is the fix I had argued belonged in the client layer rather than here, and it turned out to be both — done here, but at the right layer.

Four of the twelve commits are @arkanoider's, including the last three that closed out the final round. Credit where it is due.

**One follow-up worth filing.** `nostr.rs` sets the provider's `max_age` *to* `max_price_staleness_seconds` while the store stamps `as_of = now`, so an event accepted at 1799s old is served for another 1800s — roughly 2× the configured budget on defaults. @arkanoider already called a protocol-level origin timestamp "a reasonable follow-up" during review, and this is that. Happy to open it as its own issue unless you would rather it stay in the backlog notes.

Configuration and the deployment model are documented in `docs/PRICE_PROVIDERS.md` §11.7.

PR #843, opened 2026-07-28
fix: bound outbound network latency in message handling

grunch:
## What

Bounds how long a single message handler can wait on an outbound network call, and stops doing one of those calls where it isn't needed yet.

Message handlers run sequentially on one event-loop task (`app::run`), so any request a handler makes delays every other message behind it for the full duration. A few of those requests had no effective upper bound, which left overall responsiveness dependent on third-party hosts the node doesn't control — a slow or unreachable LNURL endpoint, or an LND that isn't answering, would hold up unrelated traffic.

## Changes

**1. Cheap checks first in `order_action`**

Fiat currency, fiat amount, sats amount, range limits and the quote lookup are all local. They now run before the payment request is looked at, so a malformed order is rejected without any further work.

**2. Order creation validates the payment request without network I/O**

`validate_invoice` takes a new `InvoiceCheck` argument:

- `Offline` — syntax plus the full local BOLT11 rules, no HTTP.
- `Online` — additionally resolves lightning addresses and LNURLs, as before.

Order creation uses `Offline`. It only records the payment request (the validation result was already discarded), and a lightning address still has to resolve when the payout is actually made — which is where an unreachable host is worth acting on. `take_sell` and `add_invoice` keep using `Online`: there the buyer is committing to a payout destination, so reachability is genuinely part of the decision. **No change in behavior for those two paths.**

**3. Real bounds on the remaining calls**

- LNURL: a total wall-clock budget (`LNURL_TOTAL_BUDGET`, 5s) covering every round-trip in one operation, a 2s connect timeout, per-request timeout 10s → 4s, and a finite redirect chain (3). The per-request timeout alone wasn't sufficient, since `resolv_ln_address` makes two sequential requests where the second host comes from the first response.
- `LndConnector::new`: a 10s connect timeout, where previously there was none. The RPCs after it are deliberately left unbounded — a timeout on `settle_invoice` would report failure for an operation LND may well have completed, which is a worse failure mode than waiting.

## Test plan

- [x] `cargo test` — 1051 passed, 0 failed
- [x] `cargo fmt --all`

New coverage:

- `lnurl_validation_against_unresponsive_host_is_time_bounded` — a host that completes the TCP handshake and then never answers cannot hold validation open past the budget. Uses an ephemeral port via LNURL rather than the fixed 8080 that `start_test_server` owns, so it doesn't race the existing tests.
- `offline_validation_of_lnurl_does_no_network_io` — the offline path returns immediately against that same unresponsive host.
- `offline_validation_still_rejects_malformed_payment_request` / `..._still_enforces_bolt11_amount_checks` — dropping the round-trip doesn't drop any local check.
- `validate_invoice_paths` now runs its whole table under both modes, pinning down that they agree on every BOLT11 case.

Manual check: create an order with a lightning address on a host that is slow to respond, and confirm the order is created without the daemon pausing; then take a sell and confirm an unreachable address is still rejected at that point.

## Follow-ups (not in this PR)

- Per-pubkey throttling in `SpamGate` — it currently has no counter or token bucket, and the existing gate only applies to the `nip44` transport while `gift-wrap` is still the default.
- `show_hold_invoice` opens a fresh `SqlitePool` per take (`util.rs`) instead of reusing `ctx.pool()`.
- Moving slow work off the event loop properly. Worth noting that naive per-message `tokio::spawn` is *not* the answer: the current serialization is what keeps the order state machine free of races, so that needs per-order locking designed alongside it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved invoice validation to reject unreachable Lightning payout destinations earlier during online payment flows.
  - Order creation now performs local checks before network-dependent validation, reducing unnecessary delays.
  - Added offline validation for Lightning Addresses and LNURLs when network access is not required.
  - Added safeguards so LNURL lookups and Lightning node connections time out instead of hanging indefinitely.
  - Preserved payment amount, fee, expiration, and format checks for Lightning invoices.

### DIFF
--- a/src/app/add_invoice.rs
+++ b/src/app/add_invoice.rs
@@ -1,7 +1,7 @@
 use crate::app::context::AppContext;
 use crate::util::{
     enqueue_order_msg, get_order, notify_taker_reputation, show_hold_invoice, update_order_event,
-    validate_invoice,
+    validate_invoice, InvoiceCheck,
 };
 use mostro_core::db::Crud;
 use mostro_core::prelude::*;
@@ -51,7 +51,12 @@ pub async fn add_invoice_action(
         return Err(MostroCantDo(CantDoReason::InvalidPeer));
     }
     // We save the invoice on db
-    order.buyer_invoice = validate_invoice(&msg, &order).await?;
+    //
+    // `Online` keeps the pre-existing behavior: the buyer is submitting the
+    // destination their payout will go to, so an unreachable one should be
+    // rejected now rather than at release. Bounded by
+    // `lnurl::LNURL_TOTAL_BUDGET`.
+    order.buyer_invoice = validate_invoice(&msg, &order, InvoiceCheck::Online).await?;
     // Buyer can add invoice orders with WaitingBuyerInvoice status
     match ord_status {
         Status::SettledHoldInvoice => {
--- a/src/app/order.rs
+++ b/src/app/order.rs
@@ -1,6 +1,6 @@
 use crate::app::context::AppContext;
 use crate::db::update_user_trade_index;
-use crate::util::{get_bitcoin_price, publish_order, validate_invoice};
+use crate::util::{get_bitcoin_price, publish_order, validate_invoice, InvoiceCheck};
 use mostro_core::prelude::*;
 use nostr_sdk::prelude::*;
 use nostr_sdk::Keys;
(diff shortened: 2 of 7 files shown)


grunch:
Superseded by #845, which reworks this branch after review.

The change to skip network validation at order creation rested on a wrong assumption: the payment request supplied there is not a throwaway value — it is persisted as `orders.buyer_invoice` and is what the payout is later sent to. Skipping the check there is only sound once a resolution failure at payout time is itself recoverable, which it was not. #845 fixes that first, then applies the rest on top, and adds the ordering fix in `add_invoice_action` and the removal of a now-unreachable timeout in `dev_fee` that this branch missed.

Issue #844, opened 2026-07-28
Per-pubkey rate limiting in SpamGate (protocol v2)


grunch:
## Context

`SpamGate` (`src/spam_gate.rs`) currently holds two pieces: the active-trade-pubkey cache and the replay guard. Both are membership checks — *is this key known*, *have I seen this event id* — and neither counts anything. There is no rate control anywhere on the message path. The only `RateLimiter` in the tree (`src/rpc/rate_limiter.rs`) guards RPC password attempts and is not reachable from the event loop.

That leaves a gap the current design assumes someone else fills. `docs/TRANSPORT_V2_SPEC.md` §6 names "relay-side rate limiting" as part of the toll (lines 22–26), and the module doc in `src/spam_gate.rs:19-20` defers to it explicitly. In practice we can't rely on that: relay policy varies per operator, a node reads from several relays at once, and none of them know how expensive a given message is for *us* to process.

Proof-of-work doesn't close it either. PoW prices the *first* message, not the tenth — once ground, a sender can reuse the same key freely. And the fast lane makes this sharper: a key in the active-trade cache skips `pow_first_contact` entirely (`src/app.rs:342`) and only faces the base `pow`, which is `0` in the shipped template (`settings.tpl.toml:54`).

The underlying asymmetry is ordinary but worth stating: sending a message costs a signature, while handling one can cost database work, Lightning RPCs and outbound HTTP. #843 bounds how expensive any single message can get; this issue is about bounding *how many* a single sender can ask for.

## Why now

Protocol v2 is what makes this implementable at all. On `gift-wrap` the outer author is a throwaway key with no reusable identity, so there is nothing stable to count against — which is exactly why the existing gate is v2-only. On `nip44` the visible author is the trade key, and the spec calls this out as the point of the transport (§lines 22, 53): a visible envelope is a *rate-limitable* one.

Since v2 becomes the only transport in 0.19 (#786, `transport` is already `#[deprecated]` at `src/config/types.rs:508`), work here lands directly on the path we're keeping. Nothing needs a `gift-wrap` equivalent, and nothing written now gets thrown away at the cutover. This slots naturally alongside the Phase 2 gates.

## Design

A token bucket per trade pubkey, with two tiers mirroring the two lanes the gate already has.

[...]

Issue #837, opened 2026-07-23
[MEDIUM] LNURL: no host validation — buyer-supplied lightning address can reach internal/link-local hosts (SSRF)


grunch:
### Severity: MEDIUM

PR #820 added a scheme allow-list (`http`/`https`) to `extract_lnurl` and `build_callback_url`, closing scheme confusion (`javascript:`, `file:`, ...). The doc comment on `build_callback_url` explicitly notes the remaining gap:

> This blocks scheme confusion (`javascript:`, `file:`, ...) but not host-based SSRF — a malicious server can still return an `http(s)` callback pointed at a private or link-local address; that's a known gap, tracked separately.

This issue is that "tracked separately".

### Where

- `src/lnurl.rs` — `extract_lnurl` (lightning address → well-known URL) and `build_callback_url` (server-returned `callback`).
- Reached from `src/lightning/invoice.rs` → `is_valid_invoice` → `validate_lightning_address` → `ln_exists`, and from `src/app/release.rs` → `do_payment` → `resolv_ln_address`.

### Why it's reachable from user input

`is_valid_invoice` accepts a buyer-supplied `buyer_invoice` as a lightning address whenever `LightningAddress::from_str` succeeds. That parser is `email_address::EmailAddress` under the hood, which happily accepts internal hostnames and bare IPs. Verified against the versions pinned in this repo:

```
"user@169.254.169.254"  -> LightningAddress::from_str OK -> GET https://169.254.169.254/.well-known/lnurlp/user
"user@localhost"        -> LightningAddress::from_str OK -> GET https://localhost/.well-known/lnurlp/user
"user@[10.0.0.1]"       -> LightningAddress::from_str OK -> Url::parse fails (rejected today, by accident)
```

So a taker/maker can make mostrod issue an outbound GET to the cloud metadata endpoint or to any service on the daemon's private network, simply by submitting it as their payout address. There is a second hop too: the LNURL server's `callback` field is remote-controlled and gets fetched with no host check either.

### What an attacker gets

Blind SSRF, mostly: the response body is parsed as LNURL JSON and discarded unless it looks like a `payRequest`. Response *timing* and success/failure are still observable through the invoice-validation error returned to the user, which is enough for internal port/host scanning.

### Suggested fix

Resolve the host and reject non-global addresses before the request, for both hops:

- Reject loopback, private (RFC1918), link-local (169.254.0.0/16, fe80::/10), unique-local, and unspecified addresses.
[...]

PR #845, opened 2026-07-28
fix: recover from failed payout resolution, and bound outbound network latency

grunch:
Supersedes #843, which was reworked after review found that one of its changes rested on a wrong assumption. Two commits, and the order matters — the first is what makes the second safe.

## 1. A failed payout address is now recorded as a payment failure

`do_payment` resolves a lightning address before paying the buyer (`release.rs`). When that resolution produced no invoice, it returned early — skipping every piece of failure bookkeeping the rest of the payment path depends on.

Following it through:

- `find_failed_payment` (`db.rs`) selects on `failed_payment == true`. That flag was never set, so the retry job could not see the order.
- Both call sites discard the error (`let _ = do_payment(...)`, in `release.rs` and `admin_settle.rs`).
- The buyer notification lives inside `check_failure_retries`, which was never reached.
- A buyer sending a corrected invoice did not help either: that path resets `payment_attempts` and then relies on the same retry job.

Net result: the order sat in `SettledHoldInvoice` with the seller's funds captured, the buyer unpaid and never told, and no route back short of operator intervention. Related to #601 and #814.

The fix routes that failure through `check_failure_retries`, exactly as a failed `send_payment` already is, so the flag is set, the buyer is notified and the retry job takes over. An empty response is treated as a failure too, since `resolv_ln_address` reports some server-side rejections as `Ok("")` (#814).

This bug predates the branch. It's fixed first because the next commit makes the path more reachable.

## 2. Outbound resolution is bounded, and local checks run first

Handlers run sequentially on one event-loop task, so a request one handler makes delays every other message behind it. Several had no effective upper bound.

- **LNURL**: a total wall-clock budget covering every round-trip, a connect timeout, and a finite redirect chain. A per-request timeout alone wasn't sufficient — `resolv_ln_address` makes two sequential requests and the second host comes from the first response. This makes the 15s timeout in `dev_fee.rs` unreachable, so it's removed rather than left advertising a limit that is no longer the real one.
- **`LndConnector::new`**: a connect timeout, where there was none. The RPCs after it are deliberately left alone — a timeout on `settle_invoice` would report failure for work LND may have completed, which is a worse failure mode than waiting.
- **`order_action`**: local checks (fiat currency, amounts, range limits, quote) run before the payment request, which is then validated without network I/O via the new `InvoiceCheck::Offline`. Worth being precise about why that's safe, since #843 got this wrong: the address **is** persisted as `orders.buyer_invoice` and **is** what gets paid out. It's safe because commit 1 makes a resolution failure at payout time recoverable. Probing liveness at creation only buys an answer that can go stale before it's used.
- **`add_invoice_action`**: rejects on order status before resolving the payment request. An order in a status that can't accept an invoice no longer pays for the round-trip first.

`take_sell` and `add_invoice` keep resolving over the network (`InvoiceCheck::Online`), where reachability is genuinely part of the decision. Behavior there is unchanged apart from the bound.

## Test plan

- [x] `cargo test` — 1053 passed, 0 failed
- [x] `cargo fmt --all`

New coverage:

- `do_payment_records_failure_when_payout_address_yields_no_invoice` — asserts `failed_payment` and `payment_attempts` are persisted. **Verified to fail against the old implementation** before the fix was applied, so it pins the actual behavior rather than restating it.
- `add_invoice_action_rejects_disallowed_status_before_resolving_address` — points the payment request at a host that completes the handshake and never answers, then asserts the status rejection returns in well under a second.
- `lnurl_validation_against_unresponsive_host_is_time_bounded` — the same unresponsive host cannot hold validation open past the budget. Uses an ephemeral port via LNURL rather than the fixed 8080 that `start_test_server` owns, so it doesn't race the existing tests.
- `offline_validation_*` — the offline path returns immediately, and still enforces every local BOLT11 rule.
- `validate_invoice_paths` now runs its whole table under both check modes, pinning down that they agree on every BOLT11 case.

Manual check: create an order with a lightning address on a slow host and confirm it's created without the daemon pausing; take a sell with an unreachable address and confirm it's still rejected there; settle an order whose payout address no longer resolves and confirm the buyer receives `PaymentFailed` and the order is picked up by the retry job.

## Follow-ups

- #844 — per-pubkey rate limiting. This PR caps what one message can cost; that one caps how many a sender can ask for. Handlers that keep the online check don't persist state before it, so they remain replayable until that lands.
- #837 — LNURL host validation (SSRF) is untouched here and still open.
- The LNURL budget, connect timeout and redirect cap are constants. If operators run against slow endpoints, they may want them configurable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Orders now reject disallowed statuses before resolving payment requests.
  - Invoice validation is now deferred until after key order/quote checks, and uses explicit online/offline modes as appropriate.
  - Failed Lightning-address and payment attempts now reliably record payout failures and increment retry counts.
  - Lightning node connection attempts no longer wait indefinitely.

- **Reliability**
  - LNURL fetching and existence checks are now fully time-bounded (with stricter redirects and an overall budget).
  - Added offline invoice validation to avoid network delays for syntactic checks.

### DIFF
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -496,9 +496,37 @@ pub async fn do_payment(
         return Err(MostroInternalErr(ServiceError::InvoiceInvalidError));
     }
     let payment_request = if let Ok(addr) = ln_addr {
-        resolv_ln_address(&addr.to_string(), amount, None)
-            .await
-            .map_err(|_| MostroInternalErr(ServiceError::LnAddressParseError))?
+        // Resolving a lightning address is a network round-trip to a host the
+        // buyer chose. When it yields no invoice — host unreachable, too slow,
+        // or answering without a `pr` — that is a payment failure like any
+        // other and has to go through the same bookkeeping.
+        //
+        // Returning early instead would leave `failed_payment = false`, and
+        // that flag is exactly the predicate `find_failed_payment` selects on
+        // (`db.rs`). The order would sit in `SettledHoldInvoice` with the
+        // seller's funds captured, the buyer unpaid and unnotified, and
+        // invisible to the retry job — and a buyer sending a corrected
+        // invoice wouldn't help either, since that path only resets
+        // `payment_attempts` and relies on the same job to act on it.
+        match resolv_ln_address(&addr.to_string(), amount, None).await {
+            Ok(pr) if !pr.is_empty() => pr,
+            outcome => {
+                match outcome {
+                    Err(e) => info!(
+                        "Order id {}: could not resolve payout address: {:?}",
+                        order.id, e
+                    ),
+                    _ => info!("Order id {}: payout address returned no invoice", order.id),
+                }
+                if let Ok(failed_payment) = check_failure_retries(ctx, &order, request_id).await {
+                    info!(
+                        "Order id {} has {} failed payments retries",
+                        failed_payment.id, failed_payment.payment_attempts
+                    );
+                }
+                return Err(MostroInternalErr(ServiceError::LnAddressParseError));
+            }
+        }
     } else {
         payment_request
     };
@@ -1651,6 +1679,55 @@ mod tests {
             result,
             Err(MostroInternalErr(ServiceError::InvoiceInvalidError))
         ));
+    }
+
+    /// A payout address that yields no invoice must be recorded as a payment
+    /// failure, not returned as a bare error.
+    ///
+    /// The distinction is the whole point: `find_failed_payment` retries on
+    /// `failed_payment == true`, so an order that fails resolution without
+    /// setting that flag is settled, unpaid and invisible to the retry job —
+    /// and the buyer never hears about it.
+    #[tokio::test]
+    async fn do_payment_records_failure_when_payout_address_yields_no_invoice() {
+        // Arrange: an order settled and awaiting payout, whose buyer invoice
+        // is a lightning address that cannot produce an invoice. Under
+        // cfg(test) the address resolves to a local well-known path that
+        // isn't served, so this never leaves the machine.
+        init_global_config();
+        let pool = create_test_pool().await;
+        let ctx = build_ctx(&pool);
+        let seller = Keys::generate().public_key();
+        let buyer = Keys::generate().public_key();
+        let mut order = fiat_sent_sell_order(seller, buyer);
+        order.status = Status::SettledHoldInvoice.to_string();
+        order.buyer_invoice = Some("nosuchpayee@localhost".to_string());
+        order.amount = 1_000;
+        order.fee = 10;
+        order.payment_attempts = 0;
+        order.failed_payment = false;
+        let order = order.create(&pool).await.unwrap();
+
+        // Act
+        let result = do_payment(&ctx, order.clone(), None).await;
+
+        // Assert: the call still fails, but it left the bookkeeping behind
+        // that makes the failure recoverable.
+        assert!(result.is_err(), "unresolvable payout address must fail");
+        let (failed_payment, payment_attempts): (bool, i64) =
+            sqlx::query_as("SELECT failed_payment, payment_attempts FROM orders WHERE id = ?")
+                .bind(order.id)
+                .fetch_one(&pool)
+                .await
+                .unwrap();
+        assert!(
+            failed_payment,
+            "failed_payment must be set so find_failed_payment can retry the order"
+        );
+        assert_eq!(
+            payment_attempts, 1,
+            "the failed attempt must be counted, not silently dropped"
+        );
     }
 
     #[tokio::test]
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -16,7 +16,36 @@ use sqlx::{Pool, Sqlite};
 use std::cmp::Ordering;
 use std::str::FromStr;
 use tokio::sync::mpsc::channel;
-use tracing::info;
+use tracing::{error, info};
+
+/// Record a failed payout attempt so the order stays recoverable.
+///
+/// Every path that gives up on paying the buyer after the hold invoice has
+/// been settled has to come through here. `find_failed_payment` selects orders
+/// on `failed_payment == true`, so an attempt that returns without setting it
+/// leaves the order settled, the buyer unpaid and unnotified, and invisible to
+/// the retry job — and a buyer sending a corrected invoice does not help,
+/// since that path only resets `payment_attempts` and relies on that same job.
+///
+/// The bookkeeping can itself fail: the `UPDATE` may error, or an order whose
+/// amount or kind no longer validates bails out on the retries-exhausted
+/// branch. That leaves exactly the unrecoverable state this function exists to
+/// prevent, so it is logged at `error` rather than dropped — both callers of
+/// [`do_payment`] discard its `Result`, which makes the log the only signal an
+/// operator ever gets.
+async fn record_payout_failure(ctx: &AppContext, order: &Order, request_id: Option<u64>) {
+    match check_failure_retries(ctx, order, request_id).await {
+        Ok(failed_payment) => info!(
+            "Order id {} has {} failed payments retries",
+            failed_payment.id, failed_payment.payment_attempts
+        ),
+        Err(e) => error!(
+            "Order id {}: could not record the payment failure ({:?}); the order may stay \
+             settled and unpaid until it is recovered manually",
+            order.id, e
+        ),
+    }
+}
 
 /// Check if order has failed payment retries
 pub async fn check_failure_retries(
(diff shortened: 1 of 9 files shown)


grunch:
Thanks both — addressed in ac2c989.

**Record LND connection timeouts as payout failures** (codex, \`lightning/mod.rs\`): correct, and it is a hole this branch widened. \`LndConnector::new()\` propagated with \`?\` ahead of both \`check_failure_retries\` calls, and the connect deadline added here makes it reachable on a wedged LND rather than only a misconfigured one. Fixed.

**Propagate retry-bookkeeping failures** (coderabbit, \`release.rs\`): agreed that discarding the result is wrong — a failed \`UPDATE orders\` produces precisely the state this branch is guarding against. I did not take the \`?\` propagation, though, because it would not reach anyone: both callers of \`do_payment\` are \`let _ = do_payment(...)\`, so the error would be dropped one frame later. Instead both sites now go through a \`record_payout_failure\` helper that logs a bookkeeping failure at \`error\` with a note that the order may need manual recovery, which is the only signal an operator actually gets. The pre-existing \`send_payment\` failure path had the same pattern and was switched over too, so all three exits are consistent.

**Clarify \`buyer_invoice\` validation before publication** (coderabbit, \`order.rs\`): checked, and the two are the same field — no fix needed, but I expanded the comment since the ask was to document it. For a \`NewOrder\` the payload is \`Payload::Order\`, and \`Message::get_payment_request\` resolves that to \`ord.buyer_invoice\`, which is exactly what \`prepare_new_order\` copies into the stored order. A \`NewOrder\` with any other payload never reaches that code at all, because \`get_order\` returns \`None\` and the block is skipped. So there is no path where one value is validated and a different one persisted.

Both new assertions were verified to fail against the previous behavior before the fixes were applied. Full suite: 1053 passed, clippy clean.

src/app/release.rs
arkanoider, line 610: Non-blocking follow-up from the adversarial review: this async arm still uses the older silent `if let Ok(check_failure_retries)` pattern, so a bookkeeping failure is dropped. The sync paths already go through `record_payout_failure`.

```suggestion
                        PaymentStatus::Failed => {
                            info!(
                                "Order Id {}: Invoice with hash: {} has failed!",
                                order.id, msg.payment.payment_hash
                            );

                            // Same bookkeeping as the sync failure paths above:
                            // log loudly if the UPDATE fails instead of dropping it.
                            record_payout_failure(&ctx, &order, request_id).await;
                        }
```

That keeps every settled-payout failure path consistent with the helper's contract.

arkanoider (approved):
I tested manually this morning, all the specified case works correctly now. I added a small follow up an non-blocking alignment with new `record_payout_failure` function. Take a look.

tACK for me

arkanoider (commented):
## Review — Approve with comments

Deep pass over `ac2c989d` (commits `7e412251` → `480166a3` → `ac2c989d`). Recommendation: **approve and merge**, with one small consistency follow-up.

---

### Verdict

This correctly supersedes #843. Commit order matters and is right:

1. **Payout recovery first** — a failed LN address resolution (or empty `pr`) is recorded as a payment failure so the order is no longer stuck `SettledHoldInvoice`, unpaid, unnotified, and invisible to `find_failed_payment`.
2. **Then** Offline validation at `NewOrder` + LNURL/LND bounds — safe *because* of (1). #843's "discarded result ⇒ skip network" assumption was wrong for the persisted `buyer_invoice`; this PR documents and fixes that.

No blocking correctness issues for the stated goals.

---

### What works well

**Stuck-payout fix (`do_payment`)**

- Resolution errors and `Ok("")` both go through bookkeeping (covers #814-style empty responses).
- LND connect failures / timeouts also go through bookkeeping — important because the new connect deadline makes that path more reachable on a wedged LND.
- `record_payout_failure` logs bookkeeping failure at `error` instead of `?` into callers that do `let _ = do_payment(...)`. Right call.
- Regression test `do_payment_records_failure_when_payout_address_yields_no_invoice` (and the tightened LND-unreachable assertions) pin the actual bug.

**Latency / DoS mitigation**

- `InvoiceCheck::Offline` on `order_action` after local checks — removes the cheap unauthenticated NewOrder stall vector.
- `Online` retained on `take_sell` / `add_invoice` where reachability is part of the decision.
- Status check before Online resolve in `add_invoice_action` (plus the blackhole test that asserts &lt;1s rejection).
- `LNURL_TOTAL_BUDGET` across *both* hops (per-request timeout alone was insufficient for `resolv_ln_address`).
- Connect timeout + finite redirects; dead 15s wrapper removed from `dev_fee.rs`.
- LND connect bounded; settle RPCs correctly left unbounded.

**Clarity**

- Comments on Offline vs Online and on why `buyer_invoice` validated == persisted are accurate and useful for the next reader.

---

### Non-blocking: async `PaymentStatus::Failed` still silent

`ac2c989d` aimed to route every failed payout attempt through `record_payout_failure`. The sync paths do; the async arm still uses the older pattern:

```rust
PaymentStatus::Failed => {
    // ...
    if let Ok(failed_payment) =
        check_failure_retries(&ctx, &order, request_id).await
    {
        info!(/* ... */);
    }
}
```

When the UPDATE (or retries-exhausted validation) fails, the error is dropped — same hole class this PR closed elsewhere, just quieter. Bookkeeping still runs when the UPDATE succeeds, so this is **not** a merge blocker.

Suggested change:

```suggestion
                        PaymentStatus::Failed => {
                            info!(
                                "Order Id {}: Invoice with hash: {} has failed!",
                                order.id, msg.payment.payment_hash
                            );

                            // Same bookkeeping as the sync failure paths above:
                            // log loudly if the UPDATE fails instead of dropping it.
                            record_payout_failure(&ctx, &order, request_id).await;
                        }
```

(Happy to land this as a one-line follow-up if you'd rather keep this PR scoped.)

---

### Other nits (non-blocking)

| Item | Note |
|------|------|
| Early `do_payment` exits (`buyer_invoice` missing / `amount == 0`) | Still skip bookkeeping — pre-existing; unlikely for a settled order. |
| LND outage burns `payment_attempts` | Better than stuck forever; ops should know retries can exhaust during node downtime. |
| `pay_new_invoice` resets attempts, leaves `failed_payment = true` | Needed for the retry job; subtle but correct. |
| Hardcoded LNURL/LND timeouts | Fine for now; configurable later if operators hit slow LNURL hosts. |

---

### Residual surface (acknowledged follow-ups — OK)

- **#844** — per-pubkey rate limit; Online paths can still HOL-block ~5s per message.
- **#837** — LNURL host / SSRF allowlisting untouched.
- Event loop remains sequential — naive `spawn` correctly deferred.

---

### Test plan check

Covered well: LN resolve → `failed_payment`; LND unreachable → bookkeeping; Offline no I/O; BOLT11 parity Offline/Online; status-before-resolve; LNURL budget vs blackhole.

Nice-to-have later: dedicated `Ok("")` fixture; e2e Offline NewOrder → settle → unresolvable address → `PaymentFailed` + scheduler pickup; `PaymentStatus::Failed` → `record_payout_failure` once that arm is fixed.

---

### Bottom line

**Approve with comments.** Ship the recovery + bounds. Please consider the `PaymentStatus::Failed` → `record_payout_failure` consistency fix in this PR or immediately after.

arkanoider:
## Independent deep review of `ac2c989d` (local checkout, verified against code)

Follow-up to my earlier tACK: I ran a second deep pass locally on this branch, verifying every load-bearing claim in the PR description and review thread against the actual code rather than taking them at face value.

**Verdict: no blocking points.** All findings below were verified on a local checkout of `ac2c989d`.

### Verified

- **Commit 1 (payout recovery)** — All three exits out of `do_payment` (LN address resolution failure / empty `pr`, LND connect failure, `send_payment` failure) go through `record_payout_failure`, which sets exactly the flag `find_failed_payment` selects on and notifies the buyer on first failure. The `Ok("")` case (#814) is covered. Logging bookkeeping failures at `error` instead of propagating is right: both callers do `let _ = do_payment(...)`, so a propagated error would vanish one frame up.
- **Field equivalence in `order_action`** — Confirmed in mostro-core 0.14.1: `Message::get_payment_request` for `Payload::Order` reads `ord.buyer_invoice`, which is exactly what `prepare_new_order` persists. One value is validated and saved; the CodeRabbit concern is fully resolved.
- **Reorder safety in `order_action`** — `calculate_and_check_quote` takes the order immutably and `get_bitcoin_price` is a cached read, so moving `validate_invoice` after the local checks does not change the amounts validation sees.
- **Bounded resolution** — `LNURL_TOTAL_BUDGET` is what actually bounds `resolv_ln_address`'s two sequential round-trips; the per-request timeout alone couldn't. `HTTP_CLIENT` is only used inside `src/lnurl.rs`, so tightening its timeout to 4s and capping redirects affects nothing else. Removing the dead 15s wrapper in `dev_fee.rs` is correct since the inner 5s budget always fires first. Leaving the settle RPCs unbounded is the right trade-off.
- **Commit 3** — Properly addresses both the Codex P1 (LND connect timeout skipping bookkeeping) and the CodeRabbit major (discarded `check_failure_retries` errors).

### Local checks

- `cargo test`: **1053 passed, 0 failed** (1 ignored — the Cashu mint test that needs a live mint, expected)
- `cargo clippy --all-targets --all-features`: clean

### Non-blocking observations (mostly already acknowledged in this thread)

- The async `PaymentStatus::Failed` arm still uses the silent `if let Ok(check_failure_retries)` pattern — confirmed, and confirmed non-blocking: bookkeeping still runs on that path, only a failure of the bookkeeping itself is dropped silently, which is pre-existing behavior. One-line swap to `record_payout_failure`, fine here or as an immediate follow-up.
- `do_payment` early exits for missing `buyer_invoice` / `amount == 0` still skip bookkeeping — pre-existing, implausible for a settled order.
- Payout-time LN address resolution is now capped at 5s, so a consistently slow-but-alive host burns `payment_attempts` across retries. Acceptable since the buyer is notified and can submit a corrected invoice; configurable budgets are already listed as a follow-up.
- Error-code precedence in `order_action` shifts slightly (a malformed order with a bad invoice now reports the local failure first) — intentional and harmless.

Mergeable as-is.

grunch:
Closing to concentrate on the five findings in the security review. The branch `fix/payout-recovery-and-bounded-resolution` is untouched and this PR can be reopened as-is.

Recording what does *not* come back on its own, so it isn't lost:

1. **The payout bookkeeping fix is unique to this PR.** None of the five findings covers it. Today a payout attempt that gives up before `check_failure_retries` leaves the order in `SettledHoldInvoice` with `failed_payment = false` — seller's funds captured, buyer unpaid and unnotified, and invisible to `find_failed_payment`, so the retry job never picks it up. A buyer sending a corrected invoice does not rescue it either, since that path only resets `payment_attempts`. That is an unrecoverable-without-manual-intervention state and it stays in `main` while this is closed.

2. **The LNURL hardening here is partial, not a fix for the SSRF finding.** Request/connect timeouts and a 3-redirect cap bound the blast radius and close redirect-chain bypasses, but there is still no host validation — private/loopback/link-local callbacks are accepted. Reopening this does not close that finding.

3. **One design conflict to settle before re-landing.** The SSRF finding recommends mapping LNURL `status:"ERROR"` to `Err(...)` so it stops consuming `payment_attempts`. This PR deliberately goes the other way: every failed resolution goes through the bookkeeping precisely so the order stays recoverable. Those cannot both be applied as written — the recoverable-order argument is the stronger one, so the finding's suggestion needs rewording rather than the code changing.

PR #848, opened 2026-07-30
refactor: removed dead code identified in issue 

### LINKED ISSUE #818: [LOW] Dead code: unreachable hex guards, is_solver==0 recheck, SQLite rebuild fallback, swallowed setup_dispute error
### Severity: LOW (maintainability / dead code)

Several unreachable or vestigial branches surfaced while writing tests. None are bugs today, but they add noise and can mask real logic.

### Instances
- **`src/app/restore_session.rs`** — hex-validity guards on values that are already guaranteed valid upstream; the error arm is unreachable.
- **`src/app/admin_take_dispute.rs`** — an `is_solver == 0` re-check after the caller has already been established as a solver; the branch can't be hit.
- **`src/db.rs`** — a SQLite table-rebuild / version-fallback path that current schema/versions never trigger.
- **`src/app/dispute.rs`** — `setup_dispute` error is swallowed in at least one arm (worth either handling or documenting).

### Suggested fix
Remove the dead arms (or convert them to `debug_assert!`/documented invariants) so the remaining code reflects reachable behavior. Keeps coverage honest and reduces reader confusion.

Arowolokehinde:
- restore_session: drop unreachable hex-validity guards on PublicKey values; nostr_sdk::PublicKey always serializes to 64-char hex so the guards could never trigger

- admin_take_dispute: drop is_solver == 0 recheck after find_solver_pubkey; the SQL already filters WHERE is_solver == true so the Ok arm is structurally guaranteed non-zero

- db: remove rebuild_disputes_table_without_tokens and the SQLite version-fallback branch inside migrate_remove_token_columns; DROP COLUMN is supported on all deployment targets (SQLite >= 3.35 everywhere: Ubuntu 24.04 = 3.45, Debian Bookworm = 3.40, Alpine = 3.43+)

- dispute: propagate setup_dispute error with map_err(MostroCantDo)? instead of silently swallowing it; the old if .is_ok() pattern skipped order.update() on failure but kept running, leaving the order flags unset while still creating the dispute row

Closes #818 

## Summary by CodeRabbit

- **Bug Fixes**
  - Dispute actions now stop and clearly report setup failures instead of continuing with incomplete processing.
  - Solver authorization handling has been updated for more consistent dispute resolution.
  - Session restoration now proceeds through the restoration workflow without unnecessary early rejection.
  - Legacy dispute data upgrades are more reliable, preserve existing records, and remove obsolete fields safely.

### DIFF
--- a/src/app/dispute.rs
+++ b/src/app/dispute.rs
@@ -169,13 +169,14 @@ pub async fn dispute_action(
     let dispute = Dispute::new(order_id, order.status.clone());
 
     // Setup dispute
-    if order.setup_dispute(is_buyer_dispute).is_ok() {
-        order
-            .clone()
-            .update(pool)
-            .await
-            .map_err(|cause| MostroInternalErr(ServiceError::DbAccessError(cause.to_string())))?;
-    }
+    order
+        .setup_dispute(is_buyer_dispute)
+        .map_err(MostroCantDo)?;
+    order
+        .clone()
+        .update(pool)
+        .await
+        .map_err(|cause| MostroInternalErr(ServiceError::DbAccessError(cause.to_string())))?;
 
     // Save dispute to database
     let dispute = dispute
--- a/src/db.rs
+++ b/src/db.rs
@@ -79,105 +79,14 @@ pub async fn find_active_trade_pubkeys(pool: &SqlitePool) -> Result<Vec<String>,
     Ok(keys.into_iter().collect())
 }
 
-/// Helper function to rebuild disputes table without token columns when DROP COLUMN is unsupported.
-async fn rebuild_disputes_table_without_tokens(pool: &SqlitePool) -> Result<(), MostroError> {
-    tracing::info!("Rebuilding disputes table without token columns (SQLite compatibility mode)");
-
-    // Create temporary table with new schema (without token columns)
-    sqlx::query(
-        r#"
-        CREATE TABLE IF NOT EXISTS disputes_temp (
-            id char(36) primary key not null,
-            order_id char(36) unique not null,
-            status varchar(10) not null,
-            order_previous_status varchar(10) not null,
-            solver_pubkey char(64),
-            created_at integer not null,
-            taken_at integer default 0
-        )
-        "#,
-    )
-    .execute(pool)
-    .await
-    .map_err(|e| {
-        MostroInternalErr(ServiceError::DbAccessError(format!(
-            "Failed to create temporary disputes table: {}",
-            e
-        )))
-    })?;
-
-    // Copy data from original table to temporary table (excluding token columns)
-    sqlx::query(
-        r#"
-        INSERT INTO disputes_temp (id, order_id, status, order_previous_status, solver_pubkey, created_at, taken_at)
-        SELECT id, order_id, status, order_previous_status, solver_pubkey, created_at, taken_at
-        FROM disputes
-        "#,
-    )
-    .execute(pool)
-    .await
-    .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(format!(
-        "Failed to copy data to temporary table: {}", e
-    ))))?;
-
-    // Drop original table
-    sqlx::query("DROP TABLE disputes")
-        .execute(pool)
-        .await
-        .map_err(|e| {
-            MostroInternalErr(ServiceError::DbAccessError(format!(
-                "Failed to drop original disputes table: {}",
-                e
-            )))
-        })?;
-
-    // Rename temporary table to disputes
-    sqlx::query("ALTER TABLE disputes_temp RENAME TO disputes")
-        .execute(pool)
-        .await
-        .map_err(|e| {
-            MostroInternalErr(ServiceError::DbAccessError(format!(
-                "Failed to rename temporary table: {}",
-                e
-            )))
-        })?;
-
-    tracing::info!("Successfully rebuilt disputes table without token columns");
-    Ok(())
-}
-
-/// Migrates legacy disputes table by removing deprecated buyer_token and seller_token columns if present.
+/// Removes deprecated `buyer_token` and `seller_token` columns from the disputes table if present.
 ///
-/// This function uses transactions for atomic operations and includes fallback logic for older SQLite versions
-/// that don't support ALTER TABLE DROP COLUMN. The function handles both cases where columns exist (legacy databases)
-/// and don't exist (newer databases).
+/// Both drops run inside a single transaction so either both succeed or neither does.
+/// This is a no-op when neither column exists (fresh installs and already-migrated databases).
 async fn migrate_remove_token_columns(pool: &SqlitePool) -> Result<(), MostroError> {
-    // Check if token columns exist
-    let buyer_token_exists = sqlx::query_scalar::<_, i32>(
-        r#"
-        SELECT COUNT(*) 
-        FROM pragma_table_info('disputes') 
-        WHERE name = 'buyer_token'
-        "#,
-    )
-    .fetch_one(pool)
-    .await
-    .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?
-        > 0;
+    let buyer_token_exists = table_column_exists(pool, "disputes", "buyer_token").await?;
+    let seller_token_exists = table_column_exists(pool, "disputes", "seller_token").await?;
 
-    let seller_token_exists = sqlx::query_scalar::<_, i32>(
-        r#"
-        SELECT COUNT(*) 
-        FROM pragma_table_info('disputes') 
-        WHERE name = 'seller_token'
-        "#,
-    )
-    .fetch_one(pool)
-    .await
-    .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?
-        > 0;
-
-    // If no token columns exist, no migration needed
     if !buyer_token_exists && !seller_token_exists {
         tracing::debug!(
             "No deprecated token columns found in disputes table - migration not needed"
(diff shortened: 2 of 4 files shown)


ToRyVand (commented):
Reviewed the four changes against `main`. The substance is correct — both dead-code claims hold up under checking — but I found three things worth surfacing before merge.

**Verified as genuinely unreachable (both claims correct):**

- `restore_session.rs` — the hex/length guards were dead. `master_key` and `trade_key` come from `event.identity` / `event.sender`, which are already `nostr_sdk::PublicKey`; `to_string()` on those can only ever produce 64-char hex, so neither guard could fire. The `PublicKey::from_hex` calls further down in `send_restore_session_response` / `send_restore_session_timeout` still validate the string form where it actually matters, and their tests still cover the invalid-key path.
- `admin_take_dispute.rs` — the `is_solver == 0` recheck was structurally unreachable: `find_solver_pubkey` queries `WHERE pubkey == ?1 AND is_solver == true` via `fetch_one`, so an `Ok` arm necessarily carries `is_solver == true`.

**Three findings:**

**1. The branch is 32 commits behind `upstream/main`** and needs a rebase before merge. Side effect while reviewing: the suite here runs 1045 tests vs 1187 on current `main`, which initially looked like deleted test coverage and isn't — it's just the older base. Worth rebasing so a reviewer doesn't have to rule that out.

**2. The SQLite justification reaches the right conclusion from the wrong premise.** The commit message argues DROP COLUMN is safe by enumerating distro versions (Ubuntu 24.04 = 3.45, Bookworm = 3.40, Alpine 3.43+). But `mostrod` never uses the host's SQLite: `libsqlite3-sys` 0.37.0 compiles the bundled amalgamation (`cargo:rerun-if-changed=sqlite3/sqlite3.c`) and links it statically (`cargo:rustc-link-lib=static=sqlite3`), pinning **SQLite 3.51.3** at compile time. So removing the fallback is *safer* than argued — it cannot depend on the deployment environment at all. Only flagging because that reasoning will be the record for whoever revisits this: the distro-version framing implies a host dependency that doesn't exist, and would send someone re-adding a fallback for a machine with old system SQLite.

**3. The `dispute.rs` change is a behavior fix, not dead-code removal** — you do say so in the description, so this is context for @grunch rather than a correction. Tracing it: `setup_dispute` returns `Err(CantDoReason::DisputeCreationError)` only when the disputing party's flag was already set. Old code skipped `order.update()` but kept going and still created the dispute row; new code returns early. Returning early is right.

The nuance: `dispute_action` already guards at the top with `find_dispute_by_order_id(...).is_ok()` → `DisputeAlreadyExists`, so the normal double-dispute path never reaches `setup_dispute` twice. That means the `Err` branch is only reachable from an inconsistent DB state (order flag set with no dispute row), where the old code silently self-healed and the new code hard-fails with `DisputeCreationError` — leaving that user unable to open a dispute until the state is fixed. I think failing loudly on a violated precondition is the correct call, and it's a state that shouldn't arise. Just worth a maintainer knowing it's the trade-off being made rather than discovering it from a report later.

**Checks on the branch as-is:** 1045 passed / 0 failed, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean. The three `migrate_remove_token_columns` tests (no-op, both-columns, single-column) survive the rewrite, so the migration keeps its coverage.

Nothing here is a blocker on the code itself — my only actual ask is the rebase, plus optionally correcting the SQLite rationale so it doesn't mislead later. Contributor, not a maintainer, so this is a technical second opinion rather than a merge signal.

Arowolokehinde:
Thank you for the feedback @ToRyVand 
i just rebased onto upstream/main -  1209 tests passing, cargo fmt --check and cargo clippy clean.

Also corrected the SQLite rationale in the commit message. The distro-version framing was wrong — mostrod links libsqlite3-sys 0.37.0 which compiles and statically links the bundled SQLite source (3.51.3) at build time, so the fallback removal is independent of any host SQLite version. Updated the message to reflect that.

ToRyVand:
Verified independently against `69542b3`: `cargo test --all` → 1209 passed, 3 ignored; `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` both clean. Confirmed `libsqlite3-sys 0.37.0` in `Cargo.lock` statically links the bundled SQLite source, so the corrected commit message rationale checks out.

One note: rebase was onto `main`@2f2b813 — 3 commits have landed since, including #888 which also touches `src/db.rs`. Rebased locally against current `main` and it applies clean, no conflicts, just flagging in case this sits a while longer before merge.

src/db.rs
```diff
@@ -79,205 +79,63 @@ pub async fn find_active_trade_pubkeys(pool: &SqlitePool) -> Result<Vec<String>,
...
-
-/// Migrates legacy disputes table by removing deprecated buyer_token and seller_token columns if present.
+/// Removes deprecated `buyer_token` and `seller_token` columns from the disputes table if present.
 ///
-/// This function uses transactions for atomic operations and includes fallback logic for older SQLite versions
-/// that don't support ALTER TABLE DROP COLUMN. The function handles both cases where columns exist (legacy databases)
-/// and don't exist (newer databases).
+/// Both drops run inside a single transaction so either both succeed or neither does.
+/// This is a no-op when neither column exists (fresh installs and already-migrated databases).
 async fn migrate_remove_token_columns(pool: &SqlitePool) -> Result<(), MostroError> {
```
grunch, line 86: **MEDIUM** | _follow-up, not introduced by this PR_

**The drop path this PR keeps is itself unreachable in production.**

`migrations/20230928145530_disputes.sql` was edited **in place** by #516 (ab338c5) to remove the two token columns. sqlx 0.9 `Migrator::run` calls `validate_applied_migrations` (`sqlx-core-0.9.0/src/migrate/migrator.rs:255,274`) and returns `MigrateError::VersionMismatch` whenever a recorded checksum differs from the file on disk.

So any database old enough to still carry `buyer_token`/`seller_token` fails at `migrator.run(&conn)` inside `connect()` — and `connect()` only recovers from `duplicate column name` errors via `parse_duplicate_column_name` / `reconcile_existing_add_column_migration`, never from `VersionMismatch`. `migrate_remove_token_columns` is therefore never reached on exactly the databases it exists to fix.

This PR is trimming ~140 lines from a function whose one remaining non-trivial branch is dead for the same class of reason as the branches #818 asked you to delete. Not a blocker and out of scope here, but worth an issue: either delete `migrate_remove_token_columns` and its tests entirely, or extend the existing reconciliation to handle `VersionMismatch` if pre-#516 databases are genuinely meant to be supported.

src/db.rs
```diff
@@ -79,205 +79,63 @@ pub async fn find_active_trade_pubkeys(pool: &SqlitePool) -> Result<Vec<String>,
...
-                e
-            )))
-        })?;
-
-    tracing::info!("Successfully rebuilt disputes table without token columns");
-    Ok(())
-}
-
-/// Migrates legacy disputes table by removing deprecated buyer_token and seller_token columns if present.
+/// Removes deprecated `buyer_token` and `seller_token` columns from the disputes table if present.
```
grunch, line 82: **NIT** | _documentation accuracy_

"Both drops run inside a single transaction so either both succeed or neither does" is accurate for the two `ALTER TABLE` statements, but the `table_column_exists` calls that *decide* what to drop run on a separate pooled connection, before `pool.begin()`.

Harmless in practice — single caller, at startup, before the daemon serves anything — and unchanged from the previous implementation. Just worth not implying the whole read-decide-write sequence is atomic, since a future reader may rely on that claim.

src/app/dispute.rs
```diff
@@ -169,13 +169,14 @@ pub async fn dispute_action(
...
 
     // Setup dispute
-    if order.setup_dispute(is_buyer_dispute).is_ok() {
-        order
-            .clone()
-            .update(pool)
-            .await
-            .map_err(|cause| MostroInternalErr(ServiceError::DbAccessError(cause.to_string())))?;
-    }
+    order
```
grunch, line 172: **LOW** | _test coverage_

**The new error path has no test.** `grep -rn DisputeCreationError src/` returns nothing.

This change turns a silently swallowed error into a user-visible `MostroCantDo(CantDoReason::DisputeCreationError)` response — a behavior change, as your description says — and nothing exercises it. The existing `dispute_action_*` tests cover `NotFound`, already-exists, non-disputable status, missing seller/buyer pubkey, non-party sender, and both happy paths; this new arm is the one gap.

It's reachable in a test by inserting an order with `buyer_dispute = 1` and **no** row in `disputes`, then asserting `dispute_action` returns `MostroCantDo(CantDoReason::DisputeCreationError)` and that no dispute row was created — which also pins the intended "fail loudly on inconsistent state" semantics that @ToRyVand flagged as the trade-off being made here.

src/app/dispute.rs
```diff
@@ -169,13 +169,14 @@ pub async fn dispute_action(
...
+        .setup_dispute(is_buyer_dispute)
+        .map_err(MostroCantDo)?;
```
grunch, line 174: **NIT** | _note for future readers_

In mostro-core 0.14.5 (`order.rs:537`) `setup_dispute` assigns `self.status = Status::Dispute.to_string()` **before** the `DisputeCreationError` return, so on `Err` the local `order` is left dirty.

The new code returns immediately and never persists it, so this is strictly safer than the old `if .is_ok()` form. But anyone who later moves that `?` or reuses `order` after this point inherits a trap. A one-line comment would make the invariant explicit:

```rust
// setup_dispute leaves order.status dirty on Err; we return before any persist.
order.setup_dispute(is_buyer_dispute).map_err(MostroCantDo)?;
```

grunch (approved):
Verified locally against current `main` (rebased onto `d3f4804`): merges clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, **1224 passed / 0 failed / 2 ignored**.

All four items in #818 are addressed. I re-derived the two dead-code claims independently rather than trusting the description, and I'm not repeating @ToRyVand's three points — they're resolved or already on the record.

### Independent confirmations

- **`restore_session.rs`** — `event.identity` and `event.sender` are `PublicKey` (non-`Option`) in mostro-core 0.14.5 `nip59.rs:71/75`. `to_string()` can only yield 64-char hex, so both guards were unreachable. ✅
- **`admin_take_dispute.rs`** — `find_solver_pubkey` (`db.rs:1404`) filters `WHERE pubkey == ?1 AND is_solver == true`, and `users.is_solver` is `integer not null default 0`, written only as 0/1. The `Ok` arm is structurally `is_solver == 1`. ✅ The test doc-comment at `admin_take_dispute.rs:522` already states the invariant, so coverage stays honest.
- **`db.rs`** — the legacy columns were plain `integer not null` (41f182c), and there is no index, trigger, or view on `disputes` anywhere in `migrations/`. None of SQLite's remaining `DROP COLUMN` restrictions (PK, UNIQUE, indexed, partial index, CHECK, FK, generated column, view, trigger) apply, so removing the fallback is safe on the merits. ✅

### Findings

Four inline comments below. Only the missing test for the new `dispute.rs` error path is worth acting on inside this PR; the `migrate_remove_token_columns` reachability issue is pre-existing and belongs in its own issue.

### Stale prior comment — no action needed

CodeRabbit's remaining nitpick asks for a positive-path migration test. It already exists: `migrate_remove_token_columns_drops_legacy_columns_and_keeps_rows` (`src/db.rs:4794`) recreates both legacy columns, inserts a dispute row, runs the migration, and asserts both columns are gone while the row survives. `migrate_remove_token_columns_handles_single_legacy_column` (`src/db.rs:4822`) covers the asymmetric case. Safe to resolve.

PR #849, opened 2026-07-30
test: mutation testing for Nostr event handling

### LINKED ISSUE #636: test: mutation testing for Nostr event handling
## Context

Follow-up from #618. Implement mutation testing for Nostr event handling — the communication layer between Mostro and its clients.

## Scope

- `src/nip59.rs` or NIP-59 gift wrap handling
- `src/nip33.rs` or parameterized replaceable events
- Event validation, signature verification
- Message routing and action dispatching

## Goal

- Run mutation testing on Nostr modules
- Kill surviving mutants
- Target: **>70% mutation score**

## Priority Mutants to Kill

- Event kind validation
- Signature verification logic
- Gift wrap encryption/decryption
- Message deserialization and action routing
- Pubkey validation

## Acceptance Criteria

- [ ] Baseline mutation report for Nostr modules
- [ ] Critical mutants in event validation killed
- [ ] Critical mutants in action routing killed
- [ ] Mutation score documented in PR

Ref: #618

ToRyVand:
## Summary

Closes #636. Mutation testing for the Nostr event-handling path: `accept_event`
and its gates, `create_event`'s NIP-40 expiration logic, and the
restore-session timeout.

The two functions worth extracting came out of it — `is_stale` and
`missing_inner_signature` are now pure predicates with boundary tests, instead
of conditions reachable only through a live wrapped event.

## Mutation score

Measured on this branch (base `afc39a2`, v0.18.7) with `cargo-mutants 27.1.0`,
via `make mutation-test` — mutant workers serial, test threads parallel (see
[CI and tooling](#ci-and-tooling)).

| Scope | Mutants | Caught | Missed | Unviable |
|---|---|---|---|---|
| `accept_event` (6) + `create_event` (2) + 2 `check_trade_index` struct-field mutants¹ | 10 | 10 | 0 | 0 |
| `is_stale` + `missing_inner_signature` | 9 | 9 | 0 | 0 |
| `restore_session.rs` | 6 | 3 | **3** | 0 |
| **Total** | **25** | **22** | **3** | 0 |

**88% killed**, against #636's >70% target. Reproduce with:

```
make mutation-test ARGS="--file src/app.rs --file src/nip33.rs -F 'in accept_event' -F 'in create_event' -F is_stale -F missing_inner_signature"
make mutation-test ARGS="--file src/app/restore_session.rs"
```

The two predicates are in the count on purpose: they hold the stale check and
the inner-signature check that were inline in `accept_event` before this PR, so
leaving their 9 mutants out would understate the scope.

¹ `delete field pubkey` / `delete field last_trade_index from struct User
expression in check_trade_index` come through the `-F 'in accept_event'` filter
in cargo-mutants 27.1.0. They are counted where the run puts them rather than
dropped.

### The survivors, named

All three are in `restore_session.rs`, and none of them is a coverage gap I can
close honestly.

**`12:46: replace * with +` and `12:46: replace * with /`** — the
`RESTORE_SESSION_TIMEOUT_SECS = 60 * 60` arithmetic. Killing these requires a
test that restates the constant on the line below it, which raises the score
without protecting anything.

An earlier revision of this branch did remove them, by writing the constant
`3_600` so there was no `*` left to mutate. That was the wrong trade and
@Catrya was right to call it: it improves the score by *deleting* a mutant
rather than killing one — the same inflation the rest of this PR exists to
remove — and it shapes production source around the tool measuring it. The
constant is back to `60 * 60`, matching `ORDER_TS_MAX_AGE_SECS` in
`src/util.rs`, and the mutants are reported instead of engineered away.

**`21:5: replace restore_session_action -> Result<(), MostroError> with Ok(())`**
— this one is more interesting, and mutation testing is what surfaced it.

`restore_session_action`'s only fallible call is
`manager.start_restore_session(pool, master_key).await?`. That function
**always returns `Ok(())`**: it spawns a blocking task and the DB error is
handled inside the closure — the `Err` arm of its `spawn_blocking` body in
`src/db.rs` logs it and never propagates it. So
the `?` in `restore_session_action` cannot fire, and the only difference the
mutant makes is skipping the `tokio::spawn`, which is not observable through
the return value.

Killing it would mean asserting on the background task's side effects, which is
timing-dependent and flaky. The survivor is accurate: it is telling us the
error path is already dead, not that the tests are thin.

### Tests

- **`src/app.rs`** — what this PR adds:
  - the two first-contact PoW lanes: an unknown sender is accepted when it
    clears `pow_first_contact` and dropped when it does not;
  - the accept/reject basics: a validly wrapped event, the wrong kind, the
    wrong receiver;
  - boundary tests for the two extracted predicates, `is_stale` (3) and
    `missing_inner_signature` (3).

  All of them sit in `accept_event_ordering_tests` on `create_migrated_ctx()`.
  The tampered-signature, replay, and invalid-signature tests in that module
  (`tampered_copy_does_not_censor_the_genuine_event`,
  `genuine_duplicate_is_still_dropped_as_a_replay`,
  `invalid_signature_is_dropped_without_a_gate`,
  `v1_gift_wrap_with_invalid_signature_is_dropped`) came with #892 and are not
  this PR's.

- **`src/nip33.rs`** — `create_event`'s expiration-tag dedup: a caller-supplied
  expiration must not be duplicated by the auto-expiration logic, whichever way
  the tag was built. The tests pin the exact `Tag::custom` shape
  `order_to_tags` emits.

- **`src/lnurl.rs`, `src/lightning/invoice.rs`** — `MOSTRO_TEST_LN_PORT` threaded
  through the test HTTP server and URL builder, so mutation runs don't collide
  with something already bound to 8080. The three copies of the port lookup are
  collapsed into one `test_ln_port()`, which also rejects `0`.

### Production behaviour

- **`src/app/restore_session.rs`** — the only behaviour change in this PR, in
  its own commit. The one-hour timeout was a bare `60 * 60` and the log
  reporting it was the independent literal `"1 hour"`; change one and the other
  silently lies. `RESTORE_SESSION_TIMEOUT_SECS` is now the single source, and
  the log prints the `Duration` actually handed to `tokio::time::timeout` — so
  the message cannot disagree with the timeout for any value of the constant.

- **`src/app.rs`** — `is_stale` and `missing_inner_signature` are extracted from
  `accept_event`. Their bodies are the expressions that were inline, so this is
  a refactor, not a behaviour change.

### CI and tooling

- **`Makefile`** — a `mutation-test` target, serial at the **worker** level
  only:

  - `CARGO_MUTANTS_JOBS=1` runs one mutant at a time (already the default in
    27.1.0; set so the guarantee is stated rather than inherited). This is the
    level that matters: `test_lnurl_validation_with_test_server` binds a fixed
    host port, so two concurrent runs collide on it, and a test that fails
    because it lost that race is scored as a **killed mutant** — inflating the
    number the target exists to measure.
  - Test threads inside each run stay parallel. That test is the only one that
    binds a fixed port, and every lookup it makes goes to its own listener;
    every other listener in the suite binds `:0`, and every test has a private
    in-memory pool. Serialising the threads would protect nothing and roughly
    double the time per mutant (the test run inside each one gets ~3× slower).
  - `MOSTRO_TEST_LN_PORT=18080` steps aside from a port a developer machine is
    likely to have in use.

  `.mutants.toml` is never read by cargo-mutants (#958), so this target is the
  only place the run's environment is actually defined.

- **`.github/workflows/mutation.yml`** — the PR job builds its `--file` args as
  a bash array rather than word-splitting PR-diff filenames, and sets the same
  two variables as the Makefile.

## Acceptance criteria (#636)

- [x] Baseline mutation report for Nostr modules
- [x] Critical mutants in event validation killed (`accept_event`'s
      first-contact PoW lane, and the extracted stale/signature predicates)
- [x] Critical mutants in NIP-33 replaceable-event logic killed (`create_event`)
- [x] Mutation score documented in PR — above, survivors included

## Test plan

Based on `main` @ `afc39a2` (v0.18.7).

- `cargo fmt --check` — clean
- `cargo test --bin mostrod` — **1336 passed**, 0 failed, 2 ignored

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -60,15 +60,27 @@ jobs:
             exit 0
           fi
 
-          # Build --file flags for each changed file
-          file_args=""
-          for f in $changed_rs; do
-            file_args="$file_args --file $f"
-          done
-
           echo "Running mutation testing for changed files:"
           echo "$changed_rs"
-          cargo mutants $file_args
+
+          # These filenames come from a PR diff, so they are
+          # attacker-controlled. Built as a bash array (never joined into a
+          # string) and passed with "${file_args[@]}" so a crafted filename
+          # can only ever be one --file value, never extra argv tokens — an
+          # unquoted string of them is word-split by the shell, which turns
+          # a crafted filename into extra flags for cargo-mutants/cargo/rustc.
+          # This deliberately bypasses `make mutation-test`, whose $(ARGS) is
+          # a plain string splice safe only for hand-typed input, so it
+          # repeats that target's environment here (see the Makefile for why
+          # each variable is set).
+          file_args=()
+          while IFS= read -r f; do
+            file_args+=(--file "$f")
+          done <<< "$changed_rs"
+
+          # Keep these two in sync with the mutation-test target in Makefile.
+          CARGO_MUTANTS_JOBS=1 MOSTRO_TEST_LN_PORT=18080 \
+            cargo mutants "${file_args[@]}"
 
       - name: Upload mutation report
         uses: actions/upload-artifact@v4
@@ -195,7 +207,7 @@ jobs:
           tool: cargo-mutants
 
       - name: Run full mutation testing
-        run: cargo mutants
+        run: make mutation-test
         # Note: Do NOT fail on low score initially (report only mode)
         continue-on-error: true
 
--- a/Makefile
+++ b/Makefile
@@ -66,3 +66,42 @@ docker-build-startos:
 	cd docker && \
 	docker compose build mostro-startos
 
+# The run is serial at the worker level only. One test binds a fixed host
+# port — `test_lnurl_validation_with_test_server` in src/lightning/invoice.rs
+# — and every lookup it makes, LNURL and lightning address alike, goes to its
+# own listener; every other listener in the suite binds :0. So two test
+# threads in one run cannot collide on that port, but two concurrent runs
+# can, and a test that fails because it lost that race is scored as a killed
+# mutant, silently inflating the number this target measures.
+#
+# CARGO_MUTANTS_JOBS=1 serialises the mutant workers. cargo-mutants already
+# does this by default (verified against 27.1.0: one scratch dir on a
+# 16-CPU host with no -j); it is set explicitly so the guarantee is stated
+# rather than inherited.
+#
+# Test threads inside each run stay parallel on purpose. Serialising them
+# (RUST_TEST_THREADS=1, or `--test-threads=1`) protects nothing, and
+# roughly doubles the time per mutant — enough to push the PR job for a few
+# changed files past GitHub's 6h job limit. If a test ever flakes under
+# parallel threads, name it here as the reason before serialising.
+#
+# MOSTRO_TEST_LN_PORT moves that test off 8080 in case something on the
+# host already holds it. It does NOT make the suite hermetic: two concurrent
+# runs would still collide on 18080. Serialising is what makes the run
+# trustworthy; the port override only dodges a pre-existing listener.
+#
+# The 18080 here deliberately differs from the code's own 8080 default, so
+# `cargo test` and `make mutation-test` do bind different ports. That is the
+# point: plain `cargo test` keeps exercising the default path, and only this
+# target — which runs the suite hundreds of times over — steps aside from a
+# port a developer machine is likely to have in use.
+#
+# ARGS is spliced into the shell command as plain text — only pass
+# hand-typed, trusted values (e.g. `make mutation-test ARGS="--file
+# src/foo.rs"`). Never build ARGS from PR-diff filenames or other
+# attacker-controlled input; that class of data must be turned into a bash
+# array and passed to `cargo mutants` directly (see the PR job in
+# .github/workflows/mutation.yml).
+mutation-test:
+	CARGO_MUTANTS_JOBS=1 \
+	MOSTRO_TEST_LN_PORT=$${MOSTRO_TEST_LN_PORT:-18080} cargo mutants $(ARGS)
--- a/src/app.rs
+++ b/src/app.rs
@@ -296,6 +296,24 @@ async fn handle_message_action(
     }
 }
 
+/// True when a rumor's `created_at` falls outside the 10s replay window.
+/// Pulled out of `accept_event` so the boundary can be hit directly instead
+/// of only through a live wrapped event.
+fn is_stale(created_at: Timestamp, since_time: u64) -> bool {
+    created_at.as_secs() < since_time
+}
+
+/// True when the inner rumor is missing a required signature. Full-privacy
+/// clients reuse the trade key as identity and send unsigned rumors; any
+/// other identity/sender split must carry a valid inner signature.
+fn missing_inner_signature(
+    identity: PublicKey,
+    sender: PublicKey,
+    signature: Option<Signature>,
+) -> bool {
+    identity != sender && signature.is_none()
+}
+
 /// Decode and fully validate one relay event into a dispatchable
 /// `(action, message, unwrapped)` triple, or `None` if it must be skipped
 /// (failed PoW, wrong kind, invalid event signature, spam-gate drop, decrypt
@@ -406,7 +424,7 @@ async fn accept_event(
         .checked_sub_signed(chrono::Duration::seconds(10))
         .unwrap()
         .timestamp() as u64;
-    if unwrapped.created_at.as_secs() < since_time {
+    if is_stale(unwrapped.created_at, since_time) {
         return None;
     }
     let message = unwrapped.message.clone();
@@ -415,7 +433,7 @@ async fn accept_event(
     // unsigned rumors. Any other shape must carry a valid inner
     // signature — unwrap_message already verified it, so if identity
     // and sender differ here without a signature we bail out.
-    if unwrapped.identity != unwrapped.sender && unwrapped.signature.is_none() {
+    if missing_inner_signature(unwrapped.identity, unwrapped.sender, unwrapped.signature) {
         tracing::warn!(
             "Missing inner signature: identity {} differs from trade key {}",
             unwrapped.identity,
@@ -703,6 +721,46 @@ mod tests {
     }
 
     #[test]
+    fn is_stale_rejects_events_older_than_the_cutoff() {
+        assert!(is_stale(Timestamp::from(99), 100));
+    }
+
+    #[test]
+    fn is_stale_accepts_events_exactly_at_the_cutoff() {
+        assert!(!is_stale(Timestamp::from(100), 100));
+    }
+
+    #[test]
+    fn is_stale_accepts_events_newer_than_the_cutoff() {
+        assert!(!is_stale(Timestamp::from(101), 100));
+    }
+
+    #[test]
+    fn missing_inner_signature_true_when_identity_and_sender_differ_unsigned() {
+        let identity = create_test_keys().public_key();
+        let sender = create_test_keys().public_key();
+        assert!(missing_inner_signature(identity, sender, None));
+    }
+
+    #[test]
+    fn missing_inner_signature_false_when_identity_equals_sender_unsigned() {
+        let same = create_test_keys().public_key();
+        assert!(!missing_inner_signature(same, same, None));
+    }
+
+    #[test]
+    fn missing_inner_signature_false_when_signed_even_if_identity_differs() {
+        let identity = create_test_keys().public_key();
+        let sender_keys = create_test_keys();
+        let sig = sender_keys.sign_schnorr([7u8; 32]);
+        assert!(!missing_inner_signature(
+            identity,
+            sender_keys.public_key(),
+            Some(sig)
+        ));
+    }
+
+    #[test]
     fn test_warning_msg_all_error_types() {
         let action = Action::NewOrder;
 
@@ -778,7 +836,7 @@ mod tests {
         use super::*;
         use crate::spam_gate::{SpamGate, REPLAY_WINDOW_SECS};
         use mostro_core::nip59::{wrap_message, WrapOptions};
-        use mostro_core::transport::wrap_message_nip44;
+        use mostro_core::transport::{wrap_message_nip44, wrap_message_with, Transport};
 
         /// A protocol-v2 (kind 14) event addressed to `mostro`, in full-privacy
         /// mode (trade key doubles as identity, so no identity proof is needed).
@@ -926,6 +984,48 @@ mod tests {
             );
         }
 
+        /// The first-contact lane: an unseen sender that clears
+        /// `pow_first_contact` is let through to the decrypt.
+        #[tokio::test]
+        async fn unknown_first_contact_sender_clearing_the_pow_bar_is_accepted() {
+            let ctx = create_migrated_ctx().await;
+            let gate = SpamGate::new(REPLAY_WINDOW_SECS);
+            let mostro = create_test_keys();
+
+            // `accept` pins pow_first_contact = 0, trivially cleared, and a
+            // freshly generated trade key is never known to the gate.
+            assert!(
+                accept(&ctx, &v2_event(&mostro), &mostro, &gate)
+                    .await
+                    .is_some(),
+                "an unknown sender over the PoW bar must reach the decrypt"
+            );
+        }
+
+        /// The same lane refusing: only `pow_first_contact` differs, so the
+        /// toll is what decides, not the event or the gate.
+        #[tokio::test]
+        async fn unknown_first_contact_sender_below_the_pow_bar_is_dropped() {
+            let ctx = create_migrated_ctx().await;
+            let gate = SpamGate::new(REPLAY_WINDOW_SECS);
+            let mostro = create_test_keys();
+
+            // u8::MAX demands more leading-zero bits than any real event id
+            // will ever have, so this fails deterministically.
+            let result = accept_event(
+                &ctx,
+                &v2_event(&mostro),
+                &mostro,
+                0,
+                u8::MAX,
+                NostrKind::from(crate::config::constants::DM_EVENT_KIND),
+                Some(&gate),
+            )
+            .await;
+
+            assert!(result.is_none());
+        }
+
         /// The v2-only policy lives in one place now; both event loops read it
         /// from here, so this is where it gets covered.
         #[test]
@@ -934,6 +1034,97 @@ mod tests {
             // `SpamGate::global()` is `None` unless `install_spam_gate` ran, so
             // the v2 arm can only be pinned against the installed state.
             assert_eq!(gate_for(true).is_some(), SpamGate::global().is_some());
+        }
+
+        #[allow(deprecated)] // Transport::GiftWrap: still the tested default; see #786.
+        async fn wrap_test_message(
+            identity_keys: &Keys,
+            trade_keys: &Keys,
+            receiver: PublicKey,
+        ) -> Event {
+            // RestoreSession is the one action whose payload is required to be
+            // None (mostro-core's `verify()`), which keeps this helper free of a
+            // hand-built `Payload::Order`. The trade-index path it short-circuits
+            // is covered directly by `check_trade_index_tests`.
+            wrap_message_with(
+                Transport::GiftWrap,
+                &create_test_message(Action::RestoreSession, None),
+                identity_keys,
+                trade_keys,
+                receiver,
+                WrapOptions::default(),
+            )
+            .await
+            .expect("gift wrap succeeds")
+        }
+
+        #[tokio::test]
+        async fn accepts_a_validly_wrapped_event_and_returns_the_action() {
+            let ctx = create_migrated_ctx().await;
+            let identity_keys = create_test_keys();
+            let trade_keys = create_test_keys();
+            let receiver_keys = create_test_keys();
+            let event =
+                wrap_test_message(&identity_keys, &trade_keys, receiver_keys.public_key()).await;
+
+            let result = accept_event(
+                &ctx,
+                &event,
+                &receiver_keys,
+                0,
+                0,
+                NostrKind::GiftWrap,
+                None,
+            )
+            .await;
+
+            let (action, _message, unwrapped) = result.expect("valid event should be accepted");
+            assert_eq!(action, Action::RestoreSession);
+            assert_eq!(unwrapped.sender, trade_keys.public_key());
+            assert_eq!(unwrapped.identity, identity_keys.public_key());
+        }
+
+        #[tokio::test]
+        async fn rejects_an_event_of_the_wrong_kind() {
+            let ctx = create_migrated_ctx().await;
+            let identity_keys = create_test_keys();
+            let trade_keys = create_test_keys();
+            let receiver_keys = create_test_keys();
+            let event =
+                wrap_test_message(&identity_keys, &trade_keys, receiver_keys.public_key()).await;
+
+            // The event is a real, validly-signed GiftWrap; accepted_kind is
+            // deliberately mismatched to exercise the kind-gate on its own.
+            let result = accept_event(
+                &ctx,
+                &event,
+                &receiver_keys,
+                0,
+                0,
+                NostrKind::TextNote,
+                None,
+            )
+            .await;
+
+            assert!(result.is_none());
+        }
+
+        #[tokio::test]
+        async fn rejects_an_event_not_addressed_to_the_receiver() {
+            let ctx = create_migrated_ctx().await;
+            let identity_keys = create_test_keys();
+            let trade_keys = create_test_keys();
+            let intended_receiver = create_test_keys();
+            let eavesdropper = create_test_keys();
+            let event =
+                wrap_test_message(&identity_keys, &trade_keys, intended_receiver.public_key())
+                    .await;
+
+            // unwrap_incoming can't decrypt a rumor sealed for someone else.
+            let result =
+                accept_event(&ctx, &event, &eavesdropper, 0, 0, NostrKind::GiftWrap, None).await;
+
+            assert!(result.is_none());
         }
     }
 
--- a/src/app/restore_session.rs
+++ b/src/app/restore_session.rs
@@ -3,6 +3,14 @@ use crate::{db::RestoreSessionManager, util::enqueue_restore_session_msg};
 use mostro_core::prelude::*;
 use nostr_sdk::prelude::*;
 
+/// How long a restore-session request waits for results before the requester
+/// is told to retry instead of hanging forever.
+///
+/// Named so the timeout and the message reporting it cannot drift apart: they
+/// were two independent literals, and the log said "1 hour" whatever the
+/// duration actually was.
+const RESTORE_SESSION_TIMEOUT_SECS: u64 = 60 * 60;
+
 /// Handle restore session action
 /// This function starts a background task to process the restore session
 /// and immediately returns, avoiding blocking the main application
@@ -41,7 +49,7 @@ pub async fn restore_session_action(
 /// Handle restore session results in the background
 async fn handle_restore_session_results(mut manager: RestoreSessionManager, trade_key: String) {
     // Wait for the result with a timeout
-    let timeout = tokio::time::Duration::from_secs(60 * 60); // 1 hour timeout
+    let timeout = tokio::time::Duration::from_secs(RESTORE_SESSION_TIMEOUT_SECS);
 
     match tokio::time::timeout(timeout, manager.wait_for_result()).await {
         Ok(Some(result)) => {
@@ -60,7 +68,10 @@ async fn handle_restore_session_results(mut manager: RestoreSessionManager, trad
             tracing::error!("Restore session result channel closed unexpectedly");
         }
         Err(_) => {
-            tracing::error!("Restore session timed out after 1 hour");
+            // The `Duration` itself, not a hand-converted unit: it is the value
+            // actually passed to `tokio::time::timeout` above, so the message
+            // cannot disagree with the timeout for any value of the constant.
+            tracing::error!("Restore session timed out after {timeout:?}");
             // Send timeout message to user
             if let Err(e) = send_restore_session_timeout(&trade_key).await {
                 tracing::error!("Failed to send timeout message: {}", e);
--- a/src/lightning/invoice.rs
+++ b/src/lightning/invoice.rs
@@ -323,7 +323,10 @@ mod tests {
             )
             .layer(tower_http::cors::CorsLayer::permissive());
 
-        let listener = TcpListener::bind("127.0.0.1:8080").await.unwrap();
+        // Same source of truth as the code that builds the URL.
+        let listener = TcpListener::bind(("127.0.0.1", crate::lnurl::test_ln_port()))
+            .await
+            .unwrap();
         let addr = listener.local_addr().unwrap();
         let port = addr.port();
 
--- a/src/lnurl.rs
+++ b/src/lnurl.rs
@@ -239,6 +239,29 @@ async fn lnurl_get(url: Url) -> Result<reqwest::Response, MostroError> {
         .map_err(|_| MostroInternalErr(ServiceError::NoAPIResponse))
 }
 
+/// Host port the local LNURL test server listens on, and that `cfg!(test)`
+/// lightning addresses are resolved against.
+///
+/// Overridable via `MOSTRO_TEST_LN_PORT` so a run can dodge a port already
+/// taken on the host (`make mutation-test` sets 18080) without disturbing
+/// the 8080 default. One definition on purpose: the prod path, its own
+/// test, and the test server in `lightning::invoice` must agree, and three
+/// hand-copied parses would eventually not.
+///
+/// Not gated on `#[cfg(test)]`: the caller sits behind `cfg!(test)`, which
+/// is a runtime boolean, so both branches are compiled in every profile.
+///
+/// `0` is rejected along with unset and unparseable values: it parses as a
+/// valid `u16` but means "let the OS pick" to a listener, which would bind
+/// an arbitrary port while the URL built here still said `:0`.
+pub(crate) fn test_ln_port() -> u16 {
+    std::env::var("MOSTRO_TEST_LN_PORT")
+        .ok()
+        .and_then(|v| v.parse::<u16>().ok())
+        .filter(|p| *p != 0)
+        .unwrap_or(8080)
+}
+
 /// Parse a Lightning Address or bech32 LNURL into an http(s) [`Url`].
 ///
 /// - Lightning Address (`user@domain`) → well-known LNURL-pay endpoint URL
@@ -263,7 +286,7 @@ async fn extract_lnurl(address: &str) -> Result<Url, MostroError> {
             None => return Err(MostroInternalErr(ServiceError::LnAddressParseError)),
         };
         let base_url = if cfg!(test) {
-            format!("http://{domain}:8080")
+            format!("http://{domain}:{}", test_ln_port())
         } else {
             format!("https://{domain}")
         };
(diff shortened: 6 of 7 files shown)


ToRyVand:
## Both review comments addressed — one of them turned out to be the opposite of what it looked like

Pushed `3e05ed2` and `e885508`, kept separate since they are unrelated findings.

### `src/nip33.rs` — the `Custom("expiration")` arm can't be tested, because it can't be reached

The suggestion was to add a companion test for the `TagKind::Custom("expiration")` arm. I wrote that test first and it failed on its own first assertion, which sent me to check the premise:

```rust
let t = Tag::custom(TagKind::Custom(Cow::Borrowed("expiration")), vec!["123456".to_string()]);
// kind = Expiration | is_expiration = true
```

nostr normalises the tag name **at construction**, so `t.kind()` can never return `Custom("expiration")`. The second arm of `has_expiration_tag` was unreachable — and that also makes it an equivalent mutant: deleting it changes nothing observable, so no test could ever have killed it. Adding a test there would have been writing an assertion that passes for the wrong reason.

Worth correcting one thing I had assumed too: this arm is *not* the production-load-bearing one. `order_to_tags` does build the tag as `Tag::custom(TagKind::Custom("expiration"), ..)`, but by the time it reaches `create_event` it has already normalised to `TagKind::Expiration` — so production has always taken the first arm.

So I deleted the dead arm instead, and added a test that pins the real path end to end: the exact `Tag::custom` shape `order_to_tags` emits must normalise and suppress the auto-add, asserting the caller's `123456` is the *only* expiration tag on the event. Verified it actually bites by forcing `has_expiration_tag = false` — it goes red with `left: ["1788381606", "123456"]`, the config auto-add stacked on top of the caller's tag.

**Flagging this one as your call, since it is a production change inside a `test:` PR.** I think it belongs here — an equivalent mutant is precisely the kind of thing a mutation-testing pass should surface, and leaving it in means leaving a mutant nobody can ever kill. But the smaller option is equally defensible: restore the arm with a comment saying it is unreachable, and keep this PR's diff purely additive. Say the word and I'll do that instead.

Either way the deletion is safe across every caller: `create_event` has five entry points, all passing tags built internally, and the only two that carry an expiration are `order_to_tags` (`Tag::custom`, normalises) and `price/manager.rs:528` (`Tag::expiration`, already the first arm).

### `Makefile` — two of the three points hold

**`set -o pipefail`: no change.** [`Makefile:1`](../blob/main/Makefile#L1) is `SHELL := $(shell which bash)`, which applies to every recipe, and six existing targets already rely on it. Verified the recipe runs under bash.

**Caller's port was being overwritten: fixed.** Now `MOSTRO_TEST_LN_PORT=$${MOSTRO_TEST_LN_PORT:-18080}`, so a busy 18080 can be worked around without editing the Makefile.

**Parallel workers sharing the port: fixed, and this was the one that mattered.** Each worker runs the full suite in its own temp dir but shares the host's TCP ports, so two workers collide on the LNURL listener. Under mutation testing that collision is not just noise — a test failing for its own reasons **counts as a killed mutant**, so it silently inflates the very score this target exists to measure. Dropped to `CARGO_MUTANTS_JOBS=1`. Slower, but the number means something.

That failure mode is easy to see locally: a plain `cargo test` on this machine gives 1058 passed / 1 failed with `AddrInUse` on 8080, and `MOSTRO_TEST_LN_PORT=18080 cargo test` gives 1059 passed / 0 failed.

### On the mutation score

Since this touches `create_event`, I re-ran the measurement on the new head rather than carry the old number over:

```
$ make mutation-test ARGS="--file src/app.rs --file src/nip33.rs -F 'in accept_event|in create_event'"
Found 9 mutants to test
ok       Unmutated baseline in 140s build + 41s test
9 mutants tested in 12m: 9 caught
```

0 missed, 0 timeout, 0 unviable.

**It reads 9/9 now, not 10/10, and the PR body should say so.** Removing the unreachable arm removes its `||` — and with it the `replace || with &&` mutant — from the set. Nothing became uncovered.

While tracing that: 2 of the original 10 were `delete field .. in check_trade_index`, which land in the run because `cargo-mutants 27.1.0`'s `-F` doesn't filter "delete field from struct expression" mutants. They are caught, so they never distorted the pass/fail, but the headline was really 8 targeted + 2 incidental. It is 7 + 2 now.

PR body updated to match — score line, the `mutation-test` invocation it quoted, and the `cargo test` count.

### Verification

- `cargo test`: 1059 passed, 2 ignored (with `MOSTRO_TEST_LN_PORT` set past the host's busy 8080)
- `cargo fmt --check`: clean
- `cargo clippy --all-targets --all-features -- -D warnings`: clean

ToRyVand:
## Rebased onto `main` — and the nostr 0.45 bump turned this PR's own guard test into the thing that verified it

This had gone `CONFLICTING`. It's now rebased onto current `main` (`541129f`), no merge commit, same six commits.

The conflict itself was three small hunks, but resolving them surfaced the real issue: this branch was on `nostr-sdk 0.44.1` and `main` is now on `nostr` + `nostr-sdk 0.45.1`. That's a breaking change, so the rebase is also a small 0.44 → 0.45 migration.

### What the migration touched

**`TagKind` is gone from the 0.45 API.** The branch had `matches!(t.kind(), TagKind::Expiration)`; `main` has `t.kind() == "expiration"`. I kept `main`'s form — not as a tiebreak, but because `TagKind` genuinely doesn't exist in 0.45 (it's only in 0.44.x). Same for the test helpers: `Tag::custom(TagKind::Custom(Cow::Borrowed("expiration")), ..)` is now just `Tag::custom("expiration", ..)`, matching how `admin_cancel.rs` and `admin_settle.rs` already write custom tags on `main`.

**`nostr_sdk::secp256k1` no longer resolves,** and `sign_schnorr` now takes `AsRef<[u8]>` rather than a `secp256k1::Message`. So `app.rs` goes from `sign_schnorr(&nostr_sdk::secp256k1::Message::from_digest([7u8; 32]))` to `sign_schnorr([7u8; 32])`.

**`extract_lnurl` returns `Url`, not `String`** (from the LNURL scheme validation on `main`). The port-override test now asserts `extracted.to_string()` against the formatted URL, keeping both `main`'s type and this branch's `MOSTRO_TEST_LN_PORT` override.

### The part worth flagging

Commit `45ca902` adds `a_custom_named_expiration_tag_normalises_and_suppresses_the_auto_add`, and the comment I wrote on it 11 days ago says:

> If an sdk upgrade ever stopped normalising, the auto-add would start firing on top of the caller's tag and this test goes red.

The sdk upgrade arrived. The test is green against 0.45 — nostr still normalises a custom-named `expiration` tag to the canonical NIP-40 kind, so `has_expiration_tag`'s single check is still correct and order events are not double-stamping expirations. That's the one behavioural question the 0.45 bump raised in this file, and it's now answered rather than assumed.

I rewrote that comment to describe the behaviour instead of the (now non-existent) `TagKind` variants, so it doesn't rot again.

### Verification on the rebased tree

- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --bin mostrod` — **1189 passed, 0 failed, 2 ignored**

One note on that test run: on a first pass, `lightning::invoice::tests::test_lnurl_validation_with_test_server` failed with `AddrInUse` on port 8080 — which is precisely what commit `88679a0` in this PR exists to fix. Re-running with `MOSTRO_TEST_LN_PORT=18080` gives a clean 1189/0. The failure was a live demonstration of the problem the commit addresses, not a regression.

Diff against `main` is unchanged in shape: 339 insertions, 13 deletions, same seven files.

Catrya (changes requested):
Request changes — the branch is 19 commits behind main and the merge doesn't
build. Not a mistake on your side: #892 changed `accept_event` under you while
this sat open, and `ci.yml` triggers `on: push` only, so no build ever ran here
to tell either of us.

Reproduced on the merge with current main (textually clean, no conflicts):

error[E0308]: mismatched types --> src/app.rs:1206:17
1199 | let result = accept_event(   <- arguments to this function are incorrect
1206 |     false,

Five of them (1206, 1234, 1259, 1297, 1324): the new tests pass `bool` where
main now takes `gate: Option<&SpamGate>`. #892 also inserted `event.verify()`
ahead of the spam gate.

The rebase isn't mechanical, though — three things need redoing rather than
re-applying:

**1. The `spam_gate.rs` change loses its reason to exist.** It weakens
`install_global_then_second_install_is_rejected` (no longer asserts the first
install succeeds) only because the new tests install the process-wide
`OnceLock`. On main the gate is injected as a parameter, so those tests can
build their own and never touch the global. That weakening should die in the
rebase.

**2. The nip33 story is stale, and the comment that survived is wrong.**
Commit 45ca902 says it drops the `TagKind::Custom("expiration")` arm — that arm
no longer exists on this base: the nostr-sdk 0.45.1 migration (#867, on main)
already replaced the whole `matches!(..) || matches!(..)` with
`t.kind() == "expiration"`. The production diff here is the comment alone.

And the comment claims something that doesn't hold: in 0.45 `Tag::kind()`
returns `&str` (crate source, `event/tag/mod.rs:142`), so the check compares the
serialized tag name and matches a `Tag::custom("expiration", …)` either way.
There's no dependency on the sdk normalising, and the canary the comment
promises ("if an sdk upgrade ever stopped normalising, this test goes red")
can't fire. Both tests are worth keeping — the comment and the commit message
are what need rewriting.

**3. The mutation score is measured against a tree that no longer exists.**
"9 rather than 10 because dropping the arm also drops its `||`→`&&` mutant" —
there is no `||` in `has_expiration_tag` on this base, so that mutant doesn't
exist in either version. `accept_event` also changed shape, so its mutant set
moved too. Worth re-running after the rebase.

On the `mutation-test` target: the underlying insight is good and worth writing
down — a test failing from port contention scores as a killed mutant, which
inflates exactly the number the target measures. Two reservations on the fix:

- I couldn't check whether `CARGO_MUTANTS_JOBS=1` changes anything (not
  installed here). If cargo-mutants still tests one mutant at a time by
  default, it's a no-op in CI and the rationale only applies to someone who
  passes `-j` by hand. If 27.x parallelises by default, forcing 1 makes the
  weekly full run much slower — and `mutation.yml` declares no
  `timeout-minutes`, so it inherits GitHub's 6-hour default. Worth confirming
  which, since that comment is the whole justification for the target.
- Pinning `MOSTRO_TEST_LN_PORT=18080` doesn't stop two workers colliding with
  each other — they'd both use 18080. It only avoids something else already
  holding 8080. The root fix is the test not depending on a fixed host port;
  serialising is the patch.

Nits: the env-var read is copy-pasted in three places (`lnurl.rs` prod, its
test, `invoice.rs` test) with the same parse and fallback — one helper keeps
them from drifting; and the Makefile default (18080) differs from the code
default (8080), so `cargo test` and `make mutation-test` exercise different
ports.

The good part, to be clear: pulling `is_stale` and `missing_inner_signature`
out with boundary tests is exactly what mutation testing should produce, and
the `accept_event` tests cover the first-contact PoW lane, which had nothing at
all. Worth landing once it's rebased.

ToRyVand:
Thanks @Catrya — rebased, #826 folded in as you ruled, force-pushed. 5 commits
off `main` @ `d2e114d`; #826 will be closed pointing here. History rewritten
rather than stacked on: the old messages described changes no longer on this
base (point 2).

The build failure needs `cargo build --all-targets` to show — all five broken
call sites are in test code.

**1. `spam_gate.rs` is out of the PR**, byte-identical to main. You had the
cause right: the old param was `is_v2: bool` reading the process-wide
`OnceLock`. On main the gate is injected, so each test builds its own and
passes `Some(&gate)`.

**2. The comment was wrong.** `Tag::kind()` is `&self.buf[0]`
(`event/tag/mod.rs:142`) and `Tag::custom("expiration", ..)` pushes that
literal into `buf[0]` (`:337`). Both constructors match by construction — no
normalisation exists, so the canary could never fire. Rewritten; test renamed
off `..._normalises_...`.

**3. Score re-measured** on `accept_event`, `create_event` and
`restore_session`: 17 mutants, **14 caught, 1 missed, 2 unviable**. The
survivor is the whole-function `restore_session_action -> Ok(())` replacement
— the one already documented rather than papered over.

**`CARGO_MUTANTS_JOBS` — you were right twice.** 27.1.0 is installed here and
keeps exactly one scratch dir on a 16-CPU host with no `-j`: one mutant at a
time by default. #826's `2` doubles, *and* #849's `1` is a no-op. Kept as an
explicit pin of the existing default, described as that rather than as a cap.
`MOSTRO_TEST_LN_PORT` is documented as not making the suite hermetic — two
workers still collide on 18080. Its three copy-pasted parses are now one
`lnurl::test_ln_port()`.

Kept from #826: the bash array for the PR job's `--file` flags. Dropped:
`JOBS=2`, the `.gitignore` entry, and the tautological `assert_eq!`.

That last one had a tail. Dropping the test left `60 * 60` generating two
mutants (`*`→`+`, `*`→`/`) that nothing kills — and the only test that kills
them is the assertion you just argued away. The constant is now `3_600`: the
product computes nothing and existed only to hand an operator to a mutation.
Two fewer mutants, no new test. The log reports minutes, not seconds.

**CodeRabbit's round on the pushed branch, since it changed the diff:** it
caught that the two first-contact tests paired a gift wrap with `Some(gate)` —
a combination production never produces, since the gate only reaches
`accept_event` on the v2 loop with `DM_EVENT_KIND`. Worse, I had rebuilt
fixtures `accept_event_ordering_tests` already has. Both tests moved there and
now use its `v2_event`/`accept` helpers on a real kind-14 event. Also took its
`MOSTRO_TEST_LN_PORT=0` guard: `0` parses as a valid `u16` but means "let the
OS choose" to a listener, while the URL would still say `:0`.

I did not take its third one — NUL-delimited `git diff -z` for the changed-file
list. Bash cannot hold NUL bytes in a variable, so the suggested
`changed_rs=$(git diff -z ...)` cannot work as written, and `core.quotePath`
defaults to on, so a newline in a path is emitted quoted on one line rather
than split across two. The array already removes the argv-injection vector,
which was the real problem.

Not a change here: `ci.yml` is `on: push` only, no `pull_request`, while
`cashu.yml`, `mutation.yml` and `markdown.yml` all have one — so no PR gets
fmt/clippy/build/test from upstream CI. This PR's checks are three skipped
jobs. That is the mechanism behind "no build ever ran here to tell either of
us". Happy to file it separately if useful.

ToRyVand:
Two nits from your review I hadn't answered. Both now closed out.

**The 18080 / 8080 divergence is deliberate, and the Makefile now says so.**
Plain `cargo test` keeps binding the code's own 8080 default, so the default
path stays exercised; only `make mutation-test` — which runs the suite hundreds
of times over — steps aside from a port a developer machine is likely to have
in use. Pushed as a comment on the target rather than left as something only
this thread explains.

**The `timeout-minutes` one turned out to point at something bigger, so I filed
it separately: #930.** The condition you raised it under doesn't apply —
cargo-mutants already tests one mutant at a time, so nothing here made the
weekly run slower. But checking the observation on its own terms: the scheduled
`mutation-baseline` job has been killed at GitHub's 6-hour limit on **every run
since at least July 26** — six for six, all at 6.01 h. `cargo mutants --list`
gives 2226 mutants and observed throughput is ~1 min each single-worker, so a
full pass needs 35+ hours and gets through roughly 15% before it is cut.
`continue-on-error: true` is why nobody saw it: it correctly stops a low score
from gating merges, but it also swallows the job never finishing, which is a
different thing.

I deliberately did not fix that here. Picking the timeout *is* the decision the
issue asks for — sharding, scoping to a rotating subset, or accepting a partial
run and saying so — and a testing PR is not the place to change how the
project spends runner minutes.

Which is worth saying plainly: I nearly let that nit go, because I checked
whether *this PR* made the run slower, found it didn't, and stopped there. The
observation underneath it was a month-old defect. Same shape as the mistake on
#860 — verify the line you're pointed at, then keep reading.

Catrya (changes requested):
Rebase and the three substantive points from the last round are properly
closed. I checked each rather than take them on report.

- Merges clean onto `main` @ `7e6b600`. On the merged tree:
  `cargo clippy --all-targets --all-features` clean, `cargo fmt --check` clean,
  and the 16 new/affected tests pass (`accept_event_tests` ×3, both
  first-contact ones, `is_stale` ×3, `missing_inner_signature` ×3, the two
  nip33 expiration tests, `extract_lnurl`).
- `spam_gate.rs` is out of the diff. 7 files.
- The nip33 comment is now correct. Against `nostr 0.45.4`:
  `pub fn kind(&self) -> &str { &self.buf[0] }`, and `Tag::custom(kind, ..)`
  pushes `kind.into()` into that cell. No normalisation, exactly as written.
- The `CARGO_MUTANTS_JOBS` claim holds, and I verified it upstream rather than
  from the report: `cargo-mutants` v27.1.0, `src/lab.rs`:
  `let n_threads = max(1, min(options.jobs.unwrap_or(1), mutants.len()))`.
  One at a time is the default, so the pin is a genuine no-op. My reservation
  is answered.
- `test_ln_port()` collapses the three copies and rejects `0`.

Pulling `is_stale` and `missing_inner_signature` out with boundary tests is
still the best part of this PR.

Two things I'd hold on, then smaller ones.

---

**1. `.mutants.toml` is never read, and one half of that is this PR's own
thesis.**

`config.rs` in v27.1.0: `read_tree_config` looks at
`workspace_dir.join(".cargo").join("mutants.toml")`. This repo has
`.mutants.toml` at the root and no `.cargo/` directory, so the file has never
been loaded. Moving it wouldn't be enough either — `Config` is
`#[serde(default, deny_unknown_fields)]` and has no `jobs`, `timeout`,
`output_dir` or `test_tool_options` field (the real names are
`minimum_test_timeout`/`timeout_multiplier`, `output`,
`additional_cargo_test_args`). Four of its five interesting keys are invalid.

This predates the PR (it came in with #619, and `docs/MUTATION_TESTING.md`
documents the file as active in three places). But it splits into two parts,
and only one of them is somebody else's problem.

*The part that belongs here.* The dead config also carries
`test_tool_options = ["--", "--test-threads=1"]`, commented "Use single thread
for DB tests to avoid SQLite concurrency issues". So this PR removes score
inflation from parallelism **between mutant workers** — which was already 1 by
default — while parallelism **between test threads inside each run**, which the
project explicitly tried to switch off for the same reason, is still on. That
is this PR's own argument applied at the level the PR doesn't reach. Worth
either handling here or saying plainly in the Makefile comment that the run is
serial at the worker level only.

*The part that belongs on #930.* Its analysis rests on "~2226 mutants" and
offers sharding / scoping / accepting a partial run — but option 2 is already
written down in `exclude_globs` (`scheduler.rs`, `lightning/**`, `rpc/**`,
`config/**`, `main.rs`, `cli.rs`) and simply never took effect, so the 2226 is
measured against a scope nobody chose. Not a rescue: those files are ~9.8k of
~65k lines under `src/`, so fixing the path doesn't bring a 35-hour run under
6 hours and sharding is probably still the answer. It changes the starting
point, not the conclusion. Worth a comment there so the decision isn't made on
the wrong numbers.

(Side note: `exclude_globs` lists `src/lightning/**`, which the PR job would
feed to `--file` anyway, since this PR edits `src/lightning/invoice.rs`.)

**2. The PR body is stale, and for #636 the body *is* the deliverable.**

The acceptance criterion is "Mutation score documented in PR". The body still
reports `9/9 mutants confirmed killed — 0 missed, 0 timeout, 0 unviable`, while
the latest comment reports 17 mutants, 14 caught, **1 missed**, 2 unviable. It
also still describes the `spam_gate.rs` fix (no longer in the PR), removing the
`TagKind::Custom("expiration")` arm (which doesn't exist on this base — you
said so yourself), `accept_event_tests` as the home of the spam-gate/PoW branch
(it moved to `accept_event_ordering_tests`), and `cargo test: 1059 passed` (now
~1282). `restore_session.rs`, the only production behaviour change in the PR,
isn't mentioned at all. The surviving mutant in particular has to be in the
body, not four comments down.

---

**3. `accept_event_tests` duplicates a fixture and dodges the DB path.**

`create_test_ctx()` builds an **unmigrated** `:memory:` pool, while
`create_migrated_ctx()` already exists in the enclosing `mod tests` and
`use super::*` is in scope. The tests only pass because `RestoreSession`
short-circuits `check_trade_index` — it returns `Ok(())` for anything that
isn't `NewOrder`/`TakeBuy`/`TakeSell` — so the unmigrated pool is never
touched. That's the same duplicate-fixture point CodeRabbit raised and you
fixed for the two first-contact tests; this module kept its own copy, plus four
imports that exist only to feed it. Switching to `create_migrated_ctx()` drops
the duplication and stops the module depending on the action choice to stay
green — and then the "accepts" test is free to use `NewOrder`, which is the
path that actually does work.

Related: there are now two modules testing `accept_event` with two fixture
sets. Folding `accept_event_tests` into `accept_event_ordering_tests` is the
consistent ending.

**4. `restore_session.rs` — scope and shape.**

Separate commit, which is right, but it's the only production behaviour change
here and it's absent from the body (see 2).

`3_600` instead of `60 * 60`, justified in four lines of doc comment as denying
a mutation operator, is production source shaped by the tool: it improves the
score by *removing* a mutant rather than killing one, which is a mild version
of the inflation the rest of this PR is about. House style runs the other way —
`src/util.rs` has `const ORDER_TS_MAX_AGE_SECS: u64 = 48 * 3600;`. Keep the
named constant, that part is a real improvement; move the paragraph about
mutation operators into the commit message and write the value however reads
best.

The stated goal — that the message and the timeout "cannot drift apart" —
isn't quite reached: the log hardcodes the unit and divides, so a constant of
90 prints "1 minutes" and 45 prints "0 minutes". Log the seconds, or the
`Duration`. It also changes an operator-facing string from "1 hour" to "60
minutes" with no functional need.

**5. Smaller.**

- `set -o pipefail` in the `mutation-test` recipe has no pipeline to guard.
- The PR job now hardcodes `CARGO_MUTANTS_JOBS=1 MOSTRO_TEST_LN_PORT=18080`,
  duplicating the Makefile's environment with a comment explaining why the
  Makefile can't be reused. The reasoning is right, but it leaves two places
  that have to agree — the same drift you just removed by creating
  `test_ln_port()`.
- **#849 and #842 conflict in `src/app.rs`.** I merged them:
  `CONFLICT (content): Merge conflict in src/app.rs`. Both touch the
  missing-inner-signature block — #849 replaces the condition, #842 rewrites
  the `warn!` inside it. Whichever lands second needs a one-hunk resolution
  keeping `missing_inner_signature(...)` as the condition and #842's redacted
  `warn!(… event.id)` as the body. Worth deciding the order deliberately
  rather than discovering it at merge.

ToRyVand:
All five addressed. Rebased onto `afc39a2` (v0.18.7); still 5 commits, the fixes folded into the ones they belong to rather than stacked on top.

## 1. `.mutants.toml`

Split as you suggested, but the part that belongs here has a smaller fix than moving the file.

**Here: `RUST_TEST_THREADS=1`.** The problem was never "the config file is in the wrong place", it was "test threads run in parallel". libtest honours that env var with no config file at all — verified, the suite goes **18.5s → 53.8s**, so it is genuinely taking effect. One variable on a line that already set two: no `.cargo/`, no key renames, no doc rewrites, and nothing in this PR depending on #958 landing.

The Makefile comment now says both levels explicitly and points at #958 for why the config's own attempt never worked.

**There: #958.** Filed with the measurements rather than the report. Your path and invalid-key analysis is credited there; what I added is that moving the file is not merely insufficient, it **breaks** — `Config` is `deny_unknown_fields`, so `cargo mutants` errors out with `unknown field 'jobs'` instead of falling back to defaults. Tested key by key; exactly four are invalid and your four replacements are all correct. Also: the docs describe the file as working in **four** places, and `MUTATION_TESTING.md:255` tells the reader to raise a timeout in a file that is never read.

**And on #930:** commented with the corrected numbers. `cargo mutants --list` gives **2373** today against **1943** with the config fixed — so the 2226 that issue reasons from is measured against a scope nobody chose. You are right that it does not rescue the run: 1943 × ~1 min is still ~32h against a 6h limit, so sharding remains the answer. But it does weaken option 3 there — "accept a partial run and document the score" assumed the partial score is honest, and without `--test-threads=1` every score this project has recorded is inflated by port collisions counted as kills.

## 2. The body

Rewritten around a re-measured score. You were right that this is the deliverable and not a footnote:

| Scope | Mutants | Caught | Missed |
|---|---|---|---|
| `accept_event` + `create_event` | 10 | 10 | 0 |
| `restore_session.rs` | 6 | 3 | **3** |
| **Total** | **16** | **13** | **3** |

**81%**, with all three survivors named in the body and the commands to reproduce them. Two are the `60 * 60` arithmetic (see 4). The third is the whole-function `restore_session_action -> Ok(())` you can see in the earlier comment, and chasing it turned up why it cannot be killed: its only fallible call, `start_restore_session`, **always returns `Ok(())`** — the DB error is handled inside the spawned blocking closure and logged at `db.rs:1882-1884`, never propagated. So the `?` cannot fire, and the mutant's only effect is skipping the `tokio::spawn`. The survivor is telling us that error path is already dead, not that the tests are thin.

`restore_session.rs` is in the body now, the `spam_gate.rs` and `TagKind::Custom` paragraphs are gone, and the test count is current.

## 3. `accept_event_tests`

Folded into `accept_event_ordering_tests` on `create_migrated_ctx()`; the duplicate fixture and its four imports are gone, −105 lines.

I did **not** switch the "accepts" test to `NewOrder`. With a migrated pool the module no longer depends on the action choice to stay green, which was the actual defect; `NewOrder` would need a hand-built `Payload::Order`, and `check_trade_index_tests` covers that path directly with eleven tests. Happy to do it if you'd rather it went through `accept_event` too.

## 4. `restore_session.rs`

`60 * 60` is back, matching `ORDER_TS_MAX_AGE_SECS` in `src/util.rs`, and the paragraph about mutation operators moved to the commit message. You were right about the shape of that mistake and it is the one I'd least like to have made in this PR: writing `3_600` improves the score by *deleting* a mutant rather than killing one, which is the same inflation the rest of the branch exists to remove. Both arithmetic mutants now survive and are reported.

On the log: it prints the `Duration` handed to `tokio::time::timeout` rather than a derived string, so it cannot disagree with the timeout for any value of the constant. That also drops the "1 minutes" / "0 minutes" behaviour and the gratuitous "1 hour" → "60 minutes" change.

## 5. Smaller

- `set -o pipefail` removed — you were right, no pipeline.
- The CI job still sets its own env, since it must build `--file` args as a bash array from PR-diff filenames and cannot go through `make ARGS=`. Added an explicit keep-in-sync comment rather than machinery; if you'd prefer the values in one place I'll do it, but a shared env file felt like more than the drift is worth.
- **Order decided, not discovered:** I've asked on #842 for that one to land first and this to take the rebase — it is a single hunk and #842 is down to review. The resolution has changed shape there (its `warn!` no longer carries `event.id`), so the current version is written out in that thread.

---

One thing worth flagging rather than burying: serialising the test threads makes the mutation run roughly **three times slower**. That is the correct trade and I'd rather say it out loud than have it discovered — it was faster before because it was measuring wrong.

Green: `cargo fmt --check`, `cargo clippy --all-targets --all-features`, `cargo test --bin mostrod` — **1336 passed**, 0 failed, 2 ignored.

Catrya (changes requested):
**1. Drop `RUST_TEST_THREADS=1`. The reason given for it doesn't hold, and it has a real cost.**

Only one test binds the fixed port: `test_lnurl_validation_with_test_server` (`src/lightning/invoice.rs:327`). Every other listener in the suite binds `:0`, and every test gets its own private `sqlite::memory:` pool. So test threads inside one run can't collide on the listener. The port collision only happens between concurrent processes, and `CARGO_MUTANTS_JOBS=1` already covers that. Last round I framed thread-level parallelism as this PR's own argument applied one level down. That was my mistake: the argument doesn't reach that level.

The cost, measured with `make mutation-test` on this branch: the suite takes 68s serial vs 20s parallel, and mutants averaged 87s each over 25. The PR job runs every mutant in every changed file. For this PR's own five files that's 252 mutants: about 6.1h serialised vs about 2.7h before. That's past GitHub's 6h default on a job with no `timeout-minutes` and `continue-on-error: true`, so it gets killed without anyone noticing. That's #930's failure mode, now in the PR job too.

Please remove it from both the Makefile target and `mutation.yml`, and have the Makefile comment say the run is serial at the worker level only. If a test actually flakes under parallel threads, name it in the comment as the reason and keep the variable.

**2. The body, in two places.**

- The mutation table leaves out the two predicates the PR extracts. `-F 'in accept_event'` doesn't match mutants in `is_stale` or `missing_inner_signature`. That's 9 mutants that counted as "in accept_event" before the extraction (the `<` in the stale check, the `!=`/`&&` in the signature check). I ran them: all 9 are caught. Please add `-F 'is_stale' -F 'missing_inner_signature'` to the reproduce command and a row to the table, for 22/25 (88%). Also, the "accept_event + create_event: 10" row is really 6 + 2 plus two `check_trade_index` struct-field mutants that cargo-mutants 27.1.0 lets through the `-F` filter. Worth labelling it as such.

- Under "Tests", these four are listed as this PR's, but they're already on `main` from #892: `tampered_copy_does_not_censor_the_genuine_event`, `genuine_duplicate_is_still_dropped_as_a_replay`, `invalid_signature_is_dropped_without_a_gate` and `v1_gift_wrap_with_invalid_signature_is_dropped`. What this PR adds to `app.rs` is the two first-contact tests, the three accept/wrong-kind/wrong-receiver tests and the predicate boundary tests.

**3. A doc comment got detached.**

The two first-contact tests were inserted between `gate_applies_to_v2_only`'s doc comment and its `fn`. Now "The v2-only policy lives in one place now…" documents `unknown_first_contact_sender_clearing_the_pow_bar_is_accepted`, and `gate_applies_to_v2_only` has no doc. Move the new tests above that comment.

**Nits**
- `mutation.yml`: "see the Makefile for why both variables are set" doesn't match the line that sets three. "The string-then-make(ARGS)-then-shell round trip this used to take" describes an earlier revision of this branch; `main` word-split an unquoted `$file_args`.
- `wrap_test_order` wraps a `RestoreSession`, not an order. `wrap_test_message` would say what it does.

ToRyVand:
All three addressed, plus the nits. Still 5 commits, with the fixes folded into the ones they belong to.

## 1. `RUST_TEST_THREADS=1` — removed, and you were right

Gone from both the Makefile target and `mutation.yml`. Since it was my claim you were retracting, I checked it before taking it out:

- `test_lnurl_validation_with_test_server` is the only fixed-port bind in `src/`. Every lookup it makes (the LNURL and both lightning addresses) goes to its own listener, and nothing else in the suite connects to that port. Every other listener binds `:0`.
- Every test pool is `:memory:` with no `cache=shared`, so the "SQLite concurrency" reason in `.mutants.toml` doesn't hold either. The only shared process state I found, `MOSTRO_NSEC_PRIVKEY`, is already serialised by `ENV_LOCK` in `config/util.rs`.
- Measured as well: six consecutive `cargo test --bin mostrod` runs with parallel threads, 1336/1336 each time, ~18s.
- `cargo mutants --list` over the five `.rs` files gives your 252 exactly.

The Makefile comment now says the run is serial at the worker level only, names the one fixed-port test, and says why threads stay parallel, so nobody re-adds the setting from the dead config. It also used to say "the LNURL tests" bind a fixed port. It's one test, and the commit message said the same thing. Both are fixed.

The mistake is mine more than yours. Last round you gave two options, and I picked the one that rested on a premise. Then I checked that the variable **took effect** (18.5s → 53.8s) and never checked that it **fixed anything**. Timing can prove a setting is applied. It can't prove the setting is needed.

It also didn't stay in this PR. I published the same inflation argument on #930 ("every score this project has recorded is inflated") and in #958. Worse, #958's suggested TOML carried `additional_cargo_test_args = ["--", "--test-threads=1"]`, so whoever implemented it as written would have doubled the time per mutant on the weekly baseline, the job that already dies at 6h. I've corrected both in place with the original struck through, and #958's suggested config no longer carries the setting.

## 2. The body

- I re-measured the table under the new target instead of taking the numbers from your run: 25 mutants, 22 caught, 3 missed, 0 unviable: the same 22/25 you got, and the survivors are the three the body already names. With threads parallel it ran at ~41s per mutant, roughly half what the serialised runs took on the same machine and in line with your 87s. So the setting doubled the cost and changed no outcome. The predicates row is in, and the reproduce command now has `-F is_stale -F missing_inner_signature`: **22/25 = 88%**. The "10" row is labelled as 6 + 2 + the two `check_trade_index` struct-field mutants, with a footnote on the filter.
- "Tests" now lists only what this PR adds, and names the four #892 tests as not ours.
- One more stale line you didn't flag: the nip33 bullet still said those tests came from "a surviving `||`→`&&` mutant". There is no `||` on this base, as you pointed out in round 1. That line is gone.

## 3. Doc comment

Moved. The first-contact tests now sit above `gate_applies_to_v2_only`'s doc comment, and the function has its doc back.

## Nits

- `mutation.yml`: "both variables" → "each variable", "Keep these three" → "two", and the injection comment now describes what `main` actually does (an unquoted `$file_args` word-split by the shell) instead of an earlier revision of this branch.
- `wrap_test_order` → `wrap_test_message`.
- While I was there: `test(app)`'s commit message still said the v1-gate tests "go in a new module", which stopped being true when round 2 folded them into `accept_event_ordering_tests`, and it never mentioned the predicate extraction. Reworded.

Green: `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1336 passed**, 0 failed, 2 ignored.

Merge order unchanged: #842 first.

PR #852, opened 2026-08-07
docs: note the step-2 cooperative-cancel guard is now a backstop

grunch:
## Summary

Follow-up to #851. Comment-only change — no behavior change.

#851 added two checks to `cancel_active_order`: a membership check rejecting senders outside `{buyer, seller}`, and a check requiring the recorded `cancel_initiator_pubkey` to equal `counterparty_pubkey`. Together those make the pre-existing same-party guard in `cancel_cooperative_execution_step_2` (`src/app/cancel.rs:88-93`) unreachable from its only caller: a sender that is also the initiator is rejected before it gets there.

The guard is worth keeping as defense in depth, but as it stands a reader landing on `cancel_cooperative_execution_step_2` would reasonably conclude that this `if` is what stops a third party from confirming a cooperative cancel. It is not, and that matters — the enforcing check is upstream, and anyone reasoning about the authorization from this function alone would be reasoning about the wrong code.

This adds a comment recording that, and pointing at where the authorization actually lives.

## Stacking

Branched off `fix/coop-cancel-third-party-auth` (#851), targeting `main`. Until #851 merges, the diff here shows its commits too; once it lands this collapses to the single comment hunk.

## Verification

```bash
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test --bin mostrod app::cancel::tests   # 21 passed, 0 failed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Cooperative cancellation now rejects cancellation requests from unauthorized parties.
  * Confirmation requests are rejected when the recorded initiator is not the expected counterparty.
  * Invalid cancellation attempts no longer alter the cancellation state.

### DIFF
--- a/src/app/cancel.rs
+++ b/src/app/cancel.rs
@@ -614,8 +614,11 @@ async fn cancel_active_order<L: CancelLightning + Send>(
     }
 
     // If there is already an initiator recorded, this call becomes the confirmation (step 2).
-    match order.cancel_initiator_pubkey {
-        Some(_) => {
+    match order.cancel_initiator_pubkey.as_deref() {
+        Some(initiator) => {
+            if initiator != counterparty_pubkey {
+                return Err(MostroCantDo(CantDoReason::InvalidPubkey));
+            }
             cancel_cooperative_execution_step_2(
                 ctx,
                 event,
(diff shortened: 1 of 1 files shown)


grunch:
Closing — the underlying fix (#851) is merged and this was comment-only, not worth a PR of its own. If the step-2 backstop ever needs explaining, it can ride along with the next change to src/app/cancel.rs.

PR #853, opened 2026-08-07
fix: resubscribe hold invoices correctly across a restart

grunch:
## Summary

Second of the five findings from the security review (the first landed in #851).

Three defects broke hold-invoice resubscription after a daemon restart. Together they mean every in-flight trade across any restart is mishandled, and with `[anti_abuse_bond] slash_on_waiting_timeout = true` a seller who paid the hold invoice **on time** has their bond slashed as the responsible party — a loss of user funds caused by the daemon's own missed observation.

## What was wrong

**1. The r_hash was passed as ASCII hex.** `main.rs` called `invoice_subscribe(hash.as_bytes().to_vec(), ...)`, but `orders.hash` is a 64-character hex string, so `.as_bytes()` yields 64 ASCII characters rather than the 32 raw bytes `SubscribeSingleInvoiceRequest.r_hash` expects. Every resubscribe errored instantly, and the error was only logged (`tracing::error!`), so it was silent in practice. The bond resubscriber already did this correctly with `from_hex` — this now uses the existing `decode_hash32` helper, which also length-checks and keeps the value out of the log line.

**2. The query excluded the window that matters.** `find_held_invoices` selected only `status == 'active'`, skipping `waiting-payment` and `waiting-buyer-invoice` — the phases where the `Accepted` event is load-bearing, because that is when the seller actually pays.

Worth flagging, since it is not obvious and the finding did not mention it: widening the status filter alone would **not** have fixed this. Those rows also have `invoice_held_at = 0`, since that column is only ever written by `hold_invoice_paid` — i.e. after the seller pays. The existing `invoice_held_at != 0` predicate would have kept excluding them. The new branch keys off `hash IS NOT NULL` instead, and is disjoint from the original `active` branch so that path is unchanged.

**3. `hold_invoice_paid` had no current-status guard.** Once resubscription works, a redelivered or late `Accepted` would advance whatever order it resolved to. It now no-ops outside `WaitingPayment` / `WaitingBuyerInvoice`, with a `warn` carrying the order id and status — loud on purpose, because the HTLC is real and locked while the order can no longer be credited, so an operator has to look.

That third change also closes the amplifier in finding #5 (the stale full-row write on `cancel_pending_order_from_maker`), where a canceled order could be revived by the seller paying a hold invoice that outlived the row's escrow columns. The clobber itself is untouched and still open.

## Tests

Written first, confirmed failing against the unfixed code:

- `db::tests::test_find_held_invoices_includes_orders_awaiting_seller_payment` — both pre-payment statuses are selected with `invoice_held_at = 0`.
- `db::tests::test_find_held_invoices_ignores_orders_without_hash` — no hash, nothing to subscribe to.
- `flow::tests::hold_invoice_paid_does_not_revive_a_settled_order` — a stale `Accepted` on a canceled order leaves status and `invoice_held_at` untouched, and returns `Ok`.

Existing coverage for the `active` branch and both happy paths of `hold_invoice_paid` is unchanged and still passes.

## Verification

```bash
cargo test --bin mostrod          # 1095 passed, 0 failed, 2 ignored
cargo clippy --all-targets -- -D warnings
cargo fmt --all -- --check
```

## Not covered here

The wrongful-slash consequence was code-traced through the scheduler and bond slash path, not executed against a live LND. The two enabling defects are test-proven; the fix removes their cause.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored eligible invoices in “waiting for payment” and “waiting for buyer invoice” states when valid invoice information is available.
  - Prevented duplicate or stale payments from reviving canceled, settled, or already-processed orders.
  - Improved invoice resubscription by validating stored order hashes and safely skipping invalid entries.
  - Added safeguards to exclude incomplete invoice records from recovery and prevent invalid payment notifications.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -919,12 +919,25 @@ pub async fn find_locked_cashu_orders(pool: &SqlitePool) -> Result<Vec<Order>, M
     .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))
 }
 
+/// Orders whose hold invoice still needs an LND subscription attached.
+///
+/// Two disjoint cases, and the second is the one that matters after a restart:
+///
+/// - `active` with `invoice_held_at != 0` — the seller already paid, so the
+///   subscription is there to observe settle/cancel.
+/// - `waiting-payment` / `waiting-buyer-invoice` with a hash — the invoice
+///   exists but nobody has paid it yet. `invoice_held_at` is still 0 here
+///   (only [`crate::flow::hold_invoice_paid`] ever writes it), which is why
+///   that column cannot gate this branch: these rows would never match.
+///   Missing this window is what makes a seller who paid during a restart
+///   look like a seller who never paid.
 pub async fn find_held_invoices(pool: &SqlitePool) -> Result<Vec<Order>, MostroError> {
     let order = sqlx::query_as::<_, Order>(
         r#"
           SELECT *
           FROM orders
-          WHERE invoice_held_at !=0 AND  status == 'active'
+          WHERE (invoice_held_at != 0 AND status == 'active')
+             OR (status IN ('waiting-payment', 'waiting-buyer-invoice') AND hash IS NOT NULL)
         "#,
     )
     .fetch_all(pool)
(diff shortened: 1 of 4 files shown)


grunch:
@coderabbitai review

grunch:
Review follow-up pushed in 637e793.

**Finding 1 (P1) — fixed.** The guard now also requires `invoice_held_at == 0`. That column is written at the end of the flow and read nowhere else in the codebase, which is what makes it a trustworthy already-processed marker — status alone cannot be one, because `WaitingBuyerInvoice` is both an entry and an exit state of this function. Regression test `flow::tests::hold_invoice_paid_is_a_noop_on_redelivery` was written first and reproduced the reported overwrite exactly (`1700000000` → now) before the fix.

Worth recording that the reported symptom and the actual harm differ. `invoice_held_at` has no reader other than `find_held_invoices`, whose `active` branch only tests `!= 0`, so overwriting it changes no decision. The real cost of the replay is the duplicated side effects: another `AddInvoice` to the buyer, another reputation notification to the maker, and a republished Nostr event, on every restart.

**Finding 4 — fixed.** `find_held_invoices`' `Err` is logged instead of being swallowed by `if let Ok`.

**Finding 5 — fixed.** Guard moved above `get_buyer_pubkey`/`get_seller_pubkey`. Verified safe: the only test exercising the missing-pubkey path seeds `WaitingPayment`, so it still returns `NostrError`.

**Finding 6 — covered.** Added `db::tests::test_find_held_invoices_includes_waiting_buyer_invoice_after_payment` plus the redelivery test above.

**Findings 2 and 3 — filed as follow-ups:** #855 (read-then-act guard should be an atomic claim, mirroring `on_bond_invoice_accepted`) and #856. Note #856 is broader than reported: `fiat-sent` has the same gap as `dispute`, and neither is a regression from this PR.

`cargo test --bin mostrod`: 1097 passed, 0 failed. clippy `-D warnings` and fmt clean.

grunch:
On the CodeRabbit review:

**Nitpick (doc comment) — applied** in 223c724. The guard's reasoning was documented inline but the function had no contract at the top saying it can fire more than once and no-ops on a replay. Fair catch.

**Inline (atomic claim) — correct diagnosis, but skipping here.** The read-then-act observation is right and already tracked in #855. The restructure as described would regress correctness if applied as stated, though.

`enqueue_order_msg` pushes onto an in-memory `RwLock<Vec<...>>` (`src/util.rs:1355`); nothing about it is durable. That makes today's ordering — send, then write `invoice_held_at` — accidentally crash-safe in the useful direction: die in between and the messages are lost, but `invoice_held_at` is still 0, so the next restart replays and regenerates them. Claim first and a crash after the claim loses the messages permanently *and* suppresses the replay that would have rebuilt them, leaving the buyer without `AddInvoice` and the escrow locked. Duplicate messages traded for lost ones.

The review does hedge with "or transaction/outbox mechanism", and that is exactly the point: the claim needs a durable outbox first. That is an architecture change, not a minimal one, and it does not belong in a PR whose job is the restart resubscribe. Recorded the constraint and the required ordering on #855.

The suggested concurrent-redelivery test also does not look reachable: `invoice_subscribe` has two production call sites, `show_hold_invoice` and the restart resubscribe, which belong to different process lifetimes for the same order. Two `Accepted` deliveries cannot race each other within one process. The race that *is* reachable — subscriber task versus a cancel on the main loop — is what #855 covers.

`cargo test --bin mostrod`: 1097 passed, 0 failed. clippy and fmt clean.

PR #862, opened 2026-08-12
fix: enforce trade escrow CLTV deadline

grunch:
## Summary

Third finding from the security review (#851, #853 landed before).

LND auto-cancels an **accepted** hold invoice `invoices.holdexpirydelta` blocks (default 12) before the HTLC's CLTV expiry, refunding the seller off-chain. mostrod only *logged* that cancellation, and no scheduler job covered `Active`/`FiatSent` orders — so with the default `hold_invoice_cltv_delta = 144`, **~22 hours after the seller paid, the trade escrow silently evaporated while the order still looked live**. A buyer who sent fiat after (or before) that point lost it: `release` fails with "invoice already canceled", a dispute can't help (`admin_settle` settles the same dead invoice), and a malicious seller keeps both the fiat and the refunded sats. No attacker sophistication needed — just stalling.

## The fix — two layers

**1. Guardian job** (`job_enforce_escrow_deadline`, Lightning mode only, 5-min tick). At `cltv_delta - escrow_deadline_margin_blocks` blocks after the escrow was observed paid, it resolves the trade **while the escrow still exists**:

- `active` — nobody claims fiat moved: cancel the hold invoice proactively (clean, observed refund instead of LND's silent one), order → `canceled`, both parties notified, range maker bonds closed like the waiting-state timeout path. Cancel failures use the bond module's `classify_cancel_error`: already-done proceeds, transient retries next tick — unless the hard horizon already passed, in which case the state is fixed regardless.
- `fiat-sent` — the fiat leg may already be with the seller: open a dispute **without touching the escrow** (it is the solver's only leverage). A dispute already *open* is left alone; a row left over from a dispute the users resolved themselves is **reopened** (there is at most one dispute row per order); otherwise a fresh buyer-initiated dispute is created, both parties are notified, and the dispute event is published best-effort.

**2. Reactive handler** (`hold_invoice_canceled`). Covers the daemon-down window: when a cancel lands on an escrow-backed order (`Active`/`FiatSent`/`Dispute`) **past** the guardian's action deadline — e.g. the restart replay of an LND auto-cancel — it now notifies both parties (`HoldInvoicePaymentCanceled`), raises an operator-facing `error!`, and closes `Active` orders instead of leaving them looking live. Before the deadline it stays log-only, so in-flight cooperative/admin cancels are not misread as evaporations.

## Config change

New `[lightning]` setting **`escrow_deadline_margin_blocks`** (default 24, serde-defaulted so existing `settings.toml` files keep working). It must exceed the LND node's `invoices.holdexpirydelta` (LND default 12); a misconfigured margin `>= hold_invoice_cltv_delta` is clamped to half the window with an `error!` instead of firing on brand-new escrows. Documented in `settings.tpl.toml` and `docs/STARTUP_AND_CONFIG.md`.

**Intentional behavior change:** `Active` trades no longer linger indefinitely — they are canceled at ~`cltv_delta - margin` blocks (~20 h on defaults). Operators with slow fiat rails can raise `hold_invoice_cltv_delta`.

## Compatibility note

No wire-format or tag changes. Clients may observe existing actions in new situations: `HoldInvoicePaymentCanceled` (previously never emitted), `Canceled` for expired-escrow trades, and daemon-initiated `DisputeInitiatedByYou`/`DisputeInitiatedByPeer`.

## Tests

Written against the unfixed code first where applicable (the PoC for the finding demonstrated `hold_invoice_canceled` leaving a `FiatSent` order untouched and `release` failing on a canceled invoice):

- `db`: `find_escrow_deadline_orders` selects exactly the due `active`/`fiat-sent` rows with a hash (excludes `dispute`, waiting states, no-hash, `invoice_held_at = 0`, not-yet-due).
- `util`: pure CLTV-deadline math, including the misconfiguration clamp.
- `flow`: `hold_invoice_canceled` closes a past-deadline `Active` order, ignores pre-deadline cancels, leaves `FiatSent` for a human, and keeps the legacy no-op when `invoice_held_at = 0`.
- `scheduler`: the guardian pass with a stub escrow — due `Active` canceled + seller refunded + parties notified; transient cancel failure before the horizon retries without state change; already-canceled at LND still fixes state; `FiatSent` opens a dispute and never touches the escrow; an open dispute is left to the solver; a resolved dispute is reopened; inside-window orders untouched.

```text
cargo test --bin mostrod          # 1133 passed, 0 failed, 2 ignored
cargo clippy --all-targets --all-features   # clean
cargo fmt --all -- --check                  # clean
```

## Not covered here

Orders already in `Dispute` are not touched by the guardian (a human is already in the loop); if the solver does not act before the horizon, the reactive path raises the operator alarm. A client-side display of the per-trade escrow deadline (the daemon already advertises `hold_invoice_cltv_delta` in the info event) is follow-up material. The LND auto-cancel behavior (`InvoiceExpiryWatcher`, `holdexpirydelta = 12` default since v0.13) was code-verified upstream, not exercised against a live node here.

## Summary by CodeRabbit

* **New Features**
  * Added configurable escrow deadline safety margins, defaulting to 24 Lightning blocks.
  * Added automatic monitoring of overdue escrow orders.
  * Active orders are canceled after the deadline with notifications to both parties.
  * Fiat-sent and disputed orders now open or reopen disputes when appropriate.
  * Added setup and configuration documentation for the new setting.

* **Bug Fixes**
  * Improved handling of canceled hold invoices, retries, missing timestamps, and already-canceled invoices.

### DIFF
--- a/docs/STARTUP_AND_CONFIG.md
+++ b/docs/STARTUP_AND_CONFIG.md
@@ -110,6 +110,7 @@ Configuration is loaded from `~/.mostro/settings.toml` (template: `settings.tpl.
 - `hold_invoice_expiration_window` (u32): Hold invoice expiration in seconds (default: 300)
 - `payment_attempts` (u32): Max payment retry attempts (default: 3)
 - `payment_retries_interval` (u32): Retry interval in seconds (default: 60)
+- `escrow_deadline_margin_blocks` (u32): Safety margin in blocks before the hold invoice's CLTV horizon; at `hold_invoice_cltv_delta - escrow_deadline_margin_blocks` blocks after the escrow was paid, mostrod cancels the trade (Active) or opens a dispute (FiatSent) before LND auto-refunds the escrow. Must exceed the LND node's `invoices.holdexpirydelta` (LND default: 12) (default: 24)
 
 **Mostro** (`src/config/types.rs:76-108`):
 
--- a/settings.tpl.toml
+++ b/settings.tpl.toml
@@ -22,11 +22,12 @@ payment_retries_interval = 60
 max_final_cltv_expiry_delta = 432
 # Safety margin (in blocks) before the hold invoice's CLTV horizon. LND
 # auto-cancels an accepted hold invoice `invoices.holdexpirydelta` blocks
-# (default 12) before its HTLC expires, refunding the seller; at
-# hold_invoice_cltv_delta - escrow_deadline_margin_blocks blocks after the
-# escrow was paid, mostrod cancels the trade (Active) or opens a dispute
-# (FiatSent) while the escrow still exists. Must comfortably exceed your
-# node's holdexpirydelta.
+# (default 12) before its HTLC expires, refunding the seller; once the
+# chain tip is within escrow_deadline_margin_blocks of the accepted
+# HTLC's expiry height (nominal 10-minute-block clock as fallback when
+# LND cannot answer), mostrod cancels the trade (Active) or opens a
+# dispute (FiatSent) while the escrow still exists. Must comfortably
+# exceed your node's holdexpirydelta.
 escrow_deadline_margin_blocks = 24
 
 [nostr]
--- a/settings.tpl.toml
+++ b/settings.tpl.toml
@@ -20,6 +20,14 @@ payment_retries_interval = 60
 # Maximum min_final_cltv_expiry_delta (in blocks) accepted on a payout invoice
 # supplied by a user. Bounds how long a payee can hold the outgoing HTLC.
 max_final_cltv_expiry_delta = 432
+# Safety margin (in blocks) before the hold invoice's CLTV horizon. LND
+# auto-cancels an accepted hold invoice `invoices.holdexpirydelta` blocks
+# (default 12) before its HTLC expires, refunding the seller; at
+# hold_invoice_cltv_delta - escrow_deadline_margin_blocks blocks after the
+# escrow was paid, mostrod cancels the trade (Active) or opens a dispute
+# (FiatSent) while the escrow still exists. Must comfortably exceed your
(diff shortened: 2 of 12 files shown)


grunch:
Pushed 08ed323 addressing the review findings:

- **Codex P1 (block height vs wall clock)**: the guardian now probes LND each tick — `GetInfo` for the chain tip and `LookupInvoice` for the accepted HTLC's actual `expiry_height` — and acts when `tip >= expiry - margin`. The DB timestamp filter became a candidate prefilter at half the nominal window, so a stretch of fast blocks cannot slip past it. The 10-minute nominal clock survives only as a fallback when LND cannot answer.
- **Codex P1 (invoice minimum treated as actual expiry)**: the timestamp-based "past the hard horizon ⇒ auto-refunded" inference is gone. After a transient cancel failure, state is only fixed on chain evidence: the HTLC verifiably past its expiry height, no accepted HTLC backing the invoice, or LND reporting the invoice already canceled. Otherwise the order stays eligible and is retried.
- **Codex P1 (stuck fiat-sent after partial dispute write)**: an `initiated`/`in-progress` dispute row on a still-`fiat-sent` order is now treated as a half-completed transition — the pass reuses the row and finishes the order flip, notifications and event publication instead of skipping.
- **CodeRabbit (notify both-or-neither)**: parties are notified independently; a missing key logs a warning for that party only, mirroring `flow::hold_invoice_canceled`.
- **CodeRabbit (inclusive boundary test)**: added a row at exactly the cutoff, expected count bumped to 3.

New regression tests cover fast blocks, a live HTLC overriding the nominal clock, transient cancel failures with and without chain evidence, and the resumed dispute transition. 1138 tests green.

AndreaDiazCorreia (approved):
tACK

PR #864, opened 2026-08-12
fix: handle do_payment early-exit failures so buyer is always retried

### LINKED ISSUE #806: [HIGH] Payment failures in release/admin_settle are silently swallowed (seller settled, buyer unpaid)
### Severity: HIGH

Payment failures on release are silently swallowed: the seller is settled but the buyer may never be paid, with no `failed_payment` mark, no retry, and no log.

### Where
- `src/app/release.rs:305` — `let _ = do_payment(...).await;`
- `src/app/admin_settle.rs:255` — `let _ = do_payment(...).await;`

### Problem
On the early-return branch, by the time `do_payment` is invoked the seller's hold invoice has already been **settled** (funds captured). Discarding the result with `let _ =` means:
- a failed/rejected payout to the buyer is not detected,
- `failed_payment` is never set, so the retry/scheduler path (`check_failure_retries`) never picks it up,
- nothing is logged or surfaced to the admin/user.

Net effect: the seller is charged, the buyer is not paid, and the system believes the order completed.

### Suggested fix
- Handle the `do_payment` result: on error, mark `failed_payment = true` / set the appropriate status so the retry scheduler takes over, and log the failure with the order id.
- Emit the buyer/seller Nostr messages that correspond to a payout-in-progress vs payout-failed state.

Arowolokehinde:
## Summary

- `do_payment` is called after the seller's hold invoice is already settled (irreversible). On three early-exit paths — missing buyer invoice, zero amount after fee, and LND unreachable — the function returned `Err` without calling `check_failure_retries_or_log`, leaving `failed_payment = false` in the database.
- Because the retry scheduler (`job_retry_failed_payments`) only picks up orders where `failed_payment = true`, these orders were permanently stranded: seller charged, buyer never paid, no log, no retry.
- Fix: call `check_failure_retries_or_log` on each of the three remaining early exits inside `do_payment` so the retry scheduler always finds the order regardless of which failure path was hit.
- The LNURL resolution and `send_payment` sync-failure paths were already handled by prior commits (#858, #861); this PR closes the three remaining gaps to make `do_payment` fully self-contained on all failure paths.

## Test plan

- [ ] New regression test `do_payment_error_marks_failed_payment_for_retry`: creates an order in `SettledHoldInvoice` with no buyer invoice, calls `do_payment` with `let _ =` (mirroring both call sites), and asserts `failed_payment = true` in the database — proving the retry scheduler will find the order

Closes #806 

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed payments now record retry status when invoices are missing, payment amounts are zero, payout details are invalid, or payment connections fail.
  * Payment destinations and resolved invoices continue to be validated before submission.
  * Settled hold-invoice status is saved before buyer payment begins.
  * Improved reliability when retrying eligible failed payments without overwriting unrelated order updates.

### DIFF
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -57,6 +57,15 @@ pub async fn check_failure_retries(
 
     // Count payment retries up to limit
     order.count_failed_payment(retries_number);
+    // Only update payment-retry fields to avoid overwriting fields modified by
+    // concurrent processes (dev_fee_paid, dev_fee_payment_hash, status, etc.)
+    sqlx::query("UPDATE orders SET failed_payment = ?, payment_attempts = ? WHERE id = ?")
+        .bind(order.failed_payment)
+        .bind(order.payment_attempts)
+        .bind(order.id)
+        .execute(pool)
+        .await
+        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
 
     let buyer_pubkey = order.get_buyer_pubkey().map_err(MostroInternalErr)?;
 
@@ -107,16 +116,6 @@ pub async fn check_failure_retries(
         )
         .await;
     }
-
-    // Only update payment-retry fields to avoid overwriting fields modified by
-    // concurrent processes (dev_fee_paid, dev_fee_payment_hash, status, etc.)
-    sqlx::query("UPDATE orders SET failed_payment = ?, payment_attempts = ? WHERE id = ?")
-        .bind(order.failed_payment)
-        .bind(order.payment_attempts)
-        .bind(order.id)
-        .execute(pool)
-        .await
-        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
 
     Ok(order)
 }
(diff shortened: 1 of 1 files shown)


AndreaDiazCorreia (changes requested):
Nice catch on these three paths. Making `do_payment` self-contained is the right level to fix #806, and I confirmed locally that the missing-invoice path now recovers properly: three attempts, then `AddInvoice` reaches the buyer, and the scheduler stops. A couple of things before this can land, though. `cargo fmt --check` currently fails (trailing whitespace on the `InvoiceInvalidError` return, plus a double blank line after the new test), so CI will be red. More importantly, the `amount == 0` path now loops forever: with the production default of `payment_attempts = 3`, `check_failure_retries` hits the `amount <= 0` guard and returns `Err(InvalidAmount)` before the `UPDATE`, so the counter sticks at 2 in the database and `job_retry_failed_payments` re-runs the order every 60 seconds indefinitely, without ever sending `AddInvoice`. The smallest fix that keeps the "always recoverable" design is to move that `UPDATE` above the retry-exhausted branch. I tried it and all 37 `app::release` tests still pass, including `check_failure_retries_rejects_non_positive_amount`.

On tests and style: `do_payment_fails_without_buyer_invoice`, `do_payment_fails_when_amount_is_consumed_by_fee` and `do_payment_fails_fast_when_lnd_is_unreachable` already cover exactly these three paths, so instead of a fourth test it'd be cleaner to persist the order with `create(&pool)` in those three and add the `failed_payment` assertion to each. As it stands, only one of the three paths you fixed is actually covered. It'd also be good to switch the new `tracing::error!` calls to the `warn!("Order id {}: ...", order.id)` style the rest of the function uses, and to extend the `do_payment` doc comment, which still lists only the old set of paths going through `check_failure_retries_or_log`. Last one, non-blocking: `LndConnector::new` has no connect timeout, so the new bookkeeping there fires on a misconfigured LND but not on a wedged one, which is the case that actually strands orders. Fine to leave for a follow-up, just worth knowing that gap is only half closed.

Arowolokehinde:
All changes addressed -  thanks for the thorough review @AndreaDiazCorreia 

Moved the UPDATE above the retry-exhausted branch in check_failure_retries to fix the amount == 0 infinite loop. All 36 tests still pass including check_failure_retries_rejects_non_positive_amount.

Dropped the standalone fourth test and instead added create(&pool) plus assert!(db_order.failed_payment) to the three existing tests so all three fixed paths are now covered. Switched the new log calls to warn!("Order id {}: ...", order.id) and extended the do_payment doc comment to list the new paths. cargo fmt --check is clean.

On the LndConnector timeout gap — noted.
i would create a follow-up

Catrya (changes requested):
Hi @Arowolokehinde  thanks for your work, good job! but this is already addressed in PR #908 
Will close this PR as duplicated. New contributions are welcome

Catrya:
Hi @Arowolokehinde thanks for your work, good job! but this is already addressed in PR #908
Will close this PR as duplicated. New contributions are welcome.

Issue #856, opened 2026-08-07
find_held_invoices does not resubscribe dispute or fiat-sent orders


grunch:
Follow-up from the review of #853.

`find_held_invoices` (`src/db.rs`) selects two disjoint sets:

- `active` with `invoice_held_at != 0`
- `waiting-payment` / `waiting-buyer-invoice` with a hash

An order that has moved past `active` still has a live hold invoice, but matches neither branch. That covers **`dispute` and `fiat-sent`** — the review only named `dispute`, but `fiat-sent` has exactly the same shape, since an active order reaches it with the escrow still locked.

This is not a regression from #853: the previous query (`invoice_held_at != 0 AND status == 'active'`) excluded them too.

It is inert today because `hold_invoice_settlement` and `hold_invoice_canceled` (`src/flow.rs`) only log. But the doc comment on the `active` branch says it exists "to observe settle/cancel", and the moment either handler does something real — updating order state, releasing a bond, notifying a party — the gap becomes functional: a restart during a dispute would silently stop observing that invoice.

Fix is to extend the post-payment branch to the statuses that can still hold escrow, and to add coverage pinning which statuses are resubscribed.

PR #868, opened 2026-08-12
Enforce the escrow CLTV deadline on live trades

AndreaDiazCorreia:
## Motivation

A trade is escrowed in a hold invoice, and that escrow has a hard lifetime. The invoice is created with `hold_invoice_cltv_delta` (144 blocks by default), and LND force-cancels an accepted hold HTLC a few blocks before its expiry height — `invoices.holdexpirydelta`, 24 by default — so a channel is never forced to close over a held payment. The refund goes to the seller regardless of what the trade was doing at that moment. On the defaults that leaves roughly 20 hours of escrow from the moment the seller pays.

Nothing in the order state machine knew about that horizon:

- No scheduler job covered `active`, `fiat-sent` or `dispute`. `job_cancel_orders` only sweeps the waiting states, and `job_expire_pending_older_orders` only the pending ones.
- `hold_invoice_canceled` logged the event and returned — no status change, no notification, no dispute.

So a trade could outlive its escrow. The order stayed `active` or `fiat-sent` and kept advertising itself as a live trade with nothing backing it, and the seller's release then failed at settle, after the buyer may already have acted on the trade looking live. Slow fiat rails make that window easy to reach without anyone misbehaving.

## Summary by CodeRabbit

* **New Features**
  * Added automatic monitoring of Lightning escrow deadlines.
  * Active trades nearing expiry are canceled safely; “fiat-sent” trades enter dispute handling, with operator alerts for urgent disputed cases.
  * Escrow cancellation now notifies both parties and safely resolves related records.
  * Held invoices are restored across active, fiat-sent, and dispute states after restart.

* **Configuration**
  * Added configurable warning and safety windows, with startup validation to prevent unsafe settings.

* **Documentation**
  * Added an operational runbook covering expiry handling, alerts, retries, and tuning.

### DIFF
--- a/src/app/dispute.rs
+++ b/src/app/dispute.rs
@@ -29,6 +29,21 @@ async fn publish_dispute_event(
         false => "seller",
     };
 
+    publish_dispute_event_with_initiator(ctx, dispute, my_keys, initiator).await
+}
+
+/// Same event, for a dispute no trader opened.
+///
+/// `initiator` is a free string because mostro can open a dispute itself when
+/// a trade is about to lose its escrow (`mostro`), and claiming the buyer or
+/// the seller filed it would be wrong in the event, in the order's dispute
+/// flags, and in the solver's view of who is asking for what.
+async fn publish_dispute_event_with_initiator(
+    ctx: &AppContext,
+    dispute: &Dispute,
+    my_keys: &Keys,
+    initiator: &str,
+) -> Result<(), MostroError> {
     // Create tags for the dispute event
     let tags = Tags::from_list(vec![
         // Status tag - indicates the current state of the dispute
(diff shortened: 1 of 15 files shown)


src/flow.rs
```diff
@@ -184,21 +187,119 @@ pub async fn hold_invoice_settlement(hash: &str, pool: &SqlitePool) -> Result<()
...
+    let order_updated = crate::util::update_order_event(ctx.keys(), Status::Canceled, &order)
+        .await
+        .map_err(|e| MostroInternalErr(ServiceError::NostrError(e.to_string())))?;
+
+    // Publishing before the claim mirrors `release_action`, and is safe here
+    // for a narrower reason: the only other writer that can win this race is
+    // the scheduler's deadline cancel, which publishes the very same
+    // `Canceled` status. A settled invoice never emits a cancel, so a release
+    // cannot be overwritten by the event above.
+    if !crate::db::claim_escrow_order_canceled(pool, order.id, &order_updated.event_id).await? {
```
grunch, line 246: _🗄️ Data Integrity & Integration_ | _🟠 Major_

**This safety argument doesn't hold: mostro's own cancel flows emit this event *before* they persist their terminal status, and the subscriber can win the claim against them.**

The comment's premise — "whoever asked already moved the order to its terminal status" / "the only other writer that can win this race is the scheduler's deadline cancel" — is not true for two existing flows:

- `admin_cancel_action` calls `ln_client.cancel_hold_invoice(hash)` at `src/app/admin_cancel.rs:120`, then does dispute updates and Nostr publishes, and only persists `CanceledByAdmin` at lines 168–172.
- Cooperative cancel calls `cancel_hold_invoice` at `src/app/cancel.rs:98` and persists `CooperativelyCanceled` afterwards (lines 108–112).

LND pushes `InvoiceState::Canceled` onto the subscription stream as soon as the RPC lands, so this function runs concurrently with the tail of those handlers. The order row still reads `dispute`/`active` at that point, so it passes the live-status guard, and then:

1. `claim_escrow_order_canceled` **succeeds** (the status is still one of the three claimable ones), stamping the order `canceled` instead of `canceled-by-admin`/`cooperatively-canceled`.
2. Both traders get an `Action::Canceled` DM on top of the admin/co-op flow's own messages.
3. `release_taker_bonds_for_order_or_warn` runs — racing `apply_bond_resolution` in `admin_cancel_action` (`src/app/admin_cancel.rs:209-231`). If the subscriber gets there first, a bond the solver explicitly directed to **slash** is released to the trader instead. That is an irreversible money outcome, not just bookkeeping.
4. Even when the claim *loses*, the `Canceled` replaceable event above has already been published, and can overwrite the admin's `canceled-by-admin` event on relays depending on timestamp ordering.

Two complementary fixes, either of which closes the money-outcome race:

- Persist the terminal status (or take a claim) **before** `cancel_hold_invoice` in `admin_cancel_action` and the cooperative-cancel paths, so the subscriber's status guard sees a resolved order; and/or
- Have this function distinguish cancels mostro itself requested — e.g. record the hash in a short-lived "we canceled this" set before every internal `cancel_hold_invoice`, and treat a hit as the confirmation case. At minimum, move the Nostr publish to *after* the claim succeeds so a lost race publishes nothing.

The redelivery test covers the sequential case; this race is invisible to it because both writers run in the same process within milliseconds of a real admin cancel.

src/scheduler.rs
```diff
@@ -596,6 +600,116 @@ async fn job_cancel_orders(ctx: AppContext) {
...
+/// minimum the payer had to honour, so the horizon cannot be derived from the
+/// order row. [`escrow_deadline_action`] turns the distance into a decision.
+///
+/// Failures are always "skip and retry next tick": one unreachable node or one
+/// unreadable invoice must not end trades, and there is a whole warning window
+/// worth of ticks before anything is actually due.
+async fn job_enforce_escrow_deadline(ctx: AppContext) {
+    let mut ln_client = match LndConnector::new().await {
+        Ok(client) => client,
+        Err(e) => return error!("escrow_deadline: failed to create LND client: {e}"),
```
grunch, line 622: _🩺 Stability & Availability_ | _🟠 Major_

**A transient LND outage at boot permanently disables the escrow-deadline safety net for the whole run.**

`LndConnector::new()` failing here returns from `job_enforce_escrow_deadline` before the loop is ever spawned — one `error!` line and the deadline enforcement this PR exists to add is silently off until the daemon restarts. A node that boots while LND is still starting (a common systemd ordering situation) runs every trade without the sweep, which is exactly the failure mode the PR description calls out as dangerous.

`job_cancel_orders` shares this weakness, but this job is the backstop for irreversible escrow loss, and the repo already has the right pattern: `job_process_bond_payouts` retries `LndConnector::new()` with capped exponential backoff inside the spawned task (`src/scheduler.rs:864-877`). Moving the construction inside the spawn with that same retry loop makes the sweep self-healing:

```rust
tokio::spawn(async move {
    let mut backoff_secs = 5u64;
    let mut ln_client = loop {
        match LndConnector::new().await {
            Ok(client) => break client,
            Err(e) => {
                warn!("escrow_deadline: LndConnector::new failed: {e} — retrying in {backoff_secs}s");
                tokio::time::sleep(tokio::time::Duration::from_secs(backoff_secs)).await;
                backoff_secs = (backoff_secs * 2).min(300);
            }
        }
    };
    // ... existing loop
});
```

src/app/dispute.rs
```diff
@@ -237,6 +252,91 @@ pub async fn dispute_action(
...
+/// `mostro`, because recording the buyer or the seller as the filer would
+/// misrepresent the dispute to the solver — and to the reputation and
+/// restore-session views that read those flags.
+///
+/// Errors instead of unwinding on failure: the order stays in `fiat-sent` and
+/// the sweep retries on its next tick, which is what we want while the escrow
+/// is still there.
+pub async fn open_deadline_dispute(ctx: &AppContext, order: &Order) -> Result<(), MostroError> {
+    let pool = ctx.pool();
+    if find_dispute_by_order_id(pool, order.id).await.is_ok() {
```
grunch, line 274: _🧹 Nitpick_ | _🟡 Minor_

**`is_ok()` conflates "no dispute exists" with a failed query.**

`find_dispute_by_order_id` maps *every* sqlx error — `RowNotFound` and a transient DB failure alike — into `DbAccessError`, so `.is_ok()` here reads a transient failure as "no dispute, proceed". The function then flips the order to `Dispute` and only trips on `dispute.create(pool)` (or worse, succeeds and files a second dispute if there is no unique index on `disputes.order_id`).

The same pattern exists in `dispute_action` above, so this is consistent with the codebase — but this call site is new, runs unattended on a scheduler tick, and its whole retry story depends on failures being classified correctly. Distinguishing `RowNotFound` from other errors (propagating the latter so the tick retries cleanly) would make the duplicate guard actually mean what it says.

grunch (changes requested):
**Actionable comments posted: 3**

Strict review of the escrow CLTV deadline enforcement. I read the existing bot reviews first and none of the findings below repeat them — the atomicity of the order/dispute writes, the mostro-initiated dispute being unusable by the solver flows, the unbounded `alerted` set and the context construction order are already reported and are not re-raised here.

The overall design is solid: the pure policy function, the deliberate asymmetry between `active`/`fiat-sent`/`dispute`, the conditional claim for the cancel race, and the test coverage (policy table, DB claims, subscriber no-op on redelivery) are all well done. The `find_held_invoices` fix for #856 is correct and well tested.

What blocks approval is a race the new `hold_invoice_canceled` body introduces against mostro's *own* cancel flows (inline comment in `src/flow.rs`) — it can double-notify traders, mislabel the terminal status on relays, and release bonds a solver directed to slash. The other two comments are an availability gap in the new scheduler job and a minor error-conflation issue.

<details>
<summary>🔍 Verification performed</summary>

- Read both existing bot reviews (Codex, CodeRabbit) and all inline comments to avoid duplicates.
- Traced `admin_cancel_action` (`src/app/admin_cancel.rs:120` → `168-172`) and cooperative cancel (`src/app/cancel.rs:98` → `108-112`) to confirm the invoice is canceled **before** the terminal status is persisted in both flows.
- Confirmed `apply_bond_resolution` (slash included) runs *after* `cancel_hold_invoice` in `admin_cancel_action` (`src/app/admin_cancel.rs:209-231`), so the subscriber's `release_taker_bonds_for_order_or_warn` genuinely races a pending slash directive.
- Compared the new job's LND bootstrap with `job_process_bond_payouts`, which already retries `LndConnector::new()` with capped exponential backoff (`src/scheduler.rs:864-877`).

</details>

AndreaDiazCorreia:
This covers the same finding as #862, which came first and is further along. I missed it when I picked this up, let's merge #862.

Thanks for the review; the `flow.rs` race against `admin_cancel_action` and the cooperative-cancel path is a real one.

I'll close this and send what #862 does not cover as separate PRs:

- A separate, earlier warning window for `fiat-sent`, so a solver has time to act on the dispute before the escrow goes (#862 acts on `active` and `fiat-sent` at the same point).
- An operator alert for `dispute` orders approaching the horizon, deduplicated per order and escalating inside the safety window.
- The `find_held_invoices` fix for #856, which is independent of the CLTV work.
- The operations runbook (`docs/LIGHTNING_OPS.md`).

PR #874, opened 2026-08-13
fix: keep the Nostr inbox alive across relay closures

AndreaDiazCorreia:
## Problem

Mostro receives every user action — `TakeSell`, `AddInvoice`, `FiatSent`, `Release`, `Dispute` — over a single Nostr subscription opened once at startup.

Relays can end a subscription at any time by sending `CLOSED`, which is normal protocol behaviour: rate limiting, a policy change, a restart, a relay that starts requiring NIP-42 authentication. nostr-sdk reacts by removing the subscription for almost every reason prefix (and for no prefix at all), and a removed subscription is never re-REQ'd, not even after a reconnect.

The event loop matched only `ClientNotification::Event`, discarding the entire relay control plane. So a single `CLOSED` from a single relay left the daemon running and connected while receiving nothing at all — with no log to show for it, since the SDK reports it at `debug` and release builds filter that out (`RUST_LOG=none,mostro=info`).

Meanwhile the scheduler kept working normally: `job_cancel_orders` went on cancelling hold invoices and, where anti-abuse bonds are enabled, slashing them, for orders whose messages the daemon was no longer in a position to receive.

## What this PR does

Six focused commits:

1. **Stable inbox identity** — the subscription gets a fixed id and lives in `src/inbox.rs`, so a `CLOSED` frame can be attributed to it. Also logs the per-relay outcome of the REQ, which was previously discarded whole.
2. **Recover on `CLOSED`** — the event loop consumes the full notification stream; a `CLOSED` naming the inbox is logged and the REQ re-sent to that relay, under per-relay backoff (immediate first retry, doubling to a five-minute ceiling, cleared on `EOSE`). Also handles `Shutdown`, which used to spin the outer loop on an empty stream at 100% CPU.
3. **NIP-42** — the daemon and price clients authenticate with the node's key, so relays that gate reads behind AUTH become usable instead of silently  undeliverable. The AUTH event is bound to the relay's challenge and URL, and the node already publishes events signed with this key to these relays.
4. **Watchdog** — every 30s, audit each connected read relay and re-subscribe any that stopped serving the inbox. Covers the losses that produce no frame the loop can observe: a notification channel that dropped messages under lag, a REQ that failed to go out, a relay added after startup.
5. **Timeouts stand down while the inbox is down** — a timeout means "the user did not answer in time", which only holds while Mostro can hear. Skip the tick entirely while blind, and give the downtime back afterwards.
6. **Docs**.

## Design notes worth reviewing

**Messages lost while blind are not recovered.** `accept_event` rejects anything whose `created_at` is older than ten seconds, so a message sent into a dead inbox is already too old by the time the subscription returns — re-subscribing with `since` instead of `limit(0)` would not help. That is what makes commit 5 necessary rather than cosmetic: the user has to send again, so punishing them for the silence would be punishing them for the node's failure.

**Health is judged by the subscription, not by traffic.** An instance with no trades in flight is legitimately silent, so treating quiet as failure would raise false alarms on an idle node and stop its timeout machinery for no reason.

**A relay re-subscribed during an audit does not count as listening until the next round.** Sending a REQ says nothing about whether the relay will honour it; counting the attempt would report a healthy inbox indefinitely against a relay that closes it on principle. Costs one extra interval before the inbox is declared healthy, and errs toward keeping the timeout clock frozen slightly
longer than needed.
**The timeout debt over-compensates slightly.** Grace reaches every waiting order, not only those in flight during the outage, so an order taken *during* the blind window gets a little more than it lost. Deliberate: the error delays a cancellation, where the opposite would cancel an escrow the node never had grounds to cancel.

## Testing

19 new tests. `cargo fmt` applied and `cargo clippy --all-targets
--all-features` is clean.

    test result: ok. 1157 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
    test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out

Both behavioural regression tests were verified to fail with the fix disabled: the recovery test times out with `the inbox never recovered from the relay's CLOSED`, and the NIP-42 test fails on the subscription being dropped. Two tests pin current SDK behaviour (a CLOSED subscription is dropped; an auth-required one is dropped without an authenticator) so a future nostr-sdk bump that changes it is noticed rather than silently making this module moot.

## No config or schema changes

No new settings keys, no migrations, no protocol or tag changes. Relay backoff and the watchdog interval are documented constants, matching how the other maintenance intervals in `scheduler.rs` are handled.

## Unrelated change, flagged

The sequence diagram in `docs/EVENT_ROUTING.md` did not render: the participant named `Loop` collides with mermaid's `loop` keyword. Fixed while editing that file. Happy to split it out if preferred.

## Follow-ups, not in this PR

- `job_expire_pending_older_orders` still runs while the inbox is down. Its damage profile is different (no escrow or bond at stake, only an untaken order being cancelled) but a `TakeSell` that never arrived still expires an order unfairly.
- Broadcast lag is not observable from application code: nostr-sdk 0.45.1's notification stream swallows `Lagged` internally and exposes no raw receiver. The watchdog mitigates the consequence — a `CLOSED` missed due to lag is caught within 30s — but trade messages dropped to lag leave no trace.

## Summary by CodeRabbit

* **New Features**
  * Added NIP-42 authentication support for relays that require it.
  * Added automatic inbox subscription recovery and a 30-second health watchdog.
* **Bug Fixes**
  * Paused order timeouts while inbox connectivity is unconfirmed and compensated for downtime after recovery.
  * Added blameless order unwinding after outages lasting up to three hours, releasing bonds without slashing.
* **Documentation**
  * Documented inbox routing, authentication, recovery, health monitoring, missed-message handling, and outage behavior.


src/inbox/mod.rs
```diff
@@ -0,0 +1,1550 @@
...
+        }
+        let registered = relay.subscription(subscription.id()).await.is_some();
+        // Without a health record there is nowhere to have stored an
+        // acknowledgement, so fall back to registration alone.
+        let acknowledged = health
+            .as_ref()
+            .map(|h| h.has_acknowledged(url))
+            .unwrap_or(true);
+
+        if registered && acknowledged {
```
grunch, line 427: _🎯 Functional Correctness_ | _🟠 Major_

**A relay's `EOSE` credit outlives the websocket session it was earned in.**

`acknowledged` is keyed only by `RelayUrl`, so the entry set by the keeper on the first `EOSE` stays valid across a TCP drop and reconnect. On reconnect the SDK re-sends the REQ on its own — `post_connection` → `resubscribe()` → `should_resubscribe()` returns `true` when `stats.connected_at() > subscribed_at && stats.success() > 1` — and nothing in this PR invalidates the ack at that point: there is no `ClientNotification` for relay status in 0.45.1, the watchdog `continue`s past disconnected relays without touching the set, and a relay that silently ignores the replacement REQ sends no frame the keeper could react to.

Sequence:
1. Relay answers the initial REQ → `acknowledged` contains it.
2. Connection drops and comes back; the SDK re-REQs under the same id (`registered` stays `true`).
3. Relay holds the socket open and ignores the REQ (or only serves it after an AUTH that never completes). No `EOSE`, no `CLOSED`.
4. Next audit: `registered && acknowledged` → `Listening` → `job_cancel_orders` resumes slashing while the node is deaf on that relay.

This is the same class of bug as the stale-ack-after-`CLOSED` blocker, reachable through reconnect instead of `CLOSED`. The module docs explicitly promise the opposite ("a relay that keeps the connection open and quietly drops the REQ still appears subscribed there — and the daemon would resume timeouts while deaf").

<details>
<summary>🛠️ Proposed fix — bind the ack to the connection it was received on</summary>

`Relay::stats().connected_at()` is already a `Timestamp` on the relay the audit holds. Store when the ack was recorded and only trust it if it post-dates the current connection:

```diff
-    acknowledged: HashSet<RelayUrl>,
+    /// Relay → wall-clock second the last `EOSE` for the inbox was seen.
+    acknowledged: HashMap<RelayUrl, i64>,
```

```diff
     pub fn note_relay_acknowledged(&self, relay: &RelayUrl) {
-        ...acknowledged.insert(relay.clone());
+        ...acknowledged.insert(relay.clone(), now_secs());
     }

-    pub fn has_acknowledged(&self, relay: &RelayUrl) -> bool {
-        ...acknowledged.contains(relay)
+    /// Whether `relay` answered the inbox REQ *on its current connection*.
+    pub fn has_acknowledged_since(&self, relay: &RelayUrl, connected_at: i64) -> bool {
+        ...acknowledged.get(relay).is_some_and(|&at| at >= connected_at)
     }
```

```diff
         let registered = relay.subscription(subscription.id()).await.is_some();
+        let connected_at = relay.stats().connected_at().as_secs() as i64;
         let acknowledged = health
             .as_ref()
-            .map(|h| h.has_acknowledged(url))
+            .map(|h| h.has_acknowledged_since(url, connected_at))
             .unwrap_or(true);
```

Plus a regression test in the spirit of `a_closed_frame_invalidates_the_relays_earlier_acknowledgement`: seed an ack, bump the relay's connection (or simply construct the check with a `connected_at` later than the ack), and assert the audit does not count it. `>=` rather than `>` keeps an `EOSE` that lands in the same second as the connect.

</details>

src/inbox/mod.rs
```diff
@@ -0,0 +1,1550 @@
...
+    /// backoff. Everything else (`OK`, `NOTICE`, other subscriptions' frames)
+    /// is not this module's business.
+    pub async fn on_relay_message(
+        &mut self,
+        client: &Client,
+        relay_url: &RelayUrl,
+        message: &RelayMessage<'_>,
+    ) {
+        match message {
+            RelayMessage::Closed {
```
grunch, line 197: _🩺 Stability & Availability_ | _🟡 Minor_

**`auth-required:` closures are now handled twice.**

With the NIP-42 authenticator from this PR, the SDK does *not* remove the subscription on `CLOSED "auth-required: …"` — `handle_relay_message` maps it to `HandleClosedMsg::MarkAsClosed`, and the `ingester` task re-sends every marked subscription once the AUTH round-trip succeeds. The keeper reacts to the very same frame: it `unsubscribe`s (sends `CLOSE`, drops the entry the SDK was about to re-REQ), re-sends the REQ immediately — typically before the AUTH `OK` has come back, so the relay answers `auth-required` again — and arms the 2s backoff. It converges, but costs an extra REQ/CLOSED round-trip per auth-gated relay per connection and logs a spurious `warn!` + backoff for a relay that is behaving correctly. The same applies to `rate-limited:`, which the SDK also only marks.

<details>
<summary>💡 Suggested shape</summary>

Let the SDK own the prefixes it already recovers from, and keep the keeper for the ones it discards:

```rust
RelayMessage::Closed { subscription_id, message }
    if subscription_id.as_ref() == self.subscription.id() =>
{
    match MachineReadablePrefix::parse(message) {
        // The SDK marks these and re-REQs after AUTH / on reconnect.
        Some(MachineReadablePrefix::AuthRequired) | Some(MachineReadablePrefix::RateLimited) => {
            info!("Relay {relay_url} closed the inbox provisionally (\"{message}\"); leaving recovery to the SDK");
        }
        _ => {
            warn!("Relay {relay_url} closed the Mostro inbox subscription: \"{message}\"");
            self.resubscribe(client, relay_url).await;
        }
    }
}
```

The watchdog still covers the case where the SDK's own re-REQ never lands (the subscription is marked `closed`, not acknowledged, so the audit re-sends it after 30s). If you'd rather keep the keeper unconditional, a one-line comment acknowledging the double handling would do.

</details>

src/inbox.rs
```diff
@@ -0,0 +1,1550 @@
...
+
+        if registered && acknowledged {
+            listening += 1;
+        } else {
+            // Not serving it: a CLOSED the event loop never saw (the
+            // notification channel drops frames when it lags), a REQ that
+            // failed to go out, a relay re-added after startup — or one that
+            // took the REQ and never answered it.
+            warn!("Relay {url} is connected but not serving the Mostro inbox; re-subscribing");
+            resubscribe_relay(relay, subscription, health.as_deref()).await;
```
grunch: _🩺 Stability & Availability_ | _🟡 Minor_

**The watchdog re-REQs a refusing relay every 30s, which makes the keeper's 5-minute backoff ceiling mostly cosmetic.**

`RESUBSCRIBE_MAX_BACKOFF` is documented as the rate at which "a relay that has refused the inbox for five minutes straight" is retried "without generating traffic that looks like an attack", and `docs/EVENT_ROUTING.md` repeats the five-minute figure. But `check_inbox_health_with` calls `resubscribe_relay` unconditionally for every connected relay that is not `registered && acknowledged`, every `INBOX_WATCHDOG_INTERVAL`. A relay that answers every REQ with `CLOSED` (pubkey allowlist, ban) therefore gets a REQ every 30s forever — the keeper's backoff only governs the *second* REQ between two audits.

Not a correctness problem, but either the backoff should be shared (e.g. move `allow_attempt` into `InboxHealth` so the audit consults it too), or the docs/constant comments should say that the effective floor is the watchdog interval. `watchdog_stays_blind_against_a_relay_that_keeps_closing` would be the natural place to pin whichever behaviour you pick.

src/scheduler.rs
```diff
@@ -54,10 +55,53 @@ pub async fn start_scheduler(ctx: AppContext) {
...
+/// all. This job is the backstop, and the only thing that notices when *every*
+/// relay has gone quiet.
+async fn job_inbox_watchdog(ctx: AppContext) {
+    #[allow(deprecated)]
+    let event_kind = ctx.settings().mostro.transport.event_kind();
+    let subscription = crate::inbox::InboxSubscription::new(ctx.keys().public_key(), event_kind);
+
+    tokio::spawn(async move {
+        loop {
+            // Sleep first: at startup `main` has just subscribed, and a REQ
```
grunch: _📝 Documentation_ | _🔵 Nitpick_

Stale since b6dd1a5: `main` no longer sends the inbox REQ — `app::run` / `app::run_cashu` do, after taking the notification stream (and `main.rs:124-128` now says exactly that). Suggest:

```diff
-            // Sleep first: at startup `main` has just subscribed, and a REQ
+            // Sleep first: at startup the event loop has just subscribed, and a REQ
             // still in flight would look exactly like a missing one.
```

docs/EVENT_ROUTING.md
```diff
@@ -6,6 +6,37 @@ How Nostr events become actions and side effects.
 - Source: `src/app.rs:run`
 - Steps: POW check → signature verify → recency guard → NIP-59 unwrap → parse `mostro_core::Message` → inner verify → `check_trade_index` → dispatch.
 
+## The Inbox Subscription
+- Source: `src/inbox.rs`
+- Every trade message reaches Mostro over a single long-lived subscription, created at startup by `main` with a stable id (`InboxSubscription`) so that later frames can be attributed to it. Its filter is p-tagged to the node, restricted to the configured transport's event kind, and carries `limit(0)`: only live traffic is wanted.
```
grunch: _📝 Documentation_ | _🔵 Nitpick_

Same drift as the scheduler comment: the subscription is "created at startup by `main`" here, but it is sent from `app::run` / `app::run_cashu` once the notification stream exists (b6dd1a5), and the reason it moved — a receiver created after the REQ never sees the relay's `EOSE` — is worth one sentence here since the rest of this section depends on that `EOSE` being observed.

```diff
-- Every trade message reaches Mostro over a single long-lived subscription, created at startup by `main` with a stable id (`InboxSubscription`) so that later frames can be attributed to it.
+- Every trade message reaches Mostro over a single long-lived subscription, sent by the event loop (`app::run` / `app::run_cashu`) right after it takes its notification stream — a receiver created after the REQ would miss the relay's `EOSE` — with a stable id (`InboxSubscription`) so that later frames can be attributed to it.
```

grunch (changes requested):
**Actionable comments posted: 5**

> [!NOTE]
> Reviewed head `379e819` against nostr-sdk 0.45.1 sources (`relay/inner.rs`: `should_resubscribe`, `post_connection`, `handle_relay_message` CLOSED arm, `ingester`). Findings already raised by Codex / CodeRabbit / ermeme (stale ack after `CLOSED`, blind-at-startup, unbounded pause, decaying grace) are fixed on this head and are not repeated here.

<details>
<summary>🧭 Review summary</summary>

**What is solid**
- `InboxKeeper` + `check_inbox_health` close the "one `CLOSED` = permanently deaf node" hole, and the SDK-pinning tests (`without_the_keeper_a_closed_inbox_stays_dead`, `auth_gated_relay_keeps_the_subscription_only_when_authenticating`) will flag a future nostr-sdk bump that changes the contract.
- Subscribing from inside `run` after `client.notifications()` exists is correct — `NotificationStream` wraps a `BroadcastStream`, so anything emitted before the receiver is created is unobservable.
- Per-order downtime credit (`blind_seconds_since(taken_at)`) + over-selecting query + exact per-row check is the right shape, and the 3h blameless bound keeps hold invoices from sitting until CLTV.
- `Shutdown` handling and the paced re-attach fix the 100% CPU spin on an ended stream.

**What blocks**
- 🟠 The acknowledgement (`EOSE`) credit survives a websocket reconnect. The SDK silently re-REQs on reconnect (`should_resubscribe`: `connected_at > subscribed_at && success > 1`), so a relay that reconnects and quietly ignores the replacement REQ is still counted as listening — the exact failure mode the ack set was added to exclude (#874, ermeme's blocker), reachable through a different path. Inline comment with a fix sketch on `src/inbox.rs`.

**Non-blocking**
- Keeper and SDK both react to `CLOSED "auth-required:"` when an authenticator is configured (double REQ, spurious backoff).
- The watchdog re-REQs a refusing relay every 30s regardless of the keeper's 5-minute backoff ceiling the docs advertise.
- Two comments/docs still say `main` sends the inbox REQ; it moved to `app::run` in b6dd1a5.
- No test exercises the scheduler gate itself (`is_confirmed_listening` → skip / `unconfirmed_for_secs` ≥ 3h → blameless). `InboxHealth` is well covered; the branch in `job_cancel_orders` that decides between slash and blameless release is not. Not blocking given how that loop is structured, but worth a follow-up that extracts the decision into a pure function.

</details>

<details>
<summary>✅ Verified locally</summary>

- `gh pr checks 874`: build / test / clippy / fmt / MSRV / markdownlint all green on `379e819`.
- SDK behaviour cited above read from `~/.cargo/registry/src/*/nostr-sdk-0.45.1/src/relay/inner.rs` and `src/stream.rs`.

</details>

src/inbox/mod.rs
```diff
@@ -0,0 +1,1934 @@
...
+                subscription_id,
+                message,
+            } if subscription_id.as_ref() == self.subscription.id() => {
+                if is_provisional_closure(message) {
```
grunch, line 201: _🩺 Stability & Availability_ | _🔴 Critical_ | **Blocking**

**A provisional `CLOSED` keeps the relay's stale acknowledgement, so the watchdog never re-sends the REQ.**

The comment three lines below states the recovery contract for this branch:

> the relay is not acknowledged, so the next audit re-sends the REQ

That is not what happens. This branch returns before touching `InboxHealth`, and `note_relay_resubscribed` (the only thing that clears an acknowledgement) is never reached. The test at `a_provisional_closure_is_left_to_the_sdk` asserts the opposite of the comment on purpose:

```rust
assert!(acked(&health, &relay), "the subscription is still registered and still answered, so the credit stands");
```

Now combine that with what the SDK actually does on `MarkAsClosed` (`relay/inner.rs`, `subscription_closed`): it only sets `data.closed = true`. The entry stays in the subscription map, so `relay.subscription(id)` still returns `Some` and `check_inbox_health` computes `registered == true`. With the acknowledgement also intact, `registered && acknowledged` is true and the relay is counted as **listening** — permanently.

`Relay::resubscribe()` has exactly two callers in 0.45.1: `post_connection` and the ingester's post-AUTH success path. Neither fires here:

- **`rate-limited` on a connection that never drops.** The SDK carries the TODO in-tree: `// TODO: add something like MarkAsRateLimited? // TODO: And retry after some time to re-subscribe`. There is no timer. The REQ is re-sent only on the next websocket reconnect, which may never come.
- **`auth-required` with an authenticator whose AUTH is then rejected** (relay allowlist, pubkey ban, policy flip after the relay had already served the REQ and EOSE'd). The ingester emits `AuthenticationFailed` and returns *without* calling `resubscribe()`.

In both cases the node is deaf on that relay for the life of the connection, the watchdog reports the inbox healthy, `is_confirmed_listening()` stays true, and `job_cancel_orders` keeps cancelling escrows and slashing bonds over messages it cannot receive. That is precisely the failure mode this PR was opened to eliminate, reintroduced through a narrower door.

Standing down on the re-REQ is still the right call — racing the SDK's AUTH round-trip is worse. What must not be stood down on is the *health verdict*. Invalidate the credit and let the audit decide:

```diff
                 if is_provisional_closure(message) {
+                    // The REQ is the SDK's to re-send, but until a relay
+                    // answers the replacement it is not serving the inbox —
+                    // and `MarkAsClosed` leaves it looking registered. Drop
+                    // the credit so the audit judges it on the new evidence.
+                    if let Some(health) = &self.health {
+                        health.note_relay_resubscribed(relay_url);
+                    }
                     info!(...);
                     return;
                 }
```

With that, the happy path is unchanged (AUTH succeeds → SDK re-REQs → `EOSE` → credit restored, one audit interval later at worst), and the two failure paths degrade into a re-subscribe under the shared backoff instead of silent deafness. Please also flip the assertion in `a_provisional_closure_is_left_to_the_sdk`, and add a case covering "provisional closure, SDK never re-REQs" → audit reports `Blind`.

The name `note_relay_resubscribed` no longer fits both uses; `note_relay_unacknowledged` (with the resubscribe path calling it) would read better.

src/db.rs
```diff
@@ -533,10 +534,23 @@ pub async fn find_order_by_date(pool: &SqlitePool) -> Result<Vec<Order>, MostroE
...
+/// waiting, which this query cannot express, so the caller applies the exact
+/// per-order figure to the rows returned.
+pub async fn find_order_by_seconds(
+    pool: &SqlitePool,
+    grace: Duration,
+) -> Result<Vec<Order>, MostroError> {
     let mostro_settings = Settings::get_mostro();
     let exp_seconds = mostro_settings.expiration_seconds as u64;
-    let expire_time = Timestamp::now() - exp_seconds;
+    let expire_time = Timestamp::now() - exp_seconds - grace.as_secs();
```
grunch: _🎯 Functional Correctness_ | _🟠 Major_ | **Blocking**

**`grace` narrows the selection instead of widening it, which makes the per-order credit unreachable.**

The doc comment directly above says:

> `grace` widens the window ... so that no order the caller may still have to spare is filtered out here. It deliberately **over-selects**

Subtracting `grace` moves `expire_time` further into the past, and the predicate is `taken_at < ?1`. So an order is selected iff `waited > exp_seconds + grace` — strictly **fewer** rows, not more. The test added in this PR encodes the narrowing behaviour and reads correctly (`grace only postpones the deadline`); it is the two callers' documentation that is inverted.

The consequence is in `scheduler.rs`. With `G = max_blind_seconds()` and `owed = blind_seconds_since(order.taken_at)`, `owed <= G` holds by construction (the overlap of a subset of windows with one interval cannot exceed the sum of all window durations). Every row that survives the query already satisfies `waited > exp_seconds + G >= exp_seconds + owed`, so

```rust
if waited < exp_seconds as i64 + owed {   // scheduler.rs:625
```

can never be true. **The entire per-order compensation path is dead code**, and the behaviour that ships is the single global allowance the design notes explicitly reject:

> A single global allowance cannot do this: it would either under-credit an order that waited through the whole outage or hand the same credit to one taken long after it ended.

Worked example — `exp_seconds = 900`, one outage `[0, 300]`:

| order | taken_at | owed | should cancel at | actually cancels at |
|---|---|---|---|---|
| A (waited through it) | 0 | 300 | 1200 | 1200 ✅ |
| B (taken after recovery) | 400 | 0 | 1300 | 1600 ❌ (+300s) |

And `G` is the sum over `BLIND_WINDOW_RETENTION_SECS` (7 days), capped only by `MAX_BLIND_WINDOWS = 512`. A node with flapping relays that accumulates, say, two hours of blind time over a week extends **every** order's deadline by two hours — hold invoices encumbered that much longer, counterparties waiting that much longer — and it only decays as windows age out.

The fix is to let the query do what the comment claims. Since the per-order check only ever spares, the correct over-selecting query is simply the original one:

```diff
-    let expire_time = Timestamp::now() - exp_seconds - grace.as_secs();
+    // Caller applies the exact per-order credit to these rows; widening here
+    // would be wrong in the other direction (see job_cancel_orders).
+    let expire_time = Timestamp::now() - exp_seconds;
```

which makes the `grace` parameter unnecessary — drop it and revert the call site, keeping the per-order `continue` in the scheduler as the only place downtime is credited. Please add a regression test asserting the table above: an order taken *after* the outage window closed is cancelled at its nominal deadline, not `deadline + max_blind_seconds`.

src/scheduler.rs
```diff
@@ -584,39 +718,50 @@ async fn job_cancel_orders(ctx: AppContext) {
...
-                                );
-                                continue;
+                        //
+                        // Past the pause bound (`blameless`) none of that
+                        // applies: the node has been unable to hear for hours,
+                        // so the timeout says nothing about the user. The order
+                        // is still unwound — otherwise the escrow sits until
+                        // CLTV expiry — but every bond is released rather than
+                        // settled.
+                        if blameless {
```
grunch, line 795: _🎯 Functional Correctness_ | _🟠 Major_

**The blameless path drops a failed bond release on the floor, then persists the order out of its own retry window.**

`release_on_timeout_without_slashing` returns `()`. It funnels into `release_bonds_for_order_or_warn` / `release_taker_bonds_for_order_or_warn`, whose documented contract is to swallow the error and `warn!` — `release_bonds_for_order_or_warn_swallows_db_error` pins exactly that. So a DB failure here is invisible to this call site, execution falls through to cancel/republish, and the order leaves `waiting-*`.

That is the failure the `else` branch four lines down goes out of its way to avoid, in a comment written by this same codebase:

> Falling through to cancel/republish would persist the order out of `find_order_by_seconds`'s waiting-state eligibility window, and the next tick would never re-evaluate it

The `blameless` branch has the identical exposure with none of the protection: the bond stays `Locked` forever, with no tick that will ever look at it again. It also fires at the worst possible moment — this branch runs only after three hours of blindness, on *every* waiting order in one pass, which is precisely when a DB under stress is most likely to fail.

Give it the same treatment as its sibling:

```rust
if blameless {
    if let Err(e) = bond::release_on_timeout_without_slashing_checked(pool, &order).await {
        warn!("scheduler_timeout: blameless bond release failed for {} ({});                skipping cancel/republish so next tick retries", order.id, e);
        continue;
    }
}
```

i.e. add an error-propagating variant alongside the `_or_warn` one (`release_bonds_for_order` / `release_active_bonds` already return `Result`), and let the caller decide. I'd keep the `_or_warn` wrappers for the call sites that genuinely cannot retry; this one can.

Note this is not a pre-existing-behaviour argument: `slash_or_release_on_timeout` returns `Result` specifically so this call site can `continue`. The new path opted out of that.

src/inbox.rs
```diff
@@ -0,0 +1,1934 @@
...
+}
+
+impl Default for InboxHealth {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl InboxHealth {
+    pub fn new() -> Self {
```
grunch: _🩺 Stability & Availability_ | _🟡 Minor_

**Blind windows are process-local, so the largest outage of all — the daemon being down — is never credited.**

`InboxHealth::new()` starts with `windows: Vec::new()` and `installed_at = now`. A crash, an OOM kill, an LND-driven restart, a deploy: for the entire time the process was not running, Mostro could not hear a single message, which is the strongest possible form of the condition this module exists to compensate for. None of it is recorded.

After the restart the first audit lands within 30s, `is_confirmed_listening()` flips true, and an order that spent 20 minutes of its 15-minute window waiting on a daemon that was not running is owed exactly zero — cancelled and, where bonds are enabled, slashed on the first tick. The user did nothing wrong and had no way to know.

The module's stated premise covers this case squarely:

> punishing them for the silence would be punishing them for the node's failure

Two options, in order of preference:

1. **Persist the windows.** A small `inbox_blind_windows` table (or a single `last_seen_listening` row updated by the watchdog) turns process downtime into a window `[last_seen_listening, installed_at]` opened at startup. It also survives the 7-day retention meaningfully.
2. **At minimum, treat the startup gap as blind** by seeding a window from the last known-listening timestamp when one is recoverable, and document the restart case as uncompensated when it is not.

Not blocking on its own, but worth an explicit decision in the PR description rather than silence — right now `docs/EVENT_ROUTING.md` promises per-order fairness that does not hold across a restart.

src/scheduler.rs
```diff
@@ -54,10 +55,58 @@ pub async fn start_scheduler(ctx: AppContext) {
...
+async fn job_inbox_watchdog(ctx: AppContext) {
+    #[allow(deprecated)]
+    let event_kind = ctx.settings().mostro.transport.event_kind();
+    let subscription = crate::inbox::InboxSubscription::new(ctx.keys().public_key(), event_kind);
+
+    tokio::spawn(async move {
+        loop {
+            // Sleep first: at startup the event loop has just subscribed, and a
+            // REQ still in flight would look exactly like a missing one.
+            tokio::time::sleep(tokio::time::Duration::from_secs(INBOX_WATCHDOG_INTERVAL)).await;
```
grunch, line 110: _🩺 Stability & Availability_ | _🟡 Minor_

**The watchdog is unsupervised, and a poisoned health mutex converts a single panic into permanent silent failure.**

Two coupled robustness gaps:

1. **No supervision.** This `tokio::spawn` is fire-and-forget with no `JoinHandle`, no restart, and nothing that notices if the task dies. If `check_inbox_health` ever panics, the verdict freezes at whatever it last was. Frozen at `Listening` is the dangerous direction: `is_confirmed_listening()` stays true forever and `job_cancel_orders` resumes slashing against an inbox nobody is auditing any more — with no log line, since the panic is in a detached task. Every other guarantee in this PR is downstream of this loop still running.

2. **Poison cascade.** Every accessor on `InboxHealth` does `.lock().expect("inbox health mutex poisoned")`. `HealthState` holds only plain data and no user code runs under the guard, so a panic inside the critical section is unlikely — but an allocation failure or a future refactor makes it reachable, and the blast radius is disproportionate: `is_confirmed_listening()` panics → the `job_cancel_orders` task dies → **timeouts stop permanently**, hold invoices ride to CLTV expiry, bonds never resolve. A poisoned lock over data with no invariant to protect should be recovered, not propagated:

```rust
fn state(&self) -> MutexGuard<'_, HealthState> {
    self.state.lock().unwrap_or_else(|e| e.into_inner())
}
```

and route the ~10 call sites through it — which also removes ten copies of the same `.expect` string.

For (1), wrapping the body in `AssertUnwindSafe(...).catch_unwind()` with an `error!` and continuing, or simply re-spawning on `JoinError`, would keep a single bad audit from disarming the feature.

src/scheduler.rs
```diff
@@ -54,10 +55,58 @@ pub async fn start_scheduler(ctx: AppContext) {
...
+///
+/// Short enough that a lost ear is measured in seconds rather than the 60s
+/// timeout tick, long enough that it is not a source of traffic on its own.
+/// Hardcoded like the other maintenance intervals in this module.
+///
+/// Visible to `crate::inbox` because the two pacing knobs interact: the audit
+/// draws on the same per-relay budget as the event loop, so this interval is
+/// what decides how many doublings pass before `RESUBSCRIBE_MAX_BACKOFF` is
+/// the thing actually limiting retries.
+pub(crate) const INBOX_WATCHDOG_INTERVAL: u64 = 30;
```
grunch, line 84: _📐 Maintainability & Code Quality_ | _🔵 Trivial_

**`pub(crate)` for a constant nothing outside this module reads.**

The doc comment justifies the visibility as:

> Visible to `crate::inbox` because the two pacing knobs interact

but `src/inbox.rs` references `INBOX_WATCHDOG_INTERVAL` only in prose (the `RESUBSCRIBE_MAX_BACKOFF` and `resubscribe_relay` doc comments) — there is no code use, so nothing enforces the coupling the comment describes. Either make it `const` (private) and reword the doc to a plain cross-reference, or have `inbox.rs` actually use it, e.g. `debug_assert!(RESUBSCRIBE_INITIAL_BACKOFF.as_secs() < INBOX_WATCHDOG_INTERVAL)` so the relationship the comment relies on cannot silently break.

Separately: `src/inbox.rs` is 1934 lines (808 implementation, 1126 tests), well past the 800-line guideline the repo follows elsewhere. Splitting `InboxHealth` + `BlindWindow` into `src/inbox/health.rs` and leaving the subscription/keeper in `src/inbox/mod.rs` would track the two genuinely separate concerns — and the health half is what `scheduler.rs` actually depends on.

grunch (changes requested):
## Review summary

Strong, well-reasoned PR that fixes a real and serious hole. The problem statement is accurate, the SDK behaviour it describes matches nostr-sdk 0.45.1 (`relay/inner.rs`, `handle_relay_message`), `cargo clippy --all-targets --all-features` is clean on `ad76e0b`, and the test suite is unusually thoughtful.

Two findings are blocking, and both are in the parts that were added *after* the first round of bot review — the stand-down on provisional closures (4087027) and the per-order downtime credit (50b27a8 / 1525b55). In both cases the code and its own comments disagree about what it does.

- **[Blocking] `src/inbox.rs`** — standing down on `rate-limited` / `auth-required` leaves the relay's stale `EOSE` credit in place, so the watchdog reports a relay that is not serving the inbox as healthy. This reintroduces exactly the silent permanent deafness the PR exists to remove.
- **[Blocking] `src/db.rs` + `src/scheduler.rs`** — the grace passed to `find_order_by_seconds` *narrows* the selection instead of widening it, which makes the per-order compensation unreachable and applies the global maximum to every order, including ones taken long after the outage.

The rest are non-blocking (one correctness asymmetry, two robustness gaps, two nits).

Verdict: **REQUEST_CHANGES**.

src/inbox.rs
AndreaDiazCorreia: Agreed on the gap, and leaving it out of this PR deliberately: it needs schema, not just a code change, and the design is not obvious enough to settle as a rider here.

The part that needs thinking through is what a restart is allowed to claim. A persisted `last_seen_listening` cannot distinguish "crashed while listening" from "shut down deliberately for a week", so a naive `[last_seen_listening, installed_at]` window would open a seven-day outage on the next boot and freeze every timeout on a node that was simply offline on purpose. Bounding that — a ceiling on what a single restart gap can be credited, or a shutdown marker that closes the window cleanly — is the actual work, and it belongs in its own change with its own migration and tests.

So: not fixed here, and the limitation is real. `docs/EVENT_ROUTING.md` describes the per-order credit as it behaves within one process lifetime; across a restart an order genuinely gets nothing for the time the daemon was not running. Tracking it as a follow-up rather than closing this thread as done.

The other five threads are addressed in this branch.

Catrya (changes requested):
Request changes — on one thing. 

The diagnosis is right and the parts that are easy to get wrong are right too:
subscribing only after the notification stream exists (the REQ from `main.rs`
was throwing away the relay's EOSE and anything arriving during the rest of
boot); judging health by an EOSE bound to `connected_at` rather than by the
SDK's subscription map, which records what *we* sent; standing down on the REQ
for provisional closures but not on the verdict; running each audit in its own
task so a panic is a logged `JoinError` and not the silent end of the watchdog;
and `release_on_timeout_without_slashing` propagating the error instead of
funnelling into `_or_warn`, with a test that shows why. `4af1f39`'s predecessor
catching its own narrowing bug in `find_order_by_seconds`, with a regression
test, is the kind of self-correction that makes the rest easier to trust.

One thing I chased so nobody else has to: `on_relay_message` runs inline in the
event loop, which in this repo is the shape of a freeze. It's fine —
`Relay::subscribe`/`unsubscribe` bottom out in `send_client_msg`, a `try_send`
on the transport channel (`nostr-sdk-0.45.1/src/relay/inner.rs:88-93`), with
`wait_until_sent: None`. No network round trip, no blocking send, only short
locks on the subscription map. Worth a line in the module docs so the next
reader doesn't have to go find that out.

### The blocking one: the per-order credit has no ceiling

`blind_seconds_since(order.taken_at)` sums **every** outage window overlapping
the order's wait, and windows are retained for 7 days / 512 entries. The
3-hour bound only covers a *continuous* unconfirmed stretch. With an inbox that
flaps — blind 5 minutes, listening 1, repeat — `is_confirmed_listening()` is
true often enough that `blameless` never fires, while `owed` grows without
bound. An order in `WaitingBuyerInvoice` then never reaches its deadline: the
hold invoice stays encumbered until the CLTV machinery picks it up, and the
taker's bond stays `Locked`.

That is the state `MAX_UNCONFIRMED_INBOX_PAUSE_SECS` exists to prevent. The
reasoning for it is yours and it's written in the code — a permanently deaf
inbox must not turn into escrows held to CLTV expiry and bonds that never come
back. Flapping reaches the same place by a path the bound doesn't watch, so the
PR doesn't yet honour its own rule. Low probability, granted, but the scenario
this whole PR is written for is a deployment whose relays are misbehaving.

`owed.min(exp_seconds as i64)` (or some multiple of it) bounds the worst case
without changing anything about the design.

### Non-blocking

- **The credit is anchored on `taken_at`, and `taken_at = 0` gets none of it.**
  `find_order_by_seconds` selects on `taken_at < ?`, so such an order is always
  selected, and `waited = now - 0` can never be under `exp_seconds + owed`. Not
  caused here — #866 showed CAS pre-trade writes can drop that field — but the  new safety net inherits the same blind spot, on the orders that were already
  worst off. A guard for a non-positive `taken_at`, or a line acknowledging it,
  would settle the question.
- **The blameless unwind notifies through an outbox that is also down.** Past
  the bound every waiting order is unwound and the users are told via `send_dm`,
  which queues against the same relays that aren't answering. They find out by
  refreshing. Inherent, but worth stating in `docs/EVENT_ROUTING.md` so an
  operator knows what to expect after a recovery.
- **`job_expire_pending_older_orders`** — agreed it doesn't belong in this PR,
  but please open the follow-up issue *before* this merges rather than after. A
  `TakeSell` that never arrived still expires an order, and this is exactly the
  kind of note that gets lost once the branch is gone.

Catrya (changes requested):
Reviewed at `fc3df07`. `cargo check --all-targets` ✅, `cargo clippy --all-targets` ✅ (no warnings), `cargo test` ✅ (1281 passed, 0 failed).

## What's good

The design is sound and the coverage is real, not decorative:

- **Stable subscription id** (`mostro-inbox`) so a `CLOSED` frame can be attributed to the inbox — that's the only way to react to one, and it's the right call.
- **Provisional vs. permanent closures** (`auth-required` / `rate-limited` → the SDK re-sends; everything else → the keeper does) is verified against nostr-sdk 0.45.1's actual behaviour, with tests that would fail loudly if a future SDK bump changes it.
- **Health judged by subscription, not by traffic**, with the `EOSE` credit bound to `connected_at`. This closes the "connected relay that silently drops the REQ" hole, which is the subtlest part of the whole problem.
- **Subscribing only after the notification stream exists** (`app.rs`), with `a_receiver_created_after_the_req_misses_its_eose` pinning the reason. Nicely caught.
- No `InboxHealth` `MutexGuard` is ever held across an `.await` — I checked every call site. No deadlock risk in the event loop.
- End-to-end tests against a real `LocalRelay` (`closed_inbox_is_resubscribed_and_hears_again`, the NIP-42 read-gated relay), not just hand-built frames.

The `EVENT_ROUTING.md` write-up is among the best docs in this repo.

## 🔴 Blocking: the `downtime_credit` cap slashes honest bonds

`fc3df07` fixes my earlier finding (unbounded credit → escrow held to CLTV under a flapping relay), but it overcorrects. In `src/scheduler.rs:561`:

```rust
blind_overlap.min(exp_seconds as i64)   // exp_seconds = 900 by default
```

An order is spared iff `waited < exp + min(D, exp)`, where `D` is the outage overlapping its wait and `pre` is how long it had already been waiting when the outage started:

| Outage `D` | Spared iff | Effect |
|---|---|---|
| ≤ 15 min | same as uncapped | correct |
| 20 min | `pre < 10 min` | user loses 5 min of their window |
| **≥ 30 min** | `pre < 0` | **nobody is spared** |

So for any outage of 30 minutes or more, on the **first tick after recovery** every order in `waiting-buyer-invoice` / `waiting-payment` is cancelled and the responsible bond is **slashed**, with zero opportunity for the user to respond. That is precisely the harm this PR exists to prevent.

`blameless` does not cover it: it only arms while `!is_confirmed_listening()`, and recovery flips that back to `false` before the tick runs. The result is an inverted risk profile — **the damaging band is roughly 30 min to 3 h**. A 4-hour outage unwinds blamelessly before it ever recovers and costs nobody anything; a 45-minute one slashes everyone who was waiting.

Two clean ways out:

- **(a)** Bound the credit by `MAX_UNCONFIRMED_INBOX_PAUSE_SECS` (3 h) instead of `exp_seconds`. Worst case an escrow stays encumbered ~3 h 15 min — the same order of magnitude as the pause bound that already exists, and far short of the CLTV horizon (~22 h). The flapping hazard stays bounded, which is all the cap was for.
- **(b)** Keep the `exp_seconds` cap, but route orders whose *uncapped* credit exceeded it through `release_on_timeout_without_slashing`. The cancellation is still hastened, but nobody pays for the node's own deafness.

I'd go with **(a)**: it's one line, and (b) adds another state branch to the hottest part of the scheduler. Either way it needs a test covering "outage of 2×`exp_seconds`, order taken just before it, must not be slashed on recovery".

## Non-blocking notes

1. **The watchdog overrides the keeper's stand-down.** On an `auth-required` closure, `on_relay_message` deliberately withholds the REQ so it doesn't race the SDK's AUTH — but `check_inbox_health` sees `acknowledged == false` 30 s later and sends it anyway, producing exactly the race the keeper avoided. Harmless in practice (worst case a duplicate REQ), but the documented intent and the actual behaviour don't line up.
2. **Outage history is lost on restart.** An order that waited through an outage preceding a daemon restart gets no credit. Acceptable, but worth a line in the docs.
3. `run` and `run_cashu` now duplicate ~60 identical lines of the `ClientNotification` match. They were already partly duplicated, but this widens it.
4. `job_expire_pending_older_orders` isn't gated on inbox health. Correct today, since it only releases bonds and never slashes — worth stating in the doc so nobody adds a slash there later.

## Verdict

**Not ready to merge — one more change needed.**

95% of this (keeper, watchdog, NIP-42, timeout pause, blameless unwind) I'd approve as-is. But the cap added in the last commit trades an "escrow held forever" failure for an "undeserved slash" failure, and the second one costs honest users real money in a far more common scenario — any half-hour relay outage. It's a one-line change plus a test.

PR #875, opened 2026-08-13
fix(bond): pay the timeout-slash winner instead of forfeiting to the node

Catrya:
On a waiting-state **timeout slash**, the slashed counterparty share was
deterministically misallocated to the node instead of being paid to the
wronged party. The payout resolver derived the winner from the live order
row, but the timeout-cancel path wipes the very pubkey it needs — so the
recipient resolved to `None`, no payout invoice was ever requested, and after
the claim window the bond forfeited to the node.

This affects only deployments running the anti-abuse bond feature with
`[anti_abuse_bond] enabled = true` and `slash_on_waiting_timeout = true`
(non-default). Dispute slashes were never affected.

## The bug

1. A taker abandons a trade in a waiting state past its timeout.
2. `scheduler::job_cancel_orders` slashes t
   (`slash_or_release_on_timeout` → `Pendinty is
   owed `(1 − slash_node_share_pct) × amoun
3. The same job then calls `edit_pubkeys_order`, which NULLs the slashed
   side's trade pubkey on the order (buyer buy
   order) to reset/republish it.
4. `resolve_recipient` matches `bond.pubkey` against **both** order pubkeys
   and requires both present; with one wiped it falls through to `None`.
5. `request_payout_invoice` skips on `None`), so
   the winner is never asked for a bolt11. After `payout_claim_window_days`,
   `forfeit_bond` flips the bond to `Forfeiull
   amount.

With the default `slash_node_share_pct = 0.5` the node collects 2× its
advertised share; with `0.0` (advertised asarty")
it pockets everything. The winner has no wo
`AddBondInvoice` dies in `find_recoverable_bond_for_recipient` via the same
resolver.

## Root cause

The payout recipient was derived on demand from a mutable order row that the cancel path rewrites, and the bond row did not persist the counterparty's identity. Once the order pubkey is wiped, the winner is unrecoverable.

## Fix

Persist the winning counterparty's trade push time**, while the order is still intact (bes), and read it back at payout time.

- **Commit 1 (`refactor`)** — plumbing, ine
  nullable column `bonds.payout_recipient_pubkey` (+ migration), `Bond` model
  / `Crud` / column-alignment test wiring,
  `recipient` and writes it in the slash CAS. Both the timeout and dispute
  call sites resolve it from the intact ord
- **Commit 2 (`fix`)** — activation: `resolve_payout_recipient` now prefers
  the captured pubkey, falling back to the  rows
  slashed before the column existed. This ft
  (`request_payout_invoice`) and the proact
  (`find_recoverable_bond_for_recipient`),

## Why dispute slashes were unaffected

The admin dispute path (`AdminSettle` / `Ad
`edit_pubkeys_order`, so the order pubkeys stay intact and the resolver always found the winner. Those rows now populate a

## Testing

- Unit: resolver returns the captured recipfalls
  back correctly (including the pre-fix `None` on a wiped order with no
  captured recipient — documenting the original bug).
- Integration: composes the real timeout slpe +
  `resolve_payout_recipient`, asserting the maker is paid, not the node — the
  coverage gap that let this ship (no existth the
  payout).
- Full suite green (`cargo test`), `cargo fmt` / `cargo clippy` clean.

## Upgrade notes

The migration is additive (`ALTER TABLE bonllable)
and runs automatically on daemon start agaieset,
no downtime. Rows already slashed **and** wbe
backfilled (the pubkey was never stored ands fix
prevents future losses only. Given the feat, few
if any such rows should exist in practice.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bond-slash payouts by preserving the winning counterparty’s payout information.
  * Payouts remain resolvable even after related order details are cleared.
  * Maintained fallback behavior for older bonds without stored payout information.
  * Applied the fix consistently to standard and range-based bond slashes.

* **Reliability**
  * Added safeguards and coverage for dispute, timeout, and legacy payout scenarios.
  * Improved payout resolution when counterparty details are unavailable during processing.

### DIFF
--- a/migrations/20260813120000_bond_payout_recipient.sql
+++ b/migrations/20260813120000_bond_payout_recipient.sql
@@ -1,0 +1,21 @@
+-- Persist the winning counterparty's trade pubkey on the bond row at slash
+-- time so the Phase 3 payout resolver does not have to re-derive it from the
+-- order row later.
+--
+-- The timeout-slash path (`scheduler::job_cancel_orders` →
+-- `slash_or_release_on_timeout`) settles the responsible bond into
+-- `PendingPayout` and then calls `edit_pubkeys_order`, which NULLs the
+-- slashed side's trade pubkey on the order (buyer_pubkey on a sell order,
+-- seller_pubkey on a buy order) to reset/republish it. The payout resolver
+-- (`resolve_recipient`) matches `bond.pubkey` against both order pubkeys and
+-- needs *both* present, so once one is wiped it falls through to `None`: the
+-- wronged counterparty is never asked for a payout invoice and the bond
+-- eventually forfeits to the node instead of compensating them.
+--
+-- Capturing the recipient here — while the order is still intact, before the
+-- wipe runs — makes the winner recoverable regardless of any later mutation
+-- to the order row. Written by the slash CAS in `slash_one`, alongside
+-- `slashed_reason` / `slashed_at` / `node_share_sats`. Nullable: rows slashed
+-- before this migration (and paths that do not populate it) leave it NULL,
+-- and the resolver falls back to the order-derived lookup.
+ALTER TABLE bonds ADD COLUMN payout_recipient_pubkey char(64);
--- a/src/app/bond/crud.rs
+++ b/src/app/bond/crud.rs
@@ -43,6 +43,7 @@ pub(crate) const BOND_INSERT_COLUMNS: &[&str] = &[
     "taker_amount",
     "taker_fee",
     "taker_dev_fee",
+    "payout_recipient_pubkey",
 ];
 
 fn push_bond_insert_binds(b: &mut Separated<'_, Sqlite, &'static str>, bond: &Bond) {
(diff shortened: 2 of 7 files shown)


grunch (approved):
Reviewed the full change at head 487cdd4, including both prior bot reviews (CodeRabbit had no actionable comments; ermeme's range-maker blocker is resolved on this head).

The fix is correct and complete:

- `payout_recipient_pubkey` is captured at slash time from the still-intact order at all three slash sites: the timeout path in `slash_or_release_on_timeout`, the admin dispute path in `apply_bond_resolution` (uniformity only — that path never wipes order pubkeys), and the Phase 6 slice-slash child insert in `record_maker_slice_slash`.
- `resolve_payout_recipient` guard ordering is right: the Phase 6 maker-refund case (`parent_bond_id` set, `child_order_id` NULL) is checked before the captured pubkey, so refund-remainder rows — which never populate the column — keep paying the maker; legacy rows fall back to the order-derived lookup unchanged.
- Both consumer flows go through the same resolver (`request_payout_invoice` and `find_recoverable_bond_for_recipient`), so the scheduler poll and the proactive `AddBondInvoice` path are both fixed.
- Column plumbing is complete (`BOND_INSERT_COLUMNS`, insert/update binds, the column-alignment sentinel test) and no production query decodes bonds with an explicit column list, so nothing misses the new column. The migration is additive and nullable.
- Regression coverage is exactly what was missing pre-fix: two integration tests compose the real slash + `edit_pubkeys_order` wipe + resolver for both the taker/non-range and range-maker paths, each also documenting the pre-fix `None`.

CI is green on this head (build, test, clippy, fmt, MSRV). LGTM.

AndreaDiazCorreia (approved):
tACK

PR #876, opened 2026-08-14
fix: persist taken_at in the pre-trade take CAS writes

Catrya:
## Summary

  Regression introduced in #866, shipped in v0.18.3: **every taken order is canceled by the scheduler seconds after the take**, with the hold invoice canceled and both parties notified `canceled`.

  Observed on staging: a sell order was taken, the buyer supplied their invoice, the seller received the hold invoice to pay — and 10 seconds later the scheduler's timeout job canceled everything with "buyer did not sent regular invoice in time".

  ## Root cause

  #866 replaced the pre-trade full-row `Crud::update` writes with compare-and-swap
  helpers that enumerate their columns explicitly. That enumeration missed
  `taken_at`: the take handlers set it in memory via `order.set_timestamp_now()`
  (a mostro-core method, so no visible field assignment in the handlers), and
  before #866 it reached the DB implicitly through the full-row write. With the
  CAS writes it never gets persisted, so taken orders keep the schema default
  `taken_at = 0`.

  The scheduler's `find_order_by_seconds` selects
  `WHERE taken_at < now - expiration_seconds AND status IN
  ('waiting-buyer-invoice', 'waiting-payment')` — with `taken_at = 0` a taken
  order is "expired" the instant it enters a waiting state, and the next
  `job_cancel_orders` tick cancels it.

  ## The fix

  Add `taken_at` to the column lists of both take-side CAS writes:

  - `cas_complete_pretrade_take` (take_sell / take_buy / show_hold_invoice)
  - `cas_promote_taker_context` (bond promotion path)

  The narrow `cas_pretrade_order_status` flip is untouched — the cancel path
  does not own that column. No schema change and no new migration: `taken_at`
The narrow `cas_pretrade_order_status` flip is untouched — the cancel path
does not own that column. No schema change and no new migration: `taken_at`
has existed since the original orders migration; existing databases work
as-is with the new binary.

This completes #866's design rather than reverting it: the status guard
(first-committer-wins) is unchanged, and `taken_at` belongs to the take
transition — its only other writer (`update_order_to_initial_state`) resets
it on republish, outside the guarded window.

## Tests

- `cas_complete_pretrade_take_covers_every_legitimate_caller_state` now also
- New `cas_promote_taker_context_persists_taker_context_and_take_time`:
  db-level coverage of the promotion write, including `taken_at`, and that it
  never touches `status`.
- New regression test `freshly_taken_order_is_not_scheduler_timeout_eligible`:
  drives the real take transition (`set_timestamp_now()` + wide CAS) against
  `find_order_by_seconds` — the end-to-end contract the scheduler depends on,
  which neither #866's tests nor the scheduler's (which insert rows with
  `taken_at` pre-set) covered. Fails on v0.18.3, passes with the fix.

```text
cargo test --bin mostrod    # 1176 passed, 0 failed, 2 ignored
cargo clippy                # clean
cargo fmt                   # clean

Also verified live against a running node: a take now persists taken_at,
and the timeout tick no longer cancels freshly taken orders.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

- **Bug Fixes**
  - Pre-trade orders now retain their take timestamps when saved.
  - Newly taken orders are no longer incorrectly selected for timeout cancellation.
  - Taker details, amounts, invoices, and fees remain intact during persistence.

- **Tests**
  - Added coverage confirming take timestamps and related order details persist correctly.
  - Verified that promoting taker information does not unexpectedly change order status.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -886,7 +886,7 @@ pub async fn cas_promote_taker_context(
     };
     let query = format!(
         "UPDATE orders SET {taker_columns}, fiat_amount = ?6, amount = ?7, fee = ?8, \
-         dev_fee = ?9, created_at = ?10 \
+         dev_fee = ?9, created_at = ?10, taken_at = ?11 \
          WHERE id = ?1 AND status IN ({PRETRADE_STATUSES})"
     );
     // The unused ?5 slot for buy orders binds a throwaway `None`.
@@ -917,6 +917,7 @@ pub async fn cas_promote_taker_context(
         .bind(order.fee)
         .bind(order.dev_fee)
         .bind(order.created_at)
+        .bind(order.taken_at)
         .execute(pool)
         .await
         .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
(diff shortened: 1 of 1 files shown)


AndreaDiazCorreia (approved):
tACK

PR #877, opened 2026-08-14
feat(rpc): authenticate the admin gRPC interface

### LINKED ISSUE #807: [HIGH] Admin gRPC service has no code-level authentication
### Severity: HIGH

The admin gRPC service has no authentication at the code level — anyone who can reach the port can call privileged operations.

### Where
- `src/rpc/server.rs` — builds a plain `tonic` transport (no TLS, no auth interceptor).
- `src/rpc/service.rs` — `cancel`, `settle`, `add_solver`, `take_dispute` are exposed; `validate_db_password` is effectively a no-op.

### Problem
The admin RPC surface performs fund-moving and dispute-controlling actions (settle/cancel orders, add solvers, take disputes). There is no interceptor enforcing a shared secret / mTLS / token, and the documented password check does not actually gate the calls. It is safe **by default** only because it is typically bound to loopback and disabled, but there is no in-code guard preventing exposure if an operator binds it more broadly or reverse-proxies it.

### Suggested fix
- Add a `tonic` interceptor that enforces authentication (mTLS client cert or a bearer/HMAC token from config) on every admin RPC method.
- Make `validate_db_password` (or a replacement) actually reject unauthenticated calls, fail-closed.
- Document the exposure model and default-deny posture.

### Note
This is a design/hardening issue rather than an active exploit in the default config, but the admin surface is high-value and should not rely solely on network placement.

AndreaDiazCorreia:
Closes #807.

Adds transport-level authentication to the admin gRPC interface, plus startup guards so it cannot be exposed unintentionally.

**What changed**

- A bearer token (`MOSTRO_RPC_TOKEN`, environment-only) is required on every method, checked by a tonic interceptor in constant time.
- Startup is now fatal when `[rpc].enabled = true` without a token, when a non-loopback address is bound without `[rpc].allow_remote = true`, or when only half a TLS pair is configured.
- Optional TLS via `[rpc].tls_cert_path` / `[rpc].tls_key_path`.
- `[rpc].listen_address` is validated with the same parser `RpcServer::bind` uses, so a config that passes validation is guaranteed to bind.
- Rejected unauthenticated requests are log-throttled: the first from a peer within 60s at `warn!`, the rest at `debug!`, behind a capped peer table.
- `docs/RPC.md`, `README.md` and `settings.tpl.toml` updated accordingly.

**Config change (breaking for existing RPC users)**

`[rpc].enabled = true` now requires `MOSTRO_RPC_TOKEN` in the environment or in
`~/.mostro/.env`; the daemon refuses to start without it. The default remains
`enabled = false`, so nodes that never turned the RPC on are unaffected.
`ValidateDbPassword` now requires the token as well, which affects clients using
it as a health check.

`[rpc].listen_address` must be an IP literal, with IPv6 bracketed (`127.0.0.1`,
`[::1]`, `0.0.0.0`). Hostnames such as `localhost` passed config validation
before but never bound, so no working configuration changes behaviour: the
daemon now refuses them at validation with an actionable message instead of
exiting later on `Invalid address`.

**Dependencies**

`tonic` gains the `tls-ring` feature, and `subtle` becomes a direct dependency in
`Cargo.toml` for the constant-time credential comparison. No new crates enter the
tree: `subtle` was already there transitively, and `Cargo.lock` only adds a
single dependency edge to `tokio-rustls`, also already present.

**Tests**

```text
test result: ok. 1205 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
```

Includes an end-to-end test that serves the API on an ephemeral port and asserts an anonymous call is rejected while an authenticated one succeeds.

## Summary by CodeRabbit

- **New Features**
  - Added bearer-token authentication for administrative RPC requests.
  - Added optional TLS support and controls for enabling remote RPC access.
  - Added startup validation for tokens, bind addresses, remote exposure, and TLS settings.
  - RPC connections now require valid IP-literal addresses, including bracketed IPv6.
  - Authentication failures are logged with throttling to reduce repetitive warnings.

- **Documentation**
  - Expanded RPC configuration, authentication, TLS, security, and administration guidance.
  - Added authenticated command-line and client examples with remote-access warnings.


src/config/util.rs
```diff
@@ -73,6 +78,143 @@ fn validate_mostro_settings(settings: &Settings) -> Result<(), MostroError> {
...
+
+    Ok(())
+}
+
+/// True when `addr` can only be reached from the host itself.
+///
+/// Accepts the `localhost` literal alongside IP literals because
+/// `settings.toml` has always allowed it, and bracketed IPv6 (`[::1]`) because
+/// that is how the address is written in a `host:port` pair.
+fn is_loopback_address(addr: &str) -> bool {
```
grunch, line 128: 🟠 **Major** — _Functional Correctness_

**The validator accepts addresses the server cannot bind.**

`is_loopback_address` treats `localhost` and unbracketed `::1` as valid loopback spellings, but `RpcServer::bind` still builds the address as `format!("{}:{}", listen_address, port).parse::<SocketAddr>()`. I checked with `SocketAddr::from_str`:

```text
localhost:50051 -> Err(AddrParseError(Socket))
::1:50051       -> Err(AddrParseError(Socket))
[::1]:50051     -> Ok
127.0.0.1:50051 -> Ok
```

So with `listen_address = "::1"` (the most common way to write IPv6 loopback) or `"localhost"` the daemon passes `validate_rpc_settings`, then hits `exit(1)` in `main.rs` with `RPC server failed to start: Invalid address`. The doc comment says `settings.toml` "has always allowed it" — it has always been *accepted by the parser*, but it has never bound. `loopback_is_recognised_in_every_written_form` now asserts the contradictory half.

Suggested fix: validate the address with the same parser `bind` uses, so a config that passes validation is guaranteed to bind:

```suggestion
/// True when `addr` can only be reached from the host itself.
///
/// Parses the address exactly the way `RpcServer::bind` does, so an address
/// that passes validation is guaranteed to bind. Bracketed IPv6 (`[::1]`) is
/// accepted because that is the only form `SocketAddr` parses in a
/// `host:port` pair; `localhost` and bare `::1` are not, and are rejected
/// here with a clear message instead of failing later with `Invalid address`.
fn is_loopback_address(addr: &str) -> bool {
    format!("{addr}:0")
        .parse::<std::net::SocketAddr>()
        .is_ok_and(|socket| socket.ip().is_loopback())
}
```

and update the test fixture to `["127.0.0.1", "127.0.0.53", "[::1]"]`, plus a negative case that `"localhost"` / `"::1"` produce an error mentioning the accepted spellings (the `allow_remote` branch will currently fire for them, which is misleading — an explicit "not a parseable IP, use 127.0.0.1 or [::1]" check before the loopback check would be clearer).

If you prefer to keep `localhost` working, the alternative is resolving via `ToSocketAddrs` in `bind` — but then the validator and the binder must share that code path; today they do not.

src/rpc/auth.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        match presented {
+            Some(candidate) if credentials_match(candidate, self.token.expose_secret()) => {
+                Ok(request)
+            }
+            // One message for every failure mode: a caller learns whether the
+            // port is an admin RPC, never whether it guessed part of a token.
+            _ => {
+                match request.remote_addr() {
+                    Some(addr) => warn!("Rejected unauthenticated admin RPC from {}", addr.ip()),
```
grunch: 🟠 **Major** — _Stability & Availability_

**Unauthenticated traffic is unmetered and each request writes a log line.**

The `RateLimiter` runs inside the handler (`check_rate_limit` in `service.rs`), i.e. *after* this interceptor, so a request without a valid token never reaches it. Every such request emits one `warn!` with the peer IP. On loopback that is harmless; with the new `allow_remote = true` path it means any host that can reach the port gets one log line per request for free, forever — a zero-cost log-amplification / disk-fill vector, and it also buries genuine audit lines.

The module doc argues "deliberately not rate-limited" for *guessing*, which I agree with — the entropy argument holds. It does not cover the logging side effect.

Minimal fix without wiring a second limiter: throttle the log, not the decision. E.g. log at `warn!` only on the first rejection per peer IP within a window and at `debug!` afterwards, using a small `std::sync::Mutex<HashMap<IpAddr, Instant>>` (or `dashmap`) held in `BearerAuth`. Alternatively downgrade the per-request line to `debug!` and emit a periodic aggregated `warn!` ("N unauthenticated admin RPC attempts from M peers in the last 60s"). Either way, please also spell the trade-off out in the module doc so the next reader knows it was considered.

src/rpc/server.rs
```diff
@@ -146,33 +228,150 @@ mod tests {
...
+        use tonic::Request;
+
+        init_test_settings();
+        let _lock = RPC_TOKEN_LOCK.lock().await;
+        let token = "t".repeat(32);
+        let _guard = RpcTokenGuard::set(&token);
+
+        // Reserve an ephemeral port and release it, so parallel runs of the
+        // suite cannot collide on a fixed one.
+        let port = std::net::TcpListener::bind("127.0.0.1:0")
```
grunch: 🟡 **Minor** — _Test Reliability_

**Reserve-and-release is a TOCTOU; the kernel can hand the port to another process (or another test binary) in the gap.**

Under `cargo test` with several binaries running concurrently this will eventually flake with `Failed to bind`. `TcpIncoming::bind` already owns the listener, and `tonic::transport::server::TcpIncoming::local_addr()` exists in 0.14, so `bind` can return the resolved address alongside the future:

```rust
// in RpcServer::bind
let incoming = TcpIncoming::bind(addr).map_err(|e| format!("Failed to bind {addr}: {e}"))?;
let bound = incoming.local_addr().map_err(|e| format!("Failed to read bound address: {e}"))?;
info!("RPC server listening on {} ({})", bound, transport);
// ...
Ok((bound, builder.add_service(..).serve_with_incoming(incoming)))
```

Then the test can use `server_at("127.0.0.1", 0)` and connect to `bound`, with no gap. Side benefit: the `listening on` log line becomes correct when an operator configures `port = 0` (today it would print `:0`).

grunch (changes requested):
Strict review of head `79b0b86`. I re-checked every thread already posted (Codex: TLS failure propagation, non-ASCII tokens; CodeRabbit: timing claim, argv exposure, case-insensitive scheme, RPC readiness) and none of those are repeated below — all of them are addressed on this head, and the `bind()` split is the right shape for the fail-closed invariant.

**Verified locally on this head**
- `cargo fmt --all -- --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test --bin mostrod -- rpc:: config::util::` → 82 passed, 0 failed (includes `served_rpc_rejects_calls_without_the_token` and `bind_rejects_malformed_tls_material`) ✅
- Confirmed in tonic 0.14.5 source that `Router::serve_with_incoming` still applies `self.tls` (`serve_internal` → `ServerIoStream::new(.., self.tls)`) and that `TcpIncoming` items implement `Connected`, so `remote_addr()` is populated for the interceptor and the rate limiter.

**What is good**
- Interceptor is minimal, constant-time on the credential, single failure message, scheme case-insensitive per RFC 7235.
- Startup validation is fatal on every unsafe combination, and `bind()` re-checks the token so the server can never serve ungated even if validation is bypassed.
- Tests cover the regression that matters most (service registered *without* the interceptor) end to end.

**Why REQUEST_CHANGES (2 actionable items, both small)**
1. `validate_rpc_settings` / `is_loopback_address` accept `listen_address` spellings (`localhost`, unbracketed `::1`) that `RpcServer::bind` rejects, so the daemon passes validation and then exits with `Invalid address`. The PR adds a test that pins this contradiction in.
2. Unauthenticated requests skip the `RateLimiter` entirely and each one writes a `warn!` line — with the new `allow_remote = true` path that is a zero-cost log-amplification vector from the network.

Details inline. Neither is a hole in the auth gate itself; both are quick fixes.

**Nits (non-blocking)**
- PR description says "No new crates" — `subtle` is a new *direct* dependency in `Cargo.toml` (it was already in the tree transitively). Worth one word in the description.
- `auth.rs:46` `split_once(' ')`: RFC 7235 allows `1*SP` between scheme and token68, so `Bearer  <token>` (two spaces) is refused. Fine to keep strict, just noting it is a deliberate choice.

Catrya (dismissed):
ACK

src/rpc/server.rs
```diff
@@ -24,34 +57,84 @@ impl RpcServer {
...
+        };
 
-        if let Err(e) = server.await {
-            error!("RPC server error: {}", e);
-            return Err(Box::new(e));
-        }
+        // Binds eagerly: an occupied port is a startup error, not a surprise
+        // discovered later by whoever happens to read the logs.
+        let incoming =
+            TcpIncoming::bind(addr).map_err(|e| format!("Failed to bind {addr}: {e}"))?;
```
grunch: 🟠 **Major** — _Performance_

**Switching to `serve_with_incoming` silently turns Nagle back on.**

The old code called `Server::serve(addr)`, which goes through `Server::bind_incoming`:

```rust
// tonic-0.14.6/src/transport/server/mod.rs:653
fn bind_incoming(&self, addr: SocketAddr) -> Result<TcpIncoming, super::Error> {
    Ok(TcpIncoming::bind(addr)
        .map_err(super::Error::from_source)?
        .with_nodelay(Some(self.tcp_nodelay))   // default: true (mod.rs:132)
        .with_keepalive(self.tcp_keepalive)
        .with_keepalive_interval(self.tcp_keepalive_interval)
        .with_keepalive_retries(self.tcp_keepalive_retries))
}
```

`TcpIncoming::bind` on its own leaves `nodelay: None`, and tonic is explicit about it:

> `/// The `tcp_nodelay` and `tcp_keepalive` settings are ignored when using this method.` — `serve_with_incoming`, mod.rs:701

So the admin RPC now runs with Nagle enabled where it previously did not. For a request/response protocol with small frames that is up to ~40 ms of delayed-ACK interaction per call — invisible on loopback under low load, very visible for a remote operator, which is exactly the deployment `allow_remote` newly permits. Nothing in the PR description suggests this was intended.

The eager-bind design is worth keeping; it just has to carry the settings forward:

```diff
-        let incoming =
-            TcpIncoming::bind(addr).map_err(|e| format!("Failed to bind {addr}: {e}"))?;
+        // `serve_with_incoming` ignores the builder's TCP settings, so the
+        // defaults `Server::serve` used to apply are set on the listener here.
+        let incoming = TcpIncoming::bind(addr)
+            .map_err(|e| format!("Failed to bind {addr}: {e}"))?
+            .with_nodelay(Some(true));
```

src/rpc/server.rs
```diff
@@ -111,6 +188,49 @@ mod tests {
...
+        RpcServer {
+            listen_address: listen_address.to_string(),
+            port,
+            tls: None,
+        }
+    }
+
+    // `MOSTRO_RPC_TOKEN` is process-wide state, so the tests that touch it run
+    // serially. Async-aware because the guard is held across awaits.
+    static RPC_TOKEN_LOCK: tokio::sync::Mutex<()> = tokio::sync::Mutex::const_new(());
```
grunch: 🟠 **Major** — _Test Reliability_

**A second, independent environment lock does not serialize against the existing one.**

`src/config/util.rs:372` already has:

```rust
static ENV_LOCK: Mutex<()> = Mutex::new(());
```

guarding `set_var`/`remove_var` on `MOSTRO_NSEC_PRIVKEY` in `env_file_tests`. This adds a *different* mutex for `MOSTRO_RPC_TOKEN`. The crate has no `src/lib.rs`, so both modules compile into one test executable:

```text
$ cargo test --no-run
  Executable unittests src/main.rs (target/debug/deps/mostrod-baa8dbe54bceb886)
  Executable tests/cashu_mint.rs (...)
```

With the default parallel harness, an `env_file_tests` thread holding `ENV_LOCK` and an `rpc::server::tests` thread holding `RPC_TOKEN_LOCK` mutate `environ` concurrently. glibc's `setenv` reallocates the `environ` array without synchronization, so this is a genuine data race, not a logical one — it can corrupt an unrelated `getenv` rather than just return a stale token. (This is why `std::env::set_var` is `unsafe` from edition 2024 onward; the crate is on 2021, so the compiler will not flag it.)

There is a third path too: `startup_validation_tests` now reaches `read_rpc_token_env_var()` through `validate_mostro_settings` → `validate_rpc_settings`, i.e. a `getenv` under no lock at all.

The fix is one lock for the process environment, not one per variable. Promote `ENV_LOCK` to a crate-visible `pub(crate) static` (e.g. in `config::util` or a small `testutil` module) and have `RpcTokenGuard` take it. If holding it across `.await` is the reason a `tokio::sync::Mutex` was chosen here, note that the guards in these tests are only held across `bind` and the client calls — a `std::sync::Mutex` guard would work if acquired and released around the env mutation itself rather than for the whole test body, or keep one `tokio::sync::Mutex` and convert the `util.rs` call sites.

src/config/util.rs
```diff
@@ -73,6 +79,149 @@ fn validate_mostro_settings(settings: &Settings) -> Result<(), MostroError> {
...
+
+    match token {
+        None => {
+            return Err(MostroInternalErr(ServiceError::IOError(format!(
+                "[rpc].enabled = true but {RPC_TOKEN_ENV_VAR} is not set: the admin RPC would \
+                 accept every caller that can reach the port. Set {RPC_TOKEN_ENV_VAR} in the \
+                 environment or <settings_dir>/.env, or set [rpc].enabled = false."
+            ))));
+        }
+        Some(token) if token.expose_secret().chars().count() < MIN_RPC_TOKEN_LEN => {
```
grunch, line 169: 🟠 **Major** — _Security_

**The case for skipping rate limiting rests on entropy this check does not measure.**

`src/rpc/auth.rs` states the threat model plainly:

> The authentication *decision* is deliberately not rate-limited. `MIN_RPC_TOKEN_LEN` keeps the search space far out of reach of online guessing

That holds for a token from `openssl rand -base64 32`. It does not hold for what this validator actually accepts: 32 printable-ASCII characters, any distribution. `"a".repeat(32)` passes — the test helper `valid_token()` on line 650 is literally that string, as is `"t".repeat(32)` throughout `rpc::server::tests`. An operator who types a 32-character passphrase gets a daemon that boots happily, advertises "strong auth" in `docs/RPC_RATE_LIMITING.md`, and offers an attacker unlimited unthrottled guesses — the throttle added in `should_warn` caps the *log lines*, not the attempts, and `check_rate_limit` never runs because the interceptor rejects first.

With `allow_remote = true` those guesses come from off-host at line rate.

Two ways out, either is fine:

- Enforce a crude entropy floor alongside the length, so the docs' claim is one the code backs:
  ```rust
  let distinct = token.expose_secret().chars().collect::<std::collections::BTreeSet<_>>().len();
  if distinct < 16 { /* reject: not a randomly generated token */ }
  ```
  Cheap, no false positives on real base64 output (44 chars from `openssl rand -base64 32` yield ~30 distinct on average), and it catches every hand-typed value.
- Or keep accepting any 32 characters and add a failure counter to the interceptor, so the entropy assumption stops being load-bearing.

What should not stay is the current pairing: a length-only gate cited in three documents as the reason no counter is needed.

src/rpc/server.rs
```diff
@@ -24,34 +57,84 @@ impl RpcServer {
...
+            format!("Refusing to start the admin RPC server: {RPC_TOKEN_ENV_VAR} is not set")
+        })?;
 
-        info!("Starting RPC server on {}", addr);
+        let admin_service = AdminServiceImpl::new(my_keys, pool, ln_client);
 
-        let server = Server::builder()
-            .add_service(AdminServiceServer::new(admin_service))
-            .serve(addr);
+        let mut builder = Server::builder();
```
grunch: 🟡 **Minor** — _Stability & Availability_

**No connection, stream, or request limits on a server this PR makes remotely reachable.**

`Server::builder()` with defaults means: no `timeout`, no `max_concurrent_streams` (hyper's per-connection default applies, nothing caps connection *count*), no `http2_keepalive_interval`, no `max_connection_age`. Before this PR that was defensible — the thing only ever listened on loopback. `allow_remote = true` changes the audience, and an unauthenticated peer can hold connections open indefinitely: the interceptor rejects the *request*, it does not close the *connection*.

The bearer token stops the attacker from doing anything; it does not stop them from exhausting file descriptors on the daemon that also runs the trade engine. Worth setting the obvious three while you are already reshaping this builder:

```diff
-        let mut builder = Server::builder();
+        // The admin surface is reachable off-host under `allow_remote`, and
+        // the interceptor rejects requests without closing connections, so the
+        // transport carries its own limits.
+        let mut builder = Server::builder()
+            .timeout(Duration::from_secs(30))
+            .max_concurrent_streams(Some(64))
+            .http2_keepalive_interval(Some(Duration::from_secs(30)))
+            .http2_keepalive_timeout(Some(Duration::from_secs(10)));
```

src/main.rs
```diff
@@ -273,10 +273,30 @@ async fn main() -> Result<()> {
...
+            }
+        };
+
+        // Only the accept loop is detached; it is already listening.
         tokio::spawn(async move {
-            match rpc_server.start(rpc_keys, rpc_pool, rpc_ln_client).await {
-                Ok(_) => tracing::info!("RPC server started successfully"),
-                Err(e) => tracing::error!("RPC server failed to start: {}", e),
+            match rpc_serving.await {
+                Ok(()) => tracing::warn!("RPC server stopped"),
```
grunch: 🟡 **Minor** — _Functional Correctness_

**The `Ok(())` arm quietly abandons the invariant the rest of the PR establishes.**

The comment above `bind` argues that `[rpc].enabled = true` should be "an invariant rather than a hope", and the `Err` arm calls `exit(1)` to keep it. The `Ok` arm logs a warning and lets the daemon keep running with no admin API and no non-zero exit — the operator's monitoring sees a healthy process, and `enabled = true` has silently become `false`.

When can it fire? `serve_internal` returns `Ok(())` only when the incoming stream yields `None` (mod.rs:842), and `TcpListenerStream` never terminates, so in practice this is unreachable today. That is an argument for making it explicit rather than for leaving it inconsistent — a future `serve_with_incoming_shutdown` or a different incoming type would turn a dead branch into a silent degradation.

```diff
             match rpc_serving.await {
-                Ok(()) => tracing::warn!("RPC server stopped"),
+                // The listener stream never ends, so this is unreachable
+                // today; treat it like any other loss of the admin API rather
+                // than letting `enabled = true` silently become false.
+                Ok(()) => {
+                    tracing::error!("RPC server stopped accepting connections");
+                    exit(1);
+                }
                 Err(e) => {
```

src/config/util.rs
```diff
@@ -2,12 +2,18 @@
...
-use crate::config::secret::read_nsec_env_var;
+use crate::config::constants::{
+    ENV_FILENAME, MAX_DEV_FEE_PERCENTAGE, MIN_DEV_FEE_PERCENTAGE, MIN_RPC_TOKEN_LEN,
+    RPC_TOKEN_ENV_VAR,
+};
+use crate::config::secret::{read_nsec_env_var, read_rpc_token_env_var};
+use crate::config::types::RpcSettings;
 use crate::config::wizard;
 use crate::config::{init_mostro_settings, Settings};
+use crate::rpc::server::listen_socket_addr;
```
grunch: 🟡 **Minor** — _Maintainability_

**This points the dependency the wrong way: `config` now needs `rpc`.**

The motivation is right — validation and `bind` must not drift on how an address is parsed — but the shared parser does not have to live in `rpc::server` for that. `listen_socket_addr` is pure string→`SocketAddr` logic with no RPC in it; putting it in `config` (next to the settings it validates) and having `RpcServer::bind` call `crate::config::...::listen_socket_addr` gives the identical no-drift guarantee with `rpc → config`, which is the direction every other module already uses.

As it stands, `config::util` is the only place in the crate where a lower-level module reaches up into a service module, and `pub(crate)` on `listen_socket_addr` exists solely to permit it.

src/rpc/server.rs
```diff
@@ -146,33 +266,173 @@ mod tests {
...
+            .bind(Keys::generate(), Arc::new(pool), offline_ln_client().await)
+            .err()
+            .expect("an admin RPC without a token must never serve");
+        assert!(error.to_string().contains(RPC_TOKEN_ENV_VAR));
+    }
+
+    #[tokio::test]
+    async fn bind_rejects_malformed_tls_material() {
+        init_test_settings();
+        let _lock = RPC_TOKEN_LOCK.lock().await;
```
grunch: 🟡 **Minor** — _Test Coverage_

**The TLS path is only ever tested through its failure branch.**

`bind_rejects_malformed_tls_material` proves bad PEM is refused. Nothing proves good PEM produces a server that actually speaks TLS. The regression that would hurt is the mirror image of what `validate_rpc_settings` guards against elsewhere — "reads as TLS is on while serving plaintext":

```rust
builder = builder.tls_config(ServerTlsConfig::new().identity(Identity::from_pem(cert, key)))?;
```

Drop the reassignment (`builder.tls_config(...)?;` — a plausible edit, and one that only warns rather than errors) and the daemon reads both files, validates both PEMs, logs `RPC server listening on … (TLS)`, and serves cleartext. Every test in this file and in `rpc::auth` stays green, and `settings.toml` still says TLS is on. That is precisely the failure mode the half-configured-pair check exists to prevent, reachable by a different route.

`served_rpc_rejects_calls_without_the_token` is the right shape to extend: add `rcgen` as a dev-dependency, generate a self-signed cert for `127.0.0.1` into the temp dir, and drive the same `GetVersion` probe through a `Channel` with `ClientTlsConfig`, asserting the anonymous call is `Unauthenticated` and the authenticated one succeeds. It also gives the `tls-ring` feature — a new dependency edge in `Cargo.lock` — its first exercise.

docs/RPC.md
```diff
@@ -124,45 +148,102 @@ service AdminService {
...
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let channel = Channel::from_static("http://127.0.0.1:50051")
         .connect()
         .await?;
-    
-    let mut client = AdminServiceClient::new(channel);
-    
+
+    let token: MetadataValue<_> =
```
grunch, line 199: 🔵 **Nit** — _Documentation_

**The only client example cannot reach a TLS-configured server.**

The configuration section above now documents `tls_cert_path`/`tls_key_path`, and the Security Considerations section tells operators to *never* expose the port beyond loopback without TLS. An operator who follows that advice then arrives here and finds:

```rust
let channel = Channel::from_static("http://127.0.0.1:50051").connect().await?;
```

which fails against their own server, with a transport error that says nothing about TLS. Since this page is also the one the README points to as the safe alternative to `grpcurl` on a shared host, it is the page that has to carry the TLS variant:

```rust
use tonic::transport::{Channel, ClientTlsConfig, Certificate};

let tls = ClientTlsConfig::new()
    .ca_certificate(Certificate::from_pem(std::fs::read("/etc/mostro/rpc-cert.pem")?))
    .domain_name("mostro.example.com");
let channel = Channel::from_static("https://mostro.example.com:50051")
    .tls_config(tls)?
    .connect()
    .await?;
```

A one-line note that the plaintext example is the loopback case would also do.

src/config/util.rs
```diff
@@ -73,6 +79,149 @@ fn validate_mostro_settings(settings: &Settings) -> Result<(), MostroError> {
...
+                "{RPC_TOKEN_ENV_VAR} must contain only printable ASCII characters and no spaces: \
+                 it is sent as an HTTP header, so any other value can never authenticate a \
+                 client. `openssl rand -base64 32` produces a valid token."
+            ))));
+        }
+        Some(_) => {}
+    }
+
+    // Before the loopback check, or an unbindable address would be reported as
+    // a remote-exposure problem: `localhost` and bare `::1` read as loopback to
```
grunch, line 210: 🔵 **Nit** — _Functional Correctness_

**`port = 0` passes validation and then binds a different port on every restart.**

`listen_socket_addr("127.0.0.1", 0)` parses fine, so `[rpc].port = 0` is accepted here, `bind` succeeds, and `info!("RPC server listening on 127.0.0.1:41783")` scrolls past. Every configured admin client then fails to connect, with a fresh port after each restart. The tests rely on this behaviour deliberately (`server_at("127.0.0.1", 0)`), which is fine — it is only the *configured* value that is never a sensible 0.

Every other way to get this block wrong is startup-fatal with an actionable message; this one is cheap to bring in line:

```rust
if rpc.port == 0 {
    return Err(MostroInternalErr(ServiceError::IOError(
        "[rpc].port = 0 asks the kernel for an ephemeral port, which changes on every \
         restart and no client can be configured against. Set a fixed port."
            .to_string(),
    )));
}
```

grunch (changes requested):
## Review — strict pass over `fix/rpc-admin-authentication`

Verified locally on the PR head: `cargo clippy --all-targets --all-features` clean, `cargo fmt --check` clean, `cargo test` → **1205 passed, 0 failed, 2 ignored**. The design is sound — the interceptor is the right layer given that every handler runs as the daemon identity, `bind` failing eagerly instead of inside a detached task is a real improvement, and the `listen_socket_addr` sharing between validation and bind closes a genuine drift bug.

What follows is what survived a second pass against the `tonic` 0.14.6 sources in the lockfile.

### Blocking

1. **`TCP_NODELAY` is silently lost** (`src/rpc/server.rs`). Moving from `Server::serve(addr)` to `TcpIncoming::bind` + `serve_with_incoming` drops a setting the old path applied. Concrete latency regression, one-line fix.
2. **Test-binary environment races.** `RPC_TOKEN_LOCK` in `src/rpc/server.rs` and `ENV_LOCK` in `src/config/util.rs:372` are independent mutexes inside the *same* unittest binary (`target/debug/deps/mostrod-*` — the crate has no `lib.rs`, so every `#[cfg(test)]` module lands in one executable). `setenv`/`unsetenv` concurrent with `getenv` is a data race in glibc.

### Worth resolving before merge

3. The security argument for not rate-limiting the auth decision assumes entropy that nothing enforces (see inline on `src/config/util.rs`).
4. No connection or stream limits on a server that this PR makes remotely bindable.
5. The `Ok(())` arm in `src/main.rs` contradicts the startup invariant the PR argues for everywhere else.

### Non-blocking

6. `config` now depends on `rpc` (layering).
7. The TLS path has no positive test — only its failure branch is exercised.
8. `docs/RPC.md` documents `tls_cert_path`/`tls_key_path` but the client example only shows plaintext.
9. `[rpc].port = 0` passes validation.

The docs, `settings.tpl.toml` and `STARTUP_AND_CONFIG.md` claims all check out against the code — including the `#[serde(default)]` audit on `RpcSettings`, which is accurate as written.

AndreaDiazCorreia:
@grunch While rebasing this branch onto main I hit an overlap with #934, which landed after this PR was opened.

#934 added an optional bearer token via [rpc].auth_token in settings.toml, checked per-method on the five mutating RPCs. This PR adds a mandatory token via MOSTRO_RPC_TOKEN, enforced by a tonic interceptor on every method including GetVersion, with a 32-character minimum validated at startup and a hard refusal to boot without it.

Rebasing mechanically leaves both in place, which is worse than either alone: if [rpc].auth_token is set to a value different from MOSTRO_RPC_TOKEN, every mutating RPC fails even though the interceptor accepted the caller.

From the review thread in #934, auth_token was optional so existing deployments would keep working ("Unset keeps the historical bind-address-only model"). That argument no longer holds here, since this PR refuses to start without MOSTRO_RPC_TOKEN either way.

How would you like me to resolve it?

PR #881, opened 2026-08-16
fix: dispatch buyer payouts at most once per settled order

Catrya:
Once a hold invoice is settled, the buyer payout involves several moving parts:
`do_payment`, the payment-retry scheduler, `AddInvoice` (a buyer replacing their
invoice after a failed attempt), and daemon restarts. Each of them decided on its
own whether a payment still had to be sent, based on indirect signals
(`payment_attempts`, order status, in-memory retry tasks). Under retries, invoice
replacement, or a restart at the wrong moment, those views could disagree — an
order could get stuck in `settled-hold-invoice` with nothing driving its payment,
or more than one dispatch could end up in flight for the same escrow.

This PR makes the pending payout an explicit, persisted piece of state and the
single source of truth for every path that touches it.

### In-flight payout marker, claimed by CAS (811254c)

- New column `payout_payment_hash` on `orders` (migration): set for a
  `settled-hold-invoice` order right before its payout is handed to LND.
- `do_payment` claims the marker via compare-and-swap immediately before
  `send_payment`; a concurrent dispatch or a re-armed retry loses the CAS and is
  skipped.
- `pay_new_invoice` rejects an `AddInvoice` replacement while a payout is in
  flight, so a fresh invoice can no longer reset `payment_attempts` on top of a
  pending payment — the buyer is asked to wait until the pending one resolves.
- `find_failed_payment` (retry scheduler) skips orders with a payout in flight.
- A new reconciliation job resolves stale markers against LND:
  `Succeeded` → finalize the order, `Failed`/`Unknown` → clear the marker and
  re-arm the retry, `InFlight` → keep waiting. A stranded payout therefore
  neither blocks the order forever nor gets dispatched twice, including across
  restarts.

### Claim-age grace window for reconciliation (8a7cccf)

- `payout_claimed_at` (migration) is sealed in the same CAS as the marker, and
  reconciliation only examines claims older than the payment-retry interval.
  This keeps a reconciliation tick that lands between claiming and LND
  registering the payment from misreading it as unknown and clearing the marker
  prematurely.
- The claim is taken only after the LND connection is established, so a failed
  connect never leaves a marker with no payment behind it, and the
  claim→dispatch window narrows to the `send_payment` call itself.

### Tests (bf66db9)

- `claim_order_payout` is atomic (second claim loses the CAS) and refuses orders
  not in `settled-hold-invoice`.
- `clear_order_payout` / `fail_order_payout` release the marker;
  `fail_order_payout` re-arms the retry.
- `find_failed_payment` excludes orders with a payout in flight.
- `find_inflight_payouts` returns only marked orders.
- `pay_new_invoice` rejects an `AddInvoice` replacement while a payout is in
  flight.
- `hex_to_bytes` round-trips and rejects malformed input.

## Notes

- Two migrations, applied automatically on startup; the marker columns default
  to NULL so existing rows are unaffected.
- Happy-path behavior is unchanged: one settled order → one payment, same
  messages to both parties. Also fixes the long-standing case of an order stuck
  in `settled-hold-invoice` after a crash mid-payment, which previously required
  manual intervention.

## Summary by CodeRabbit

* **New Features**
  * Added safeguards to prevent duplicate buyer payouts.
  * Added automatic reconciliation for in-progress Lightning payouts.
  * Added periodic background checks to complete or reattempt pending payouts.

* **Bug Fixes**
  * Prevented invoice swaps while a buyer payout is pending.
  * Improved recovery after interruptions and failed or unknown payment states.
  * Preserved pending payouts when payment status cannot be confirmed.
  * Ensured successful payouts are finalized before buyer notifications are sent.

### DIFF
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -667,18 +667,21 @@ pub async fn do_payment(
                                 "Order Id {}: Invoice with hash: {} paid!",
                                 order.id, msg.payment.payment_hash
                             );
-                            let _ = payment_success(
-                                &ctx,
-                                &mut order,
-                                buyer_pubkey,
-                                &my_keys,
-                                request_id,
-                            )
-                            .await;
-                            // Terminal success: release our own claim only.
-                            let _ =
-                                crate::db::clear_order_payout(ctx.pool(), order.id, &payout_hash)
-                                    .await;
+                            // Release our claim only if the order actually
+                            // reached Success. If finalization fails, keep the
+                            // marker so reconciliation retries it — clearing it
+                            // here would strand a paid order with no recovery.
+                            if payment_success(&ctx, &mut order, buyer_pubkey, &my_keys, request_id)
+                                .await
+                                .unwrap_or(false)
+                            {
+                                let _ = crate::db::clear_order_payout(
+                                    ctx.pool(),
+                                    order.id,
+                                    &payout_hash,
+                                )
+                                .await;
+                            }
                         }
                         PaymentStatus::Failed => {
                             warn!(
@@ -686,13 +689,17 @@ pub async fn do_payment(
                                 order.id, msg.payment.payment_hash
                             );
 
-                            // Mark payment as failed and release our own claim
-                            // (scoped to this hash) so a fresh invoice can be
-                            // retried without erasing a newer claim.
-                            check_failure_retries_or_log(&ctx, &order, request_id).await;
-                            let _ =
-                                crate::db::clear_order_payout(ctx.pool(), order.id, &payout_hash)
-                                    .await;
+                            // Release our own claim (scoped to this hash) and
+                            // re-arm retry. Only do the failure bookkeeping and
+                            // buyer notification if we still owned the claim, so
+                            // a stale watcher never pollutes a newer payout's
+                            // retry state or notifies against it.
+                            if crate::db::fail_order_payout(ctx.pool(), order.id, &payout_hash)
+                                .await
+                                .unwrap_or(false)
+                            {
+                                check_failure_retries_or_log(&ctx, &order, request_id).await;
+                            }
                         }
                         _ => {}
                     }
@@ -704,60 +711,79 @@ pub async fn do_payment(
     Ok(())
 }
 
+/// Finalize a paid order: transition `settled-hold-invoice` → `Success` and
+/// notify the buyer, but only after the status CAS actually commits.
+///
+/// Returns `Ok(true)` when the order is now terminal — this call committed the
+/// transition, or a concurrent task already did — meaning the caller may safely
+/// release the payout marker. Returns `Ok(false)` when the transition could not
+/// be built/persisted (`update_order_event` failed): the caller must KEEP the
+/// marker so reconciliation retries finalization, otherwise the buyer would be
+/// paid on an order stranded in `settled-hold-invoice` with no recovery hook.
+///
+/// Buyer notifications (`PurchaseCompleted`, `Rate`) are enqueued only after a
+/// successful commit, so a retried finalization never spams the buyer.
 async fn payment_success(
     ctx: &AppContext,
     order: &mut Order,
     buyer_pubkey: PublicKey,
     my_keys: &Keys,
     request_id: Option<u64>,
-) -> Result<()> {
-    // Purchase completed message to buyer
+) -> Result<bool> {
+    let pool = ctx.pool();
+
+    let order_updated = match update_order_event(my_keys, Status::Success, order).await {
+        Ok(updated) => updated,
+        // Could not build/publish the Success event: leave the order in
+        // settled-hold-invoice and signal "not finalized" so the caller keeps
+        // the marker for reconciliation.
+        Err(_) => return Ok(false),
+    };
+
+    // Only update status and event_id to avoid overwriting fields modified by
+    // concurrent processes (dev_fee_paid, dev_fee_payment_hash, etc.)
+    // The WHERE guard prevents double success transitions from concurrent tasks.
+    let result =
+        sqlx::query("UPDATE orders SET status = ?, event_id = ? WHERE id = ? AND status = ?")
+            .bind(&order_updated.status)
+            .bind(&order_updated.event_id)
+            .bind(order_updated.id)
+            .bind(Status::SettledHoldInvoice.to_string())
+            .execute(pool)
+            .await
+            .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
+
+    if result.rows_affected() == 0 {
+        // Another task already finalized this order: it is terminal, so the
+        // caller may release the marker, but the notifications were already
+        // sent by that task — do not duplicate them.
+        tracing::warn!(
+            "Order {} not transitioned to success: already processed by another task",
+            order_updated.id
+        );
+        return Ok(true);
+    }
+
+    // Committed by us — notify the buyer now.
     enqueue_order_msg(
         None,
-        Some(order.id),
+        Some(order_updated.id),
         Action::PurchaseCompleted,
         None,
         buyer_pubkey,
         None,
     )
     .await;
-
-    let pool = ctx.pool();
-
-    if let Ok(order_updated) = update_order_event(my_keys, Status::Success, order).await {
-        // Only update status and event_id to avoid overwriting fields modified by
-        // concurrent processes (dev_fee_paid, dev_fee_payment_hash, etc.)
-        // The WHERE guard prevents double success transitions from concurrent tasks.
-        let result =
-            sqlx::query("UPDATE orders SET status = ?, event_id = ? WHERE id = ? AND status = ?")
-                .bind(&order_updated.status)
-                .bind(&order_updated.event_id)
-                .bind(order_updated.id)
-                .bind(Status::SettledHoldInvoice.to_string())
-                .execute(pool)
-                .await
-                .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
-
-        if result.rows_affected() == 0 {
-            tracing::warn!(
-                "Order {} not transitioned to success: already processed by another task",
-                order_updated.id
-            );
-            return Ok(());
-        }
-
-        // Send dm to buyer to rate counterpart
-        enqueue_order_msg(
-            request_id,
-            Some(order_updated.id),
-            Action::Rate,
-            None,
-            buyer_pubkey,
-            None,
-        )
-        .await;
-    }
-    Ok(())
+    enqueue_order_msg(
+        request_id,
+        Some(order_updated.id),
+        Action::Rate,
+        None,
+        buyer_pubkey,
+        None,
+    )
+    .await;
+    Ok(true)
 }
 
 /// Reconcile a single in-flight buyer payout against LND.
@@ -799,13 +825,27 @@ pub async fn reconcile_inflight_payout(
 
     match ln_client.lookup_payment_status(&hash_bytes).await {
         Ok(Some(PaymentStatus::Succeeded)) => {
-            if let Ok(Some(mut order)) = Order::by_id(pool, order_id).await {
-                let my_keys = ctx.keys().clone();
-                if let Ok(buyer_pubkey) = order.get_buyer_pubkey() {
-                    let _ = payment_success(ctx, &mut order, buyer_pubkey, &my_keys, None).await;
+            // Clear the marker only once the order is actually finalized. If
+            // finalization fails (or the buyer key is unreadable), keep the
+            // marker so a later tick retries it instead of stranding a paid
+            // order in settled-hold-invoice.
+            let finalized = match Order::by_id(pool, order_id).await {
+                Ok(Some(mut order)) => {
+                    let my_keys = ctx.keys().clone();
+                    match order.get_buyer_pubkey() {
+                        Ok(buyer_pubkey) => {
+                            payment_success(ctx, &mut order, buyer_pubkey, &my_keys, None)
+                                .await
+                                .unwrap_or(false)
+                        }
+                        Err(_) => false,
+                    }
                 }
+                _ => false,
+            };
+            if finalized {
+                crate::db::clear_order_payout(pool, order_id, payout_payment_hash).await?;
             }
-            crate::db::clear_order_payout(pool, order_id, payout_payment_hash).await?;
         }
         Ok(Some(PaymentStatus::Failed)) | Ok(Some(PaymentStatus::Unknown)) | Ok(None) => {
             crate::db::fail_order_payout(pool, order_id, payout_payment_hash).await?;
@@ -2008,8 +2048,8 @@ mod tests {
         // Act
         let result = payment_success(&ctx, &mut order, buyer, &my_keys, None).await;
 
-        // Assert
-        assert!(result.is_ok());
+        // Assert: committed the transition (returns true) and notified the buyer.
+        assert!(result.unwrap(), "a committed finalization returns true");
         let db_order = Order::by_id(&pool, order.id).await.unwrap().unwrap();
         assert_eq!(db_order.status, Status::Success.to_string());
         let actions = queued_actions_for(order.id).await;
@@ -2033,12 +2073,18 @@ mod tests {
         // Act
         let result = payment_success(&ctx, &mut order, buyer, &my_keys, None).await;
 
-        // Assert: early return — status untouched, no Rate message queued.
-        assert!(result.is_ok());
+        // Assert: the guarded UPDATE matched no rows (already finalized
+        // elsewhere), so the call reports terminal (`true`) — the caller may
+        // release the marker — but sends no duplicate notifications, and the
+        // status is left untouched.
+        assert!(
+            result.unwrap(),
+            "a no-op CAS (already processed) is terminal and returns true"
+        );
         let db_order = Order::by_id(&pool, order.id).await.unwrap().unwrap();
         assert_eq!(db_order.status, Status::Active.to_string());
         let actions = queued_actions_for(order.id).await;
-        assert!(actions.contains(&Action::PurchaseCompleted));
+        assert!(!actions.contains(&Action::PurchaseCompleted));
         assert!(!actions.contains(&Action::Rate));
     }
 }
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -674,8 +674,10 @@ pub async fn do_payment(
                                 request_id,
                             )
                             .await;
-                            // Terminal success: release the in-flight marker.
-                            let _ = crate::db::clear_order_payout(ctx.pool(), order.id).await;
+                            // Terminal success: release our own claim only.
+                            let _ =
+                                crate::db::clear_order_payout(ctx.pool(), order.id, &payout_hash)
+                                    .await;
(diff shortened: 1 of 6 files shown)


src/scheduler.rs
```diff
@@ -243,6 +244,56 @@ async fn job_retry_failed_payments(ctx: AppContext) {
...
+                            error!("payout reconcile for order {order_id}: {e}");
+                        }
+                    }
+                }
+                Err(e) => error!("payout reconcile: find_inflight_payouts failed: {e}"),
+            }
+            tokio::time::sleep(tokio::time::Duration::from_secs(interval)).await;
+        }
+    });
+}
```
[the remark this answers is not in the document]
Catrya: This is a fair robustness point, but I'd like to defer it out of this PR, and here's the reasoning.

LndConnector is built via fedimint_tonic_lnd::connect, which is lazy (see the note at src/lightning/mod.rs:646): it builds a tonic Channel that connects on the first RPC and, per tonic's channel semantics, re-establishes the connection on subsequent RPCs after a transient failure. So for the common cases — an LND restart or a network blip — calls fail during the outage but the same client recovers on its own once LND is back; it's not "failing forever until the daemon restarts." During the outage, reconcile_inflight_payout logging and skipping is the intended behavior (don't act on unknown state), and the markers are picked up again on the next successful tick.

Rebuilding the client would only help in the rarer cases where the lazy channel genuinely can't recover — e.g. cert/macaroon rotation, where re-reading the files is required. Those are uncommon.

Also worth noting: the create-once/reuse-forever pattern here mirrors every other long-lived LN job in the scheduler (job_process_bond_payouts, run_dev_fee_cycle, the escrow-deadline job). If we want rebuild-on-repeated-failure, I think it belongs in a separate, cross-cutting change that applies the same treatment (a shared "rebuild after N consecutive failures + escalate the log to error") to all of them consistently, rather than a one-off here that leaves the siblings on the old pattern.
Its better in a follow-up issue to track that

src/app/release.rs
```diff
@@ -724,6 +760,67 @@ async fn payment_success(
...
+    ln_client: &mut LndConnector,
+    order_id: uuid::Uuid,
+    payout_payment_hash: &str,
+) -> Result<(), MostroError> {
+    let pool = ctx.pool();
+
+    // A payment hash is exactly 32 bytes (64 hex chars). Decode with the same
+    // `FromHex` used across the codebase and length-check it; a bad-hex or
+    // wrong-length marker is corrupt, so treat it as malformed and re-arm rather
+    // than sending a truncated hash to LND.
```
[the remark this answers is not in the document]
Catrya: reconcile_inflight_payout already has a /// doc comment (the block starting "Reconcile a single in-flight buyer payout against LND…", covering the Succeeded / Failed / Unknown / InFlight branches). This comment is anchored on the inner // note about hash decoding, not on the function — the public API is already documented, so no change needed here.

src/app/release.rs
```diff
@@ -668,59 +711,153 @@ pub async fn do_payment(
...
+    if result.rows_affected() == 0 {
+        // Another task already finalized this order: it is terminal, so the
+        // caller may release the marker, but the notifications were already
+        // sent by that task — do not duplicate them.
+        tracing::warn!(
+            "Order {} not transitioned to success: already processed by another task",
+            order_updated.id
+        );
+        return Ok(true);
+    }
```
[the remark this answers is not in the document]
Catrya: Technically true that a zero-row CAS doesn't prove Success, but it can't cause a re-payment or a stuck order here. Zero rows means the order has left settled-hold-invoice, and both payout jobs are scoped to that status:

- find_failed_payment: ... AND status == 'settled-hold-invoice' AND payout_payment_hash IS NULL
- find_inflight_payouts: payout_payment_hash IS NOT NULL AND status == 'settled-hold-invoice' ...

There's no transition back into settled-hold-invoice, so once it's gone neither job touches the order again — clearing the marker is safe. Keeping it "for reconciliation" wouldn't help either, since find_inflight_payouts is also settled-hold-invoice-scoped, so it'd just leave a dead marker nothing ever clears.

Catrya:
> 2\. `src/app/release.rs:756-764`— `rows_affected() == 0`se trata como prueba de que otra tarea finalizó el pedido con éxito, pero esta consulta solo prueba que la fila ya no está en `settled-hold-invoice`. Si una ruta de tiempo de espera/cancelación/disputa mueve el pedido primero, esta rama devuelve `Ok(true)`, el llamador borra el marcador de pago y el pago LND ya liquidado ya no se concilia con `Success`. Vuelva a leer el pedido y devuelva `true`solo cuando el estado actual sea realmente `Success`; para cualquier otro estado, mantenga el marcador para que la conciliación pueda reintentar/fallar ruidosamente en lugar de descartar silenciosamente el estado de pago pagado.

Good catch that a zero-row CAS doesn't prove Success — but I don't think it can drop a paid payout or cause a double-pay here, and "keep the marker for reconciliation" wouldn't recover it.

A zero-row result means the order is no longer in settled-hold-invoice. Both jobs that could re-dispatch or reconcile a payout are scoped to that exact status: find_failed_payment matches "status == 'settled-hold-invoice' AND payout_payment_hash IS NULL", and find_inflight_payouts matches "payout_payment_hash IS NOT NULL AND status == 'settled-hold-invoice'". So once the order has moved off settled-hold-invoice (to Success or to a terminal admin/dispute state), neither job touches it again — clearing the marker is safe (no path re-enters settled-hold-invoice, so no second payout), and keeping it wouldn't let reconciliation retry, because reconciliation is itself settled-hold-invoice-scoped. It would just leave a dead marker nothing ever clears.

The real concern under this — an order moved to a non-Success terminal while its payout already settled on LND — is an inconsistency created upstream (the timeout/cancel/dispute path acting on an order with a payout in flight), not something payment_success can repair after the fact. If you'd like it surfaced rather than silent, I'm happy to add a read-back that logs at error level when the status is neither settled-hold-invoice nor Success, while still releasing the marker (since holding it can't reconcile a non-settled-hold-invoice order).

src/app/release.rs
```diff
@@ -640,23 +690,39 @@ pub async fn do_payment(
...
                             );
 
-                            // Mark payment as failed
-                            check_failure_retries_or_log(&ctx, &order, request_id).await;
+                            // Release our own claim (scoped to this hash) and
+                            // re-arm retry. Only do the failure bookkeeping and
+                            // buyer notification if we still owned the claim, so
+                            // a stale watcher never pollutes a newer payout's
+                            // retry state or notifies against it.
+                            if crate::db::fail_order_payout(ctx.pool(), order.id, &payout_hash)
```
grunch: _🗄️ Data Integrity_ | _🟠 Major (non-blocking)_ | _⚡ Quick win_

**The claimed hash is not a unique claim token when a retry reuses the same buyer invoice.**

Scoping `clear_order_payout` / `fail_order_payout` to `payout_hash` is the right instinct, and it does protect the case the commit message describes (buyer swaps to a *new* invoice → new hash → the stale watcher's release is a no-op). But the retry job re-dispatches `do_payment` with the order re-read from the DB and `buyer_invoice` unchanged (`src/scheduler.rs:236`), so for a plain BOLT11 payout — the common case, as opposed to LNURL/lightning-address where a fresh invoice is fetched per attempt — attempt #2 claims the *same* hash as attempt #1.

Concretely:

1. Attempt #1 claims `H`, `send_payment` succeeds, watcher `W1` is spawned.
2. A reconcile tick sees `Failed` for `H` first and calls `fail_order_payout(order, H)` → marker cleared, retry re-armed.
3. The retry job dispatches attempt #2 with the same invoice → `claim_order_payout(order, H)` wins.
4. `W1` finally drains its `Failed` message and calls `fail_order_payout(order, H)` — which now matches attempt #2's live claim, clears it and re-arms retry while a payout is genuinely in flight.

From there `pay_new_invoice` is unblocked, so a buyer who swaps to a different invoice in that window gets a second dispatch against the same settled escrow — the exact drain this PR closes. The residual risk is small (LND rejects a second `SendPaymentV2` for a hash already in flight, and step 2 requires reconcile to beat a live in-process watcher to the same terminal status, a sub-millisecond race against a 60s poll), which is why I'm not blocking. But the invariant would be unconditional rather than probabilistic with a claim token that is unique per claim rather than per invoice.

Cheapest fix, no schema change: `payout_claimed_at` is already sealed in the same CAS — return it from `claim_order_payout`, carry it into the watcher alongside `payout_hash`, and add `AND payout_claimed_at = ?` to both release statements. A stale caller then loses the CAS even when the hash matches.

src/app/release.rs
```diff
@@ -668,59 +734,153 @@ pub async fn do_payment(
...
+/// terminal update — a held/stranded HTLC, or a payout whose watcher task was
+/// lost across a restart. Without it, such an order would stay locked forever
+/// (regressing the do-payment-stuck bug); with it, the marker is authoritative
+/// only until LND confirms the real outcome:
+///
+/// - `Succeeded` → finalize as `Success` (idempotent via the status CAS) and
+///   clear the marker.
+/// - `Failed` / `Unknown` / not found → clear the marker and re-arm retry;
+///   `send_payment`'s own pre-send `track_payment_v2` check backstops any race
+///   where LND actually still has the payment.
```
grunch: _📐 Correctness of documented rationale_ | _🟡 Minor_ | _⚡ Quick win_

**The cited backstop does not actually cover this branch — `send_payment`'s pre-send guard queries a different hash.**

This doc justifies treating `Ok(None)` (LND does not know the hash) as "safe to re-arm" by pointing at `send_payment`'s own `track_payment_v2` check. But that check is built from `invoice.signable_hash()`:

```rust
// src/lightning/mod.rs:269
```

`signable_hash()` is the hash of the invoice's signing data, not the BOLT11 payment hash — every other call site in the repo uses `payment_hash()` (`src/app/dev_fee.rs:953`, `src/app/bond/payout.rs:571`, and line 628 of this file). LND will never have a payment registered under a signable hash, so `track` is always `Err` and the `if track.is_ok()` abort can never fire: the guard is effectively dead code and is not backstopping anything.

The branch is still safe in practice — LND itself refuses a second `SendPaymentV2` for a payment hash already in flight or settled — but the rationale as written is wrong, and someone reasoning about this code later will trust a guard that does not exist. Two options: fix `src/lightning/mod.rs:269` to use `invoice.payment_hash().as_ref()` (which would also make the guard genuinely useful, and is arguably a standalone bug fix worth its own PR), or reword this bullet to name LND's duplicate-hash rejection as the real backstop.

Pre-existing, not introduced here — flagging because this PR is what leans on it.

src/app/release.rs
```diff
@@ -668,59 +734,153 @@ pub async fn do_payment(
...
+                    }
+                }
+                _ => false,
+            };
+            if finalized {
+                crate::db::clear_order_payout(pool, order_id, payout_payment_hash).await?;
+            }
+        }
+        Ok(Some(PaymentStatus::Failed)) | Ok(Some(PaymentStatus::Unknown)) | Ok(None) => {
+            crate::db::fail_order_payout(pool, order_id, payout_payment_hash).await?;
```
grunch: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_

**Reconcile's failure branch skips the retry bookkeeping the watcher path performs.**

The in-process watcher pairs `fail_order_payout` with `check_failure_retries_or_log` (lines 720-724), which increments `payment_attempts` and — on the first failure — sends the buyer `Action::PaymentFailed` with the retry configuration. This branch only calls `fail_order_payout`, which sets `failed_payment = true` and nothing else.

Consequences for a payout that only ever resolves through reconciliation (watcher lost across a restart — the headline case this job exists for):

- `payment_attempts` does not advance, so a crash → restart → reconcile cycle can burn real dispatch attempts without ever consuming the retry budget.
- The buyer receives no `PaymentFailed` notification for that cycle; they only learn something went wrong on the next attempt that fails through `do_payment`.

Both self-heal on the next real dispatch, so this is a delay rather than a hole, but the asymmetry looks unintentional. Calling `check_failure_retries_or_log` here (after re-reading the order, as the `Succeeded` branch already does) would make the two terminal paths behave identically.

src/app/release.rs
```diff
@@ -668,59 +734,153 @@ pub async fn do_payment(
...
+///
+/// - `Succeeded` → finalize as `Success` (idempotent via the status CAS) and
+///   clear the marker.
+/// - `Failed` / `Unknown` / not found → clear the marker and re-arm retry;
+///   `send_payment`'s own pre-send `track_payment_v2` check backstops any race
+///   where LND actually still has the payment.
+/// - `InFlight` → leave as is; the payout is genuinely pending.
+pub async fn reconcile_inflight_payout(
+    ctx: &AppContext,
+    ln_client: &mut LndConnector,
```
grunch: _🧪 Test Coverage_ | _🟡 Minor_ | _🏗️ Heavy lift_

**The four reconcile branches are the recovery path and have no unit coverage.**

The DB helpers are well covered (`claim`/`clear`/`fail`, stale-hash scoping, grace window — nice tests), but `reconcile_inflight_payout` itself has none: `Succeeded → finalize + clear`, `Failed/Unknown/None → re-arm`, `InFlight → no-op`, and the malformed-hash branch are all unexercised. That is the logic deciding whether a stranded payout gets finalized, retried, or left alone.

The blocker is the concrete `&mut LndConnector` parameter. The repo already has the pattern for this — `CancelLightning` + `StubLnClient` (`src/app/cancel.rs:21`, `:792`) — so a one-method trait (`async fn lookup_payment_status(&mut self, hash: &[u8]) -> Result<Option<PaymentStatus>, MostroError>`) implemented for `LndConnector` would make all four branches testable with a stub and no LND. Worth doing given the coverage bar and that this is a money path.

src/db.rs
```diff
@@ -1190,6 +1195,126 @@ pub async fn find_failed_payment(pool: &SqlitePool) -> Result<Vec<Order>, Mostro
...
+/// Returns `(order_id, payout_payment_hash)` pairs.
+pub async fn find_inflight_payouts(
+    pool: &SqlitePool,
+    claimed_before: i64,
+) -> Result<Vec<(Uuid, String)>, MostroError> {
+    let rows = sqlx::query(
+        r#"
+          SELECT id, payout_payment_hash
+          FROM orders
+          WHERE payout_payment_hash IS NOT NULL AND status == 'settled-hold-invoice'
```
grunch, line 1314: _🗄️ Data Integrity_ | _🔵 Nit_

**A marker can be orphaned once the order leaves `settled-hold-invoice`.**

If `payment_success` commits the `Success` transition but the subsequent `clear_order_payout` fails (DB blip), the order ends up `Success` with `payout_payment_hash` still set. This query's `status == 'settled-hold-invoice'` filter means reconciliation will never pick it up again, so the marker is stranded permanently.

Functionally harmless — `find_failed_payment` and `pay_new_invoice` both also require `settled-hold-invoice`, so nothing acts on it — but it leaves misleading rows for anyone auditing in-flight payouts. Either widen the reconcile sweep to clear markers on terminal orders, or note in the doc comment that a marker on a non-`settled-hold-invoice` order is inert residue.

src/scheduler.rs
```diff
@@ -243,6 +244,56 @@ async fn job_retry_failed_payments(ctx: AppContext) {
...
 }
 
+/// Reconcile buyer payouts left in flight (`payout_payment_hash` set) against
+/// LND. Complements `job_retry_failed_payments`: that job only dispatches fresh
+/// payouts (marker NULL), while this one resolves the durable marker so a
+/// held/stranded payout — or one whose in-process watcher was lost across a
+/// restart — is finalized, re-armed, or left pending based on LND's real state,
+/// instead of blocking the order forever. Runs at startup and every tick.
+async fn job_reconcile_inflight_payouts(ctx: AppContext) {
+    let interval = 60u64;
```
grunch: _📐 Maintainability_ | _🔵 Nit_

**Hardcoded tick interval next to a config-derived grace window.**

`grace_secs` comes from `lightning.payment_retries_interval` (60 by default) but `interval` is a literal `60u64`. An operator who lowers `payment_retries_interval` to tighten retries gets a reconcile loop still polling once a minute; one who raises it gets ticks that mostly find nothing eligible. Deriving both from the same setting — or naming the literal `const RECONCILE_INTERVAL_SECS` with a comment on why it is deliberately independent — would make the relationship explicit.

Also worth a guard: `LightningSettings::default()` has `payment_retries_interval: 0` (`src/config/types.rs:497`), so `.max(1)` yields a 1-second grace window if that default is ever reached in a real run — narrower than the claim→register window it is meant to cover. A floor closer to the template's 60 would be safer than `1`.

grunch (dismissed):
## Review: approved

I re-derived the payout state machine from scratch on the current head (`8a56dfb`) instead of trusting the PR description, and traced every path that can reach `settled-hold-invoice`: `release_action`, `admin_settle`, `job_retry_failed_payments`, `AddInvoice`, the in-process watcher, the new reconcile job, and a restart in the middle of each.

**What I verified**

- **The claim CAS is reachable in every dispatch path.** Both `release_action` (`src/app/release.rs:226-247`) and `admin_settle` persist `settled-hold-invoice` to the DB *before* calling `do_payment`, so `claim_order_payout`'s `status = 'settled-hold-invoice'` guard is satisfied. A stricter status guard here would otherwise have silently swallowed every payout (`do_payment` returns `Ok(())` on a lost CAS) — it does not.
- **No full-row writer can resurrect or wipe the marker.** `payout_payment_hash` / `payout_claimed_at` are not fields of `mostro_core::order::Order`, so `Order::update()` (`src/app/fiat_sent.rs:88`, `src/flow.rs:177`, …) cannot clobber them, and `SELECT *` into `Order` ignores them. The only writers are the three CAS helpers.
- **No column collision.** `20260518120000_bond_payout_payment_hash.sql` adds the same column name to `bonds`, not `orders`; the two new migrations sort after every existing one.
- **`lookup_payment_status` semantics match the branch mapping.** `Ok(None)` is genuinely "LND does not know this hash" (`NotFound` at both the call and the stream, `src/lightning/mod.rs:377-413`), and a transport failure is `Err` — so the `Err(_) => keep the marker` decision in the `send_payment` error path and in reconcile fails on the correct side.
- **The retry budget still bounds retries.** `job_retry_failed_payments` gates on `payment_attempts < retries_number` (`src/scheduler.rs:235`), so `fail_order_payout` setting `failed_payment = true` cannot produce an unbounded dispatch loop.
- **`AddInvoice` cannot re-arm on top of a live payout.** `pay_new_invoice` is the only writer of `buyer_invoice` + `payment_attempts = 0` for this status, and the added `payout_payment_hash IS NULL` guard closes the invoice-swap re-arm drain.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (1195 passed, 0 failed) are all clean locally.

The hash-scoped release (`bad6c9d`) and the `Ok(bool)` finalization contract (`b7a21e6`) both hold under the interleavings I walked, including the double-`payment_success` case (concurrent watcher + reconcile): the status CAS elects one committer, the loser returns `true` without duplicating `PurchaseCompleted`/`Rate`, and `clear_order_payout` is idempotent.

The items below are non-blocking follow-ups. None is a confirmed defect on the happy path or on any interleaving I could construct without an implausibly stalled task, so I am not holding the PR for them — but the first two touch the invariant this PR is named after and I'd like them on record.

grunch (approved):
tACK

Issue #784, opened 2026-06-18
Restore session loses pending bond payouts on terminal-status orders


Catrya:
When a dispute is resolved, the loser's bond is slashed and the winning counterparty must claim their share via `Action::AddBondInvoice`. But the order can reach a **terminal status** (`success`, `canceledbyadmin`, …) while that bond is still in `PendingPayout`. If the winner restores their session at that point, the restore response contains **neither the order** (excluded by its terminal status), **nor the dispute** (excluded because it is no longer active), **nor the pending bond payout** (restore never queries the `bonds` table).

  The client ends up with no context for the order, so it ignores the `add-bond-invoice` messages the daemon keeps re-sending every tick. The winner can never submit an invoice and the payout is eventually lost to `Forfeited` when the claim window elapses.

  This is reproducible from **two different routes**, both of which slash a bond and then drop the order/dispute into
  a terminal/inactive status:

  - **admin-settle** → order ends as `success`
  - **admin-cancel** → order ends as `canceledbyadmin`, dispute as `SellerRefunded`

  ## Steps to reproduce

  1. Complete a sell order up to `active`, with both maker and taker bonds locked.
  2. Open a dispute; admin takes it and resolves it (**settle** or **cancel**). Loser's bond → `PendingPayout`, winner is asked for a payout invoice (`add-bond-invoice`).
  3. Let the order reach its terminal status:
     - settle path: hold invoice settled + release paid → order becomes `success`
     - cancel path: `admin_cancel_action` sets order to `CanceledByAdmin`, dispute to `SellerRefunded`
  4. As the winner, send a `RestoreSession`.

  **Observed:** restore response is empty for this trade — the order/dispute are not listed and no pending bond payout
  is signaled. The daemon keeps emitting `add-bond-invoice` for the order, but the client no longer tracks it and
  discards them. The user can never claim; the bond eventually forfeits.

  **Expected:** restore surfaces the pending bond-payout claim so the winner can submit their invoice and receive the
  slashed-bond share.

  ## Root cause
Restore-session is **order-status-centric**, while the bond payout has its own state machine in the `bonds` table, decoupled from order/dispute status:

  - `find_user_orders_by_master_key` (`src/db.rs`) filters out `EXCLUDED_ORDER_STATUSES`, which includes `'success'`, `'canceledbyadmin'`, `'settledbyadmin'`, `'completedbyadmin'`.
[...]

PR #882, opened 2026-08-16
fix: exclude hyphenated terminal order statuses

21Mill:
`EXCLUDED_ORDER_STATUSES` and `TERMINAL_ORDER_STATUSES` in `src/db.rs` list the four multi-word order statuses **without hyphens**, but `Status`' `Display` impl serializes them as kebab-case, and that is the form actually stored in the `orders` table.

```rust
// src/db.rs — before
const EXCLUDED_ORDER_STATUSES: &str = "'expired','success','canceled','dispute',
    'canceledbyadmin','completedbyadmin','settledbyadmin','cooperativelycanceled'";
```

```rust
// mostro-core 0.14.5, order.rs
Status::CanceledByAdmin       => write!(f, "canceled-by-admin"),
Status::SettledByAdmin        => write!(f, "settled-by-admin"),
Status::CompletedByAdmin      => write!(f, "completed-by-admin"),
Status::CooperativelyCanceled => write!(f, "cooperatively-canceled"),
```

`RestoredOrdersInfo::status` even documents the contract the constants break — *"Current status of the order, serialized as kebab-case"* (`message.rs:490`).

The four single-word entries (`expired`, `success`, `canceled`, `dispute`) do match, so both `NOT IN` filters keep working for those and fail silently for the rest. Two other constants in the same file — `ACTIVE_DISPUTE_STATUSES` (`'in-progress'`) and `PRETRADE_STATUSES` (`'waiting-taker-bond'`) — are hyphenated correctly, which is what makes this look like a typo rather than an alternative encoding.

## Impact

**`find_user_orders_by_master_key`** (restore session) returns orders that are already over. On a live node, 30 rows qualified — months-old `cooperatively-canceled` and `canceled-by-admin` orders — and they were handed back to their owners as restorable session state for trades their clients no longer track.

This also has a knock-on effect: `find_user_orders_by_master_key` has no `LIMIT`, while the follow-up `orders` action rejects the whole request when `ids.len() > max_orders_per_response` (`src/app/orders.rs`). Users whose accumulated dead orders push the restore list past the cap get a blanket `too_many_requests` and cannot refresh at all — the server offers a list its own limit will not accept back. That is how this bug was found.

**`find_active_trade_pubkeys`** (Phase 2 anti-spam gate) never drops the trade pubkeys of admin-closed or cooperatively-canceled orders, because those orders never count as terminal. Their keys stay in the fast-pathed "known keys" set indefinitely, which is precisely what `TERMINAL_ORDER_STATUSES` exists to prevent.

## Verification against real data

Same query, same database, only the status list differs:

| status | before | after |
|---|---|---|
| `cooperatively-canceled` | 28 | — |
| `canceled-by-admin` | 2 | — |
| `pending` | 12 | 12 |
| `waiting-maker-bond` | 5 | 5 |
| `settled-hold-invoice` | 4 | 4 |
| `waiting-taker-bond` | 1 | 1 |

30 terminal orders stop being restored; every genuinely active order is untouched.

## Why the existing tests missed it

`find_active_trade_pubkeys_covers_active_and_disputed_excludes_terminal` does test terminal exclusion, but only inserts `'success'` and `'canceled'` — the two spellings that happen to be correct. No test used a hyphenated status on the terminal side.

Both tests now cover every hyphenated status individually. Confirmed failing before the constant change and passing after:

```
---- find_user_orders_by_master_key_excludes_all_terminal_statuses stdout ----
assertion `left == right` failed: restore must return only the live order
  left: ["waiting-payment", "canceled-by-admin", "completed-by-admin",
         "settled-by-admin", "cooperatively-canceled"]
 right: ["waiting-payment"]

---- find_active_trade_pubkeys_covers_active_and_disputed_excludes_terminal stdout ----
creator_coop must NOT be known (terminal/resolved)
```

Full suite: 1187 passed, 0 failed. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` are clean.

## Interaction with #784

Worth flagging explicitly for reviewers. #784 ("Restore session loses pending bond payouts on terminal-status orders") describes the admin-cancel route on the assumption that `'canceledbyadmin'` is already excluded. It is not — because of this bug — so that branch of #784 does not currently reproduce, while the admin-settle branch (`'success'`, correctly spelled) does.

This change makes the two routes behave the same, which means the admin-cancel branch of #784 will start reproducing once this lands. It does not create that problem and does not make it worse in substance; it removes an accidental mask over half of it. Fixing #784 needs its own change to restore the `bonds` state machine, so it is deliberately out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed bond payouts are now claimable only within the configured claim window.
  * Pending bond payouts remain available without a time limit.
  * Active trade and order results now exclude failed payouts missing a slashing timestamp.
  * Claim-window behavior now correctly uses configured values or the default fallback.

* **Tests**
  * Added coverage for payout eligibility boundaries, configuration, fallback behavior, and missing timestamps.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -13,7 +13,7 @@ use std::sync::Arc;
 use uuid::Uuid;
 
 // Constants for status filtering used across restore session functions
-const EXCLUDED_ORDER_STATUSES: &str = "'expired','success','canceled','dispute','canceledbyadmin','completedbyadmin','settledbyadmin','cooperativelycanceled'";
+const EXCLUDED_ORDER_STATUSES: &str = "'expired','success','canceled','dispute','canceled-by-admin','completed-by-admin','settled-by-admin','cooperatively-canceled'";
 const ACTIVE_DISPUTE_STATUSES: &str = "'initiated','in-progress'";
 
 /// Terminal order statuses for the Phase 2 active-trade-pubkey cache: an
@@ -24,7 +24,7 @@ const ACTIVE_DISPUTE_STATUSES: &str = "'initiated','in-progress'";
 /// disputed order is still active (buyer, seller and the assigned solver keep
 /// messaging), so its trade keys must stay fast-pathed. See
 /// `find_active_trade_pubkeys` and docs/TRANSPORT_V2_SPEC.md §6 Phase 2.
-const TERMINAL_ORDER_STATUSES: &str = "'expired','success','canceled','canceledbyadmin','completedbyadmin','settledbyadmin','cooperativelycanceled'";
+const TERMINAL_ORDER_STATUSES: &str = "'expired','success','canceled','canceled-by-admin','completed-by-admin','settled-by-admin','cooperatively-canceled'";
 
 #[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
@@ -2078,6 +2078,45 @@ mod tests {
             None,
         )
         .await;
+        // Terminal statuses whose serialized form is hyphenated. These are the
+        // ones a mis-spelled TERMINAL_ORDER_STATUSES silently fails to match,
+        // so they must be covered explicitly.
+        insert_order_with_pubkeys(
+            &pool,
+            uuid::Uuid::new_v4(),
+            "cooperatively-canceled",
+            Some("creator_coop"),
+            Some("buyer_coop"),
+            Some("seller_coop"),
+        )
+        .await;
+        insert_order_with_pubkeys(
+            &pool,
+            uuid::Uuid::new_v4(),
+            "canceled-by-admin",
+            Some("creator_cba"),
+            None,
+            None,
+        )
+        .await;
+        insert_order_with_pubkeys(
+            &pool,
+            uuid::Uuid::new_v4(),
+            "settled-by-admin",
+            Some("creator_sba"),
+            None,
+            None,
+        )
+        .await;
+        insert_order_with_pubkeys(
+            &pool,
+            uuid::Uuid::new_v4(),
+            "completed-by-admin",
+            Some("creator_cpa"),
+            None,
+            None,
+        )
+        .await;
 
         // Active dispute with an assigned solver → solver key included.
         sqlx::query(
@@ -2123,12 +2162,77 @@ mod tests {
             "seller_succ",
             "creator_canc",
             "solver_settled",
+            "creator_coop",
+            "buyer_coop",
+            "seller_coop",
+            "creator_cba",
+            "creator_sba",
+            "creator_cpa",
         ] {
             assert!(
                 !keys.contains(k),
                 "{k} must NOT be known (terminal/resolved)"
             );
         }
+    }
+
+    /// Restore-session must not hand back orders that are already over.
+    ///
+    /// Four of the eight excluded statuses serialize with hyphens
+    /// (`canceled-by-admin`, `settled-by-admin`, `completed-by-admin`,
+    /// `cooperatively-canceled`). A status list written without them still
+    /// filters the single-word ones, so the query keeps *looking* correct
+    /// while quietly restoring dead orders — which is why every hyphenated
+    /// status is asserted here individually.
+    #[tokio::test]
+    async fn find_user_orders_by_master_key_excludes_all_terminal_statuses() {
+        let pool = setup_orders_db().await.unwrap();
+        let master_key = "a".repeat(64);
+
+        async fn insert_for_master(pool: &SqlitePool, status: &str, master_key: &str) {
+            sqlx::query(
+                r#"
+                INSERT INTO orders (id, kind, event_id, status, premium, payment_method,
+                                    amount, fiat_code, fiat_amount, created_at, expires_at,
+                                    master_buyer_pubkey, trade_index_buyer)
+                VALUES (?1, 'buy', 'event123', ?2, 0, 'lightning',
+                        100000, 'USD', 100, 1700000000, 1700086400, ?3, 1)
+                "#,
+            )
+            .bind(uuid::Uuid::new_v4())
+            .bind(status)
+            .bind(master_key)
+            .execute(pool)
+            .await
+            .unwrap();
+        }
+
+        // One live order — the only row the user should get back.
+        insert_for_master(&pool, "waiting-payment", &master_key).await;
+
+        for status in [
+            "expired",
+            "success",
+            "canceled",
+            "dispute",
+            "canceled-by-admin",
+            "completed-by-admin",
+            "settled-by-admin",
+            "cooperatively-canceled",
+        ] {
+            insert_for_master(&pool, status, &master_key).await;
+        }
+
+        let orders = super::find_user_orders_by_master_key(&pool, &master_key)
+            .await
+            .unwrap();
+
+        let statuses: Vec<&str> = orders.iter().map(|o| o.status.as_str()).collect();
+        assert_eq!(
+            statuses,
+            vec!["waiting-payment"],
+            "restore must return only the live order, got {statuses:?}"
+        );
     }
 
     #[tokio::test]
--- a/src/db.rs
+++ b/src/db.rs
@@ -30,15 +30,59 @@ const TERMINAL_ORDER_STATUSES: &str = "'expired','success','canceled','canceled-
 /// still owe a bond payout to the winning counterparty, so it stays visible
 /// until that obligation resolves.
 ///
-/// `'failed'` belongs beside `'pending-payout'`: a `Failed` row still accepts
-/// a payout invoice while it is inside the `payout_claim_window_days` window
-/// (`app::bond::payout::apply_invoice`), so the claim is live even though the
-/// last `send_payment` gave up. Both states are left for good on payout,
-/// forfeit or slash, so the exception expires without any code to retract it.
-/// Pinned against `BondState`'s serialization by
+/// **The two arms are deliberately asymmetric, and the asymmetry is load
+/// bearing — do not "tidy" it into a single bound.**
+///
+/// `'pending-payout'` is **unbounded**. Such a row is not guaranteed to
+/// forfeit when the claim window elapses: `process_one_bond` only forfeits
+/// while `payout_invoice.is_none()` (`app::bond::payout`), and
+/// `on_send_payment_failure` deliberately *keeps* the invoice when the retry
+/// budget runs out on an `Indeterminate` LND failure. That payout is
+/// genuinely in flight and resolves whenever LND answers definitively, so
+/// bounding this arm on `slashed_at` would strand a live claim.
+///
+/// `'failed'` is **bounded** by `payout_claim_window_days` measured from
+/// `slashed_at`, because that is exactly what `apply_invoice` enforces: past
+/// the window it rejects any invoice on a `Failed` row, so the claim is dead
+/// while the row stays `Failed` forever (nothing transitions it afterwards).
+/// Without the bound the exception would be permanent — the trade keys stay
+/// fast-pathed and the dead order stays restorable indefinitely.
+///
+/// A `Failed` row with a `NULL` `slashed_at` is excluded, matching
+/// `apply_payout_invoice`'s conservative `Rejected` on the same invariant
+/// violation.
+///
+/// The interpolated cutoff is a computed `i64`, never user input, so the
+/// inline interpolation carries no injection risk — same pattern as the
+/// status constants above. Pinned against `BondState`'s serialization by
 /// `claimable_bond_states_match_bond_state_display`.
-const HAS_CLAIMABLE_BOND_PAYOUT: &str = "EXISTS (SELECT 1 FROM bonds b \
-     WHERE b.order_id = orders.id AND b.state IN ('pending-payout','failed'))";
+fn has_claimable_bond_payout() -> String {
+    let claim_window_seconds = Settings::get_bond()
+        .map(|c| c.payout_claim_window_days as i64 * 86_400)
+        .unwrap_or(DEFAULT_CLAIM_WINDOW_SECONDS);
+    has_claimable_bond_payout_at(chrono::Utc::now().timestamp(), claim_window_seconds)
+}
+
+/// Claim window used when the global settings are not initialized, in
+/// seconds. Pinned against `AntiAbuseBondSettings::default()` by
+/// `claim_window_fallback_matches_settings_default`, so the literal cannot
+/// drift away from the config default the way the status constants did.
+const DEFAULT_CLAIM_WINDOW_SECONDS: i64 = 15 * 86_400;
+
+/// [`has_claimable_bond_payout`] with `now` and the window injected, so the
+/// window boundary is testable without a wall clock or the global settings.
+fn has_claimable_bond_payout_at(now: i64, claim_window_seconds: i64) -> String {
+    // Kept while `now - slashed_at < claim_window_seconds`, the exact
+    // complement of `apply_payout_invoice`'s `>=` rejection test — a bond
+    // slashed exactly one window ago is already unclaimable there, so it
+    // must not be kept here either.
+    let cutoff = now.saturating_sub(claim_window_seconds);
+    format!(
+        "EXISTS (SELECT 1 FROM bonds b WHERE b.order_id = orders.id AND (\
+           b.state = 'pending-payout' \
+           OR (b.state = 'failed' AND b.slashed_at IS NOT NULL AND b.slashed_at > {cutoff})))"
+    )
+}
 
 #[cfg(unix)]
 use std::os::unix::fs::PermissionsExt;
@@ -59,9 +103,10 @@ pub async fn find_active_trade_pubkeys(pool: &SqlitePool) -> Result<Vec<String>,
     // of terminal orders that still owe a bond payout: the winner has to
     // answer Mostro's `AddInvoice` nudge, and a key outside this set is held
     // to `pow_first_contact` and dropped before decryption.
+    let claimable_payout = has_claimable_bond_payout();
     let order_query = format!(
         "SELECT buyer_pubkey, seller_pubkey, creator_pubkey FROM orders \
-         WHERE status NOT IN ({TERMINAL_ORDER_STATUSES}) OR {HAS_CLAIMABLE_BOND_PAYOUT}"
+         WHERE status NOT IN ({TERMINAL_ORDER_STATUSES}) OR {claimable_payout}"
     );
     let order_rows = sqlx::query(AssertSqlSafe(order_query))
         .fetch_all(pool)
@@ -1612,7 +1657,7 @@ pub async fn is_dispute_taken_by_admin(
 /// Uses constants for excluded statuses to maintain consistency across queries.
 ///
 /// Terminal orders that still owe a bond payout are kept
-/// ([`HAS_CLAIMABLE_BOND_PAYOUT`]): the winner needs the order and its trade
+/// ([`has_claimable_bond_payout`]): the winner needs the order and its trade
 /// index to answer Mostro's `AddInvoice` nudge, and nothing else in the
 /// restore path carries bond state — `find_user_disputes_by_master_key`
 /// drops the resolved dispute too. The clause is order-scoped rather than
@@ -1629,6 +1674,7 @@ pub async fn find_user_orders_by_master_key(
         return Err(MostroCantDo(CantDoReason::InvalidPubkey));
     }
 
+    let claimable_payout = has_claimable_bond_payout();
     let sql_query = format!(
         r#"
         SELECT id as order_id, trade_index_buyer as trade_index, status FROM orders 
@@ -1639,10 +1685,7 @@ pub async fn find_user_orders_by_master_key(
         WHERE (master_seller_pubkey = ?)
         AND (status NOT IN ({}) OR {})
         "#,
-        EXCLUDED_ORDER_STATUSES,
-        HAS_CLAIMABLE_BOND_PAYOUT,
-        EXCLUDED_ORDER_STATUSES,
-        HAS_CLAIMABLE_BOND_PAYOUT
+        EXCLUDED_ORDER_STATUSES, claimable_payout, EXCLUDED_ORDER_STATUSES, claimable_payout
     );
     let orders = sqlx::query_as::<_, RestoredOrdersInfo>(AssertSqlSafe(sql_query))
         .bind(master_key)
@@ -1883,7 +1926,7 @@ mod tests {
         .await?;
 
         // `bonds` subset: the terminal-status filters join against it via
-        // `HAS_CLAIMABLE_BOND_PAYOUT`, so the table has to exist here even
+        // `has_claimable_bond_payout`, so the table has to exist here even
         // for the tests that never insert a row. Only the columns those
         // queries and their fixtures touch, mirroring the real migration's
         // NOT NULL / default shape.
@@ -1911,20 +1954,45 @@ mod tests {
         Ok(pool)
     }
 
+    /// Default claim window, in seconds: `Settings::get_bond()` is unset in
+    /// unit tests, so `has_claimable_bond_payout` falls back to 15 days.
+    const TEST_CLAIM_WINDOW_SECS: i64 = super::DEFAULT_CLAIM_WINDOW_SECONDS;
+
+    /// A `slashed_at` comfortably inside the claim window.
+    fn slashed_at_within_window() -> i64 {
+        chrono::Utc::now().timestamp() - 3_600
+    }
+
+    /// A `slashed_at` comfortably outside it.
+    fn slashed_at_outside_window() -> i64 {
+        chrono::Utc::now().timestamp() - TEST_CLAIM_WINDOW_SECS - 3_600
+    }
+
     /// Attach a bond row in `state` to `order_id`, for the payout-exception
     /// arms of the terminal-status filters.
-    async fn insert_bond(pool: &SqlitePool, order_id: uuid::Uuid, pubkey: &str, state: &str) {
+    ///
+    /// `slashed_at` is a parameter because the `'failed'` arm is bounded by
+    /// it: a fixed timestamp would silently place every fixture outside the
+    /// claim window and make the bounded arm untestable.
+    async fn insert_bond(
+        pool: &SqlitePool,
+        order_id: uuid::Uuid,
+        pubkey: &str,
+        state: &str,
+        slashed_at: i64,
+    ) {
         sqlx::query(
             r#"
             INSERT INTO bonds (id, order_id, pubkey, role, amount_sats, state,
                                slashed_reason, slashed_at, created_at)
-            VALUES (?1, ?2, ?3, 'taker', 10000, ?4, 'lost-dispute', 1700000000, 1700000000)
+            VALUES (?1, ?2, ?3, 'taker', 10000, ?4, 'lost-dispute', ?5, 1700000000)
             "#,
         )
         .bind(uuid::Uuid::new_v4())
         .bind(order_id)
         .bind(pubkey)
         .bind(state)
+        .bind(slashed_at)
         .execute(pool)
         .await
         .unwrap();
@@ -2312,7 +2380,7 @@ mod tests {
         );
     }
 
-    /// The two states [`HAS_CLAIMABLE_BOND_PAYOUT`] names are SQL literals,
+    /// The two states [`has_claimable_bond_payout`] names are SQL literals,
     /// so a rename on `BondState` would leave them silently matching nothing
     /// — the same failure mode as the unhyphenated status constants this
     /// change fixes. Pin them against the enum's own serialization.
@@ -2321,7 +2389,7 @@ mod tests {
         for state in [BondState::PendingPayout, BondState::Failed] {
             let literal = format!("'{state}'");
             assert!(
-                super::HAS_CLAIMABLE_BOND_PAYOUT.contains(&literal),
+                super::has_claimable_bond_payout().contains(&literal),
                 "{literal} missing from the claimable-payout clause"
             );
         }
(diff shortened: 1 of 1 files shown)


ToRyVand (approved):
Reviewed and verified independently against `main` — everything in the description holds up. Details of what I checked rather than took on faith:

**The premise.** Confirmed `Status`' `Display` impl in mostro-core **0.14.5** (the version this repo pins in `Cargo.toml`) does emit the hyphenated forms — `canceled-by-admin`, `settled-by-admin`, `completed-by-admin`, `cooperatively-canceled` — while both constants on `main` carry the unhyphenated spellings. Real mismatch.

**That it's a typo, not an alternative encoding.** `ACTIVE_DISPUTE_STATUSES` (`'in-progress'`) and `PRETRADE_STATUSES` (`'waiting-taker-bond'`) in the same file are hyphenated correctly, which is the detail that settles it.

**Reproduced the failure.** Reverted `TERMINAL_ORDER_STATUSES` to the pre-fix spelling on your branch and ran `find_active_trade_pubkeys_covers_active_and_disputed_excludes_terminal` — fails, exactly as you documented. Restored the fix — passes. Same for the new restore-session test.

**Completeness.** Grepped all of `src/` for any remaining unhyphenated occurrences of the four statuses: none. No other call site reintroduces the bug, and both constants feed the two functions you name (`find_user_orders_by_master_key` at the `format!` in `db.rs`, and `find_active_trade_pubkeys`).

**Checks.** Full suite 1187 passed / 0 failed, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

One thing worth calling out as reviewer rather than as a request: flagging the #784 interaction up front — that this removes an accidental mask over half of it rather than creating anything — is the right call and saved me from having to work that out. Scoping the `bonds` state machine fix out of here is correct.

Note I'm a contributor, not a maintainer, so treat this as a technical second opinion for @grunch rather than a merge signal.

AndreaDiazCorreia (approved):
tACK

AndreaDiazCorreia:
Reviewed and verified locally. The fix is correct and the tests are genuine regression tests — I reverted only the two constants while keeping the new tests, and both fail with exactly the output described; restoring the constants makes them pass.

Since the branch is based a few commits before 0.18.5, I re-ran everything against current `main` (`a7249cc`) with the patch applied rather than against the branch base:

```
cargo test --all-targets                   1223 passed; 0 failed; 2 ignored
cargo clippy --all-targets -- -D warnings  clean
cargo fmt --check                          clean
```

The patch applies to current `main` without conflict, and `src/db.rs` was untouched by the commits in between.

I also grepped `src/`, `migrations/` and the docs for any remaining unhyphenated occurrence of the four statuses: the two constants in this diff were the only ones. Worth stating explicitly for whoever merges this — since nothing ever *wrote* those spellings, no rows carry them, so this needs no data migration. The bug was read-side only.

One nice side effect: `TERMINAL_ORDER_STATUSES` in `src/db.rs` now agrees with the typed `[Status; 7]` array in `src/app/bond/db.rs:240`. Those two had silently drifted apart for as long as this bug existed.

## Follow-ups

Opened separately so they do not expand this PR's scope:

- #897 — the root cause is that these lists are hand-written SQL literals duplicating `Status`' `Display`, so a typo compiles, passes clippy, and silently stops filtering. `src/app/bond/db.rs` already does it right (typed array, bound parameters). Prevention, not live breakage.
- #898 — `find_user_orders_by_master_key` has no `LIMIT` while `src/app/orders.rs:26` rejects the whole batch above `max_orders_per_response` (default 10). This PR removes the dead-order inflation that made the cap easy to hit, but a user with more than 10 genuinely live orders still cannot refresh.

## Before merging

Worth linking #784 from here. As the description says, this removes an accidental mask rather than creating a problem — but the practical effect is that the admin-cancel branch of #784 starts reproducing once this lands. Without the link that will read as a fresh regression to whoever hits it.

Catrya (dismissed):
ACK

src/db.rs
```diff
@@ -13,7 +13,7 @@ use std::sync::Arc;
 use uuid::Uuid;
 
 // Constants for status filtering used across restore session functions
-const EXCLUDED_ORDER_STATUSES: &str = "'expired','success','canceled','dispute','canceledbyadmin','completedbyadmin','settledbyadmin','cooperativelycanceled'";
+const EXCLUDED_ORDER_STATUSES: &str = "'expired','success','canceled','dispute','canceled-by-admin','completed-by-admin','settled-by-admin','cooperatively-canceled'";
```
grunch, line 16: [P1] Preserve restore context for outstanding bond payouts

After admin_cancel_action resolves a dispute, the order is canceled-by-admin and the dispute is seller-refunded, while apply_bond_resolution can leave a slashed bond in PendingPayout awaiting the winner's invoice. With this corrected exclusion, find_user_orders_by_master_key now drops that order; find_user_disputes_by_master_key also excludes the resolved dispute, and process_restore_session_work does not query bonds. A restoring client therefore loses the order/trade-index context needed to claim the payout. If no invoice is submitted before the claim deadline, process_one_bond forfeits the bond.

This is the admin-cancel branch of #784, but it is a behavior change introduced here: before this edit, canceled-by-admin orders were returned. Please preserve terminal orders with recoverable payouts owed to the restoring user, or land the bond-aware restore fix before enabling this exclusion. Add coverage for an admin-canceled order with an outstanding payout, alongside the terminal-order-with-no-obligation case.

src/db.rs
```diff
@@ -24,7 +24,7 @@ const ACTIVE_DISPUTE_STATUSES: &str = "'initiated','in-progress'";
 /// disputed order is still active (buyer, seller and the assigned solver keep
 /// messaging), so its trade keys must stay fast-pathed. See
 /// `find_active_trade_pubkeys` and docs/TRANSPORT_V2_SPEC.md §6 Phase 2.
-const TERMINAL_ORDER_STATUSES: &str = "'expired','success','canceled','canceledbyadmin','completedbyadmin','settledbyadmin','cooperativelycanceled'";
+const TERMINAL_ORDER_STATUSES: &str = "'expired','success','canceled','canceled-by-admin','completed-by-admin','settled-by-admin','cooperatively-canceled'";
```
grunch, line 27: [P2] Keep pending payout recipients in the known-key cache

A canceled-by-admin order can still require a legitimate AddBondInvoice from the winning counterparty. After this change, find_active_trade_pubkeys removes that recipient's trade key on the next cache refresh if they have no other active order. On protocol v2 nodes configured with pow_first_contact > pow, accept_event then drops a response carrying the normal trade PoW before decrypting or dispatching it to add_bond_invoice_action. This can interrupt the payout even when the client has retained its session and order context.

Please include the recipient keys of recoverable bond payouts in the known-key query until their obligations are resolved. This needs coverage independently of restore: a terminal order with a pending payout should retain its recipient key and allow the normal-PoW response.

grunch (changes requested):
The kebab-case correction is valid, but these exclusions also remove orders and trade keys while bond payout obligations can still be outstanding. Please address the two cases below before merging.

The interaction with #784 is acknowledged in the description, but excluding canceled-by-admin introduces an observable regression for the admin-cancel recovery path: that order was previously returned by restore. Fixing restore alone also does not preserve the recipient's eligibility for the normal-PoW lane.

Validation on 4d17c8d: the restore exclusion test and both find_active_trade_pubkeys tests passed (3 tests total); cargo fmt --check passed. These tests do not cover terminal orders with outstanding bond payouts. Findings are based on tracing the code paths; I did not run the full suite or an end-to-end LND reproduction.

src/db.rs
21Mill: Fixed in 77d7beb.

Both filters now carry an exception (`HAS_CLAIMABLE_BOND_PAYOUT`, `src/db.rs:33`): a terminal order survives while a bond on it is `pending-payout` **or** `failed`. `failed` belongs there because `apply_invoice` still accepts a payout invoice against such a row inside the `payout_claim_window_days` window, so the claim is live even after `send_payment` gave up. Both states are left for good on payout, slash or forfeit, so the exception retracts itself with no code to maintain.

Restore therefore keeps the `canceled-by-admin` order and its `trade_index_buyer`/`trade_index_seller`, which is what the claim needs.

One deliberate limitation, worth your call: the clause is **order-scoped, not recipient-scoped**, so both parties see such an order rather than only the winner. Narrowing it in SQL is not really available — `bonds.pubkey` is a trade key while the restore query joins on master keys, and the Phase 6 maker-refund row (`parent_bond_id` set, `child_order_id` NULL) pays `bond.pubkey` itself rather than the counterparty, so a "not the bonded side" filter would be wrong exactly where the money is owed to the bonded side. I took the over-inclusive side; say the word if you would rather have the extra complexity.

Coverage as requested: `find_user_orders_by_master_key_keeps_terminal_orders_owing_a_payout` pins all four arms — terminal with no obligation (excluded), `pending-payout` (kept), `failed` (kept), and `slashed` (excluded, the obligation is discharged). Verified failing with the exception disabled and passing with it.

src/db.rs
21Mill: Fixed in 77d7beb, same clause applied to the order query in `find_active_trade_pubkeys` (`src/db.rs:63`).

You were right to file this separately from the restore case — it is the sharper of the two. The restore gap costs a client its context; this one strands the payout even for a client that never lost its session, because the reply is dropped before decryption and nothing on either side can tell that it happened. Keying the exception on `b.order_id = orders.id` matches what `resolve_payout_recipient` actually loads, including the Phase 6 slice-slash child rows whose `order_id` is the slice order, so no special case was needed.

Covered independently of restore by `find_active_trade_pubkeys_keeps_keys_of_terminal_orders_owing_a_payout`: a `canceled-by-admin` order with a `pending-payout` bond retains all three participant keys, while an identical order with nothing outstanding drops out as before.

Two notes on the diff beyond the two fixes:

- `setup_orders_db` builds its schema by hand and had no `bonds` table, so the new `EXISTS` broke three pre-existing tests with `no such table: bonds`. I added a `bonds` subset to the fixture plus an `insert_bond` helper — that is most of the line count.
- The clause embeds `pending-payout` and `failed` as SQL literals, which is the very hazard #897 describes and this PR was born from. `claimable_bond_states_match_bond_state_display` pins them against `BondState`'s own `Display`, so a rename fails the build instead of silently matching nothing.

Full suite 1190 passed / 0 failed / 2 ignored; `cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` clean. Both new tests verified failing with the exception disabled.

src/db.rs
```diff
@@ -24,7 +24,21 @@ const ACTIVE_DISPUTE_STATUSES: &str = "'initiated','in-progress'";
...
+
+/// Exception carved out of both terminal-status filters: a closed order can
+/// still owe a bond payout to the winning counterparty, so it stays visible
+/// until that obligation resolves.
+///
+/// `'failed'` belongs beside `'pending-payout'`: a `Failed` row still accepts
+/// a payout invoice while it is inside the `payout_claim_window_days` window
+/// (`app::bond::payout::apply_invoice`), so the claim is live even though the
+/// last `send_payment` gave up. Both states are left for good on payout,
+/// forfeit or slash, so the exception expires without any code to retract it.
```
grunch: **[P1] The `'failed'` arm never expires — a `Failed` bond past the claim window stays `Failed` for good, so this exception is permanent for those rows.**

The doc says *"Both states are left for good on payout, forfeit or slash, so the exception expires without any code to retract it."* That holds for `pending-payout` (`process_one_bond` forfeits it when the window elapses) but not for `failed`. `BondState`'s own docs (`src/app/bond/types.rs:61-63`, `:87-90`):

> Past the claim window, the row **stays `Failed`** and requires operator attention to pay the counterparty manually.

And the scheduler only enumerates `pending-payout` (`payout.rs:1553`: *"Failed is invisible to it"*), so nothing ever moves such a row. Consequences, for every `Failed` bond the operator has not manually handled:

- `find_active_trade_pubkeys` keeps all three participant keys of that closed order in the fast-path set **indefinitely** — an unbounded exception to the very invariant `TERMINAL_ORDER_STATUSES` documents.
- `find_user_orders_by_master_key` hands the dead order back on every restore, forever, and — per the description's own finding — each such row eats into `max_orders_per_response`.
- `apply_invoice` would *reject* an invoice on that row anyway (`payout.rs:1547-1549`), so the key is retained to answer a nudge the node would refuse.

The clause should mirror the predicate `apply_invoice` actually enforces, `now - slashed_at < claim_window`, with the same `Settings::get_bond()` / 15-day fallback that `process_one_bond` uses (`payout.rs:198-201`):

```rust
/// … `'failed'` counts only while `apply_invoice` would still accept an
/// invoice on it: `slashed_at` within `payout_claim_window_days`. Past
/// that, the row stays `Failed` for operator attention and no longer keeps
/// the order — or its keys — alive.
const HAS_CLAIMABLE_BOND_PAYOUT: &str = "EXISTS (SELECT 1 FROM bonds b \
     WHERE b.order_id = orders.id AND (b.state = 'pending-payout' \
        OR (b.state = 'failed' AND b.slashed_at IS NOT NULL AND b.slashed_at > ?)))";

fn claim_window_cutoff() -> i64 {
    let window = Settings::get_bond()
        .map(|c| c.payout_claim_window_days as i64 * 86_400)
        .unwrap_or(15 * 86_400);
    chrono::Utc::now().timestamp() - window
}
```

then `.bind(claim_window_cutoff())` once in `find_active_trade_pubkeys` and once per `UNION ALL` arm in the restore query (the master key is already bound twice there, so the shape is familiar). `slashed_at IS NOT NULL` matches `apply_invoice`'s "a `None` here is an invariant violation; reject" branch.

Please add the arm to the two tests: a `failed` bond with `slashed_at` inside the window is kept, one with `slashed_at` outside it is excluded — and note the fixture's hard-coded `slashed_at = 1700000000` (Nov 2023) is already outside any window, so the current "`failed` → kept" assertion would start failing with the bound in place; give `insert_bond` a `slashed_at` parameter.

src/db.rs
```diff
@@ -2131,6 +2253,200 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        let mut ids: Vec<uuid::Uuid> = orders.iter().map(|o| o.order_id).collect();
+        ids.sort();
+        let mut expected = vec![owed, failed];
+        expected.sort();
+        assert_eq!(
+            ids, expected,
+            "only the terminal orders with an open payout may be restored"
```
grunch, line 2497: **[P2] Restore drops the order's dispute while keeping the order — and that is fine, but say so where the test can hold it.**

With this clause the winner gets the `canceled-by-admin` order and its trade index back, while `find_user_disputes_by_master_key` (unchanged) still drops the resolved dispute. That is the right split — claiming the payout needs the order and trade index, not the dispute — and the function doc records it. I would like `find_user_orders_by_master_key_keeps_terminal_orders_owing_a_payout` to also assert the returned `status` is `canceled-by-admin`, so a client receiving it can tell it is a closed order with a claim attached rather than a live trade, and nobody later "fixes" restore by rewriting the status.

On the order-scoped-vs-recipient-scoped question you raised: I agree with over-inclusive. The loser seeing their own closed order in restore is harmless; missing the Phase 6 maker-refund row would not be.

grunch (changes requested):
## Review of `77d7beb` — strict pass, verified on the merged tree

**Verdict: changes requested**, one blocking item. Round 1's two findings are genuinely fixed and the tests for them are the right tests. What remains is that the new exception has an arm that never closes.

### What I verified (locally — no CI ran on this head, see below)
- **Merge with current `main`** (`mostro-core 0.14.6`) is clean.
- **Premise still holds on 0.14.6**: `Status`'s `Display` emits the hyphenated forms; `ACTIVE_DISPUTE_STATUSES` / `PRETRADE_STATUSES` in the same file were already hyphenated. Grepped `src/` for the four unhyphenated spellings: none left.
- **`HAS_CLAIMABLE_BOND_PAYOUT` literals** match `BondState::Display` (`pending-payout`, `failed`), and `claimable_bond_states_match_bond_state_display` pins them.
- **`apply_invoice` accepts `Failed` only inside the window** (`payout.rs:1547-1549`) — the premise for including `'failed'` is right; the bound is what is missing (P1).
- **Keys retained are the right keys**: `resolve_payout_recipient` reads the same `orders` columns the query selects, including the Phase 6 slice rows keyed by `b.order_id`.
- **Schema**: `bonds.order_id` is indexed (`idx_bonds_order_id`), so the correlated `EXISTS` is cheap on cache refresh; `slashed_at integer` exists for the bound proposed in P1.
- **SQL shape**: `WHERE (master = ?) AND (status NOT IN (…) OR EXISTS …)` — precedence is correct on both `UNION ALL` arms; the `find_active_trade_pubkeys` form has no `AND` to fight.
- **Round-1 fixes**: restore keeps the `canceled-by-admin` order with `trade_index = 7`; the known-key set keeps `creator/buyer/seller_owed` and drops `*_done`. Both new tests pass here and their intent matches the P1/P2 I filed.
- **Local run on the merged tree**: `cargo fmt --check` clean, `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo test` **1327 passed / 0 failed / 2 ignored** (merged with `main` at `mostro-core 0.14.6`).

### Findings
| # | Severity | Where | Summary |
|---|---|---|---|
| 1 | 🔴 P1 | `db.rs:36-41` | A `Failed` bond past `payout_claim_window_days` stays `Failed` forever (`types.rs:61`, `:87`; the scheduler never touches it), so the exception is permanent for those rows: keys stay fast-pathed and the dead order is restored indefinitely. Bound the `'failed'` arm by `slashed_at` the way `apply_invoice` does. |
| 2 | 🟡 P2 | `db.rs:2320` (test) | Assert the restored row's `status` so the "closed order with a claim attached" contract is pinned; dropping the resolved dispute alongside is correct. |

### Not blocking, for the record
- **`ci.yml` triggers on `push` only**, so a fork PR never gets build/test/clippy in CI — the only check runs on `77d7beb` are the path-filtered Cashu and Mutation workflows, both skipped. That is a repo-level gap, not this PR's; I ran the suite locally instead. Worth a separate PR adding `pull_request` to `ci.yml`.
- The order-scoped (not recipient-scoped) exception is the right trade-off; noted inline.
- `#784` framing in the description is still accurate after round 1: the admin-cancel branch now reproduces, and this PR now also keeps the context needed to act on it. The `bonds` state-machine fix stays out of scope.

ToRyVand:
Not a review — @grunch's P1 checks out (I verified the citations, and `b.slashed_at > (now - window)` is the exact complement of what `apply_invoice` enforces). Two small things.

**The reason for leaving the `pending-payout` arm unbounded is incomplete, though the asymmetry itself is right.** `process_one_bond`'s forfeit is gated on `payout_invoice.is_none()` (`payout.rs:230`), and `on_send_payment_failure` deliberately *keeps* the invoice when the retry budget is exhausted on an `Indeterminate` failure (`payout.rs:919`). That row never forfeits — and `apply_payout_invoice`'s `PendingPayout` CAS (`... AND payout_invoice IS NULL`, `payout.rs:1415`) would refuse an invoice on it anyway, which is the P1's own third bullet.

It should still stay unbounded: that payout is genuinely in flight and resolves once LND answers definitively, so bounding it on `slashed_at` would strand a live claim. Worth writing *that* as the reason in the doc comment, or the next person bounds both arms for symmetry and breaks the one that matters.

Separately, the `ci.yml` `on: push` gap in the non-blocking notes is already filed as #929, with the PR offered there.

And for the record: my Aug 17 approval verified that the typo fix worked, and never asked whether the exclusion was correct in the first place. Round 1 was the finding I should have made.

grunch:
@ToRyVand — agreed, and thanks for checking the citations. I re-verified the three: the forfeit is gated on `payout_invoice.is_none()` (`payout.rs:230`), an `Indeterminate` exhaustion deliberately keeps the invoice (`payout.rs:919`), and the `PendingPayout` CAS requires `payout_invoice IS NULL` (`payout.rs:1415`). So a `pending-payout` row with a retained invoice never forfeits either — my "the scheduler forfeits it" reasoning for that arm was incomplete.

The conclusion stands, and it is the asymmetry that matters: that row is a payout genuinely in flight, resolved by LND's definitive answer on a later tick, so bounding it on `slashed_at` would strand a live claim. Only the `failed` arm is bounded, because only there does `apply_invoice` itself stop accepting anything past the window.

@21Mill — consolidated ask for the next round, so it lives in one place:

1. **Bound the `'failed'` arm** by `slashed_at` within `payout_claim_window_days` (P1 above, `Settings::get_bond()` with the 15-day fallback that `process_one_bond` uses). Leave `'pending-payout'` unbounded.
2. **Write the real reason for the asymmetry in the doc comment**, roughly: *`pending-payout` stays unbounded because a row whose retry budget ran out on an indeterminate LND failure keeps its invoice and never forfeits — the payout is in flight and resolves when LND answers; bounding it would strand a live claim. `failed` is bounded because `apply_invoice` refuses an invoice on it past the window.* That keeps the next person from bounding both arms for symmetry.
3. **Tests**: `failed` inside the window → kept; `failed` outside → excluded; `pending-payout` with `slashed_at` outside the window → still kept (pins the asymmetry). `insert_bond` needs a `slashed_at` parameter — its hard-coded `1700000000` is already outside any window.
4. **P2**: assert the restored row's `status` in `find_user_orders_by_master_key_keeps_terminal_orders_owing_a_payout`.

CI gap tracked in #929; nothing to do here.

21Mill:
Round 3 pushed. All four items addressed, plus two extra tests covering gaps I found while writing them.

@ToRyVand — the reason for the asymmetry in the doc comment is yours, not mine; I verified the three citations against the code rather than taking them from the thread, and they hold:

- `payout.rs` — the forfeit is gated on `bond.payout_invoice.is_none() && now - slashed_at >= claim_window_seconds`.
- `payout.rs` — `on_send_payment_failure` returns early on `PaymentFailureKind::Indeterminate` once the budget is exhausted, keeping invoice and hash so `pay_counterparty` can reconcile.
- `payout.rs` — the `PendingPayout` arm of `apply_payout_invoice` CASes on `AND payout_invoice IS NULL`.

I also checked the premise under P1 independently: `BondState::is_terminal` includes `Failed`, and its only exit is the `add_bond_invoice_action` resurrection, which is itself window-bounded. So past the window a `Failed` row never leaves `Failed`, and the exception really would have been permanent.

## What changed

**1. The `'failed'` arm is bounded; `'pending-payout'` is not.**

`HAS_CLAIMABLE_BOND_PAYOUT` could not stay a `const` — the bound needs `Settings::get_bond()` and a clock — so it is now `has_claimable_bond_payout()`, which reads the window (with the 15-day fallback `process_one_bond` uses) and delegates to `has_claimable_bond_payout_at(now, claim_window_seconds)`:

```sql
EXISTS (SELECT 1 FROM bonds b WHERE b.order_id = orders.id AND (
  b.state = 'pending-payout'
  OR (b.state = 'failed' AND b.slashed_at IS NOT NULL AND b.slashed_at > {cutoff})))
```

`cutoff = now - claim_window_seconds`, and the comparison is strict `>`, the exact complement of `apply_payout_invoice`'s `now - slashed_at >= claim_window_seconds` rejection. A `Failed` row with a `NULL` `slashed_at` is excluded, matching that function's conservative `Rejected` on the same invariant violation — without the `IS NOT NULL` a `NULL` would simply drop out of the `>` and read as "unbounded".

Splitting the builder in two is what makes the boundary testable at all: with `Utc::now()` inside it, `>` and `>=` differ only when the two clock readings land in the same second, so a wall-clock boundary test would have been flaky and would have let that mutation through on most runs.

**2. The real reason for the asymmetry is in the doc comment**, with an explicit "do not tidy this into a single bound" note, phrased as you both landed on: `pending-payout` stays unbounded because a row whose retry budget ran out on an indeterminate LND failure keeps its invoice and never forfeits — that payout is in flight and resolves when LND answers, so bounding it would strand a live claim. `failed` is bounded because `apply_invoice` itself stops accepting anything past the window.

**3. Tests.** `insert_bond` takes a `slashed_at` parameter now — the hard-coded `1700000000` put every fixture outside the window and would have made the bounded arm untestable.

- `find_user_orders_by_master_key_bounds_only_the_failed_payout_arm` — `failed` inside the window kept, outside excluded, `pending-payout` with an old `slashed_at` still kept. That last one is the assertion that pins the asymmetry.
- `find_active_trade_pubkeys_bounds_only_the_failed_payout_arm` — the same three cases for the known-key set. Not asked for, but the bound has to hold on both queries or an expired claim keeps trade keys fast-pathed forever.
- `failed_bond_without_slashed_at_does_not_keep_its_order` — the `NULL` anchor case.
- `claim_window_boundary_excludes_a_bond_slashed_exactly_one_window_ago` — the edge is exclusive, one second inside is not.
- `claim_window_bound_follows_the_configured_window` — the same bond, 20 days old, is excluded under a 15-day window and kept under a configured 30-day one. Unit tests cannot set `MOSTRO_CONFIG` safely (process-wide `OnceCell` shared with the rest of the suite), so this pins the window→SQL mapping directly.
- `claim_window_fallback_matches_settings_default` — pins `DEFAULT_CLAIM_WINDOW_SECONDS` against `AntiAbuseBondSettings::default()`. A hand-written literal drifting from the config default is precisely the failure mode this PR exists to fix.

**4. P2** — `find_user_orders_by_master_key_keeps_terminal_orders_owing_a_payout` now asserts the restored rows' `status`, so a future change cannot satisfy the id assertion by quietly keeping those orders non-terminal.

## Verification

Ran on **`77d7beb`** (this PR's head, so the diff you review is the tree I tested) and again on **current `main` (0.18.7)**:

| | on `77d7beb` | on `main` 0.18.7 |
|---|---|---|
| `cargo test` | 1196 passed / 0 failed / 2 ignored | 1333 / 0 / 2 |
| `cargo clippy --all-targets --all-features -- -D warnings` | clean | clean |
| `cargo fmt --check` | clean | clean |

Every new test was mutation-checked — each one fails, and only it, when the thing it covers is removed:

| Mutation | Result |
|---|---|
| drop the `slashed_at` bound entirely | the 3 arm/NULL tests fail, 1327 pass |
| `>` → `>=` | only the boundary test fails, 1332 pass |
| ignore the injected window, always use 15 days | only the configured-window test fails, 1332 pass |

## What I did *not* verify

- **`Settings::get_bond()`'s own read.** The three lines mapping config to seconds are not covered by any test I added, for the `OnceCell` reason above. The TOML parse itself is already covered by `config::types::anti_abuse_bond_tests::toml_slash_node_share_pct_and_claim_window_override`; what is uncovered is the `.map(|c| c.payout_claim_window_days as i64 * 86_400)` in between, which mirrors `process_one_bond` and `apply_invoice` verbatim.
- **No end-to-end LND reproduction** — no live node here, so like your round 2 this rests on tracing the code paths plus the suite.

CI still will not run on this (`ci.yml` is `on: push` only, tracked in #929), so the numbers above are local.

21Mill:
@coderabbitai good catch, and it is worth more than the "trivial / low value" it was filed under. Applied.

The mechanism is real: `sqlx-core`'s statement cache is an `LruCache<String, _>` keyed by the SQL text (`common/statement_cache.rs`), with a default capacity of 100 (`sqlx-sqlite/src/options/mod.rs`). But the cost is not just that this query misses its own cache entry — it is that it **evicts everyone else's**. `job_refresh_active_pubkeys` calls `find_active_trade_pubkeys` every `active_pubkeys_refresh_interval` (default 60s), and my interpolated cutoff changed every second, so each tick inserted a brand-new key. That turns the whole 100-entry cache over roughly every 100 minutes, forcing every other statement in the daemon to be re-prepared, forever. A slow continuous leak, invisible at runtime — and one I introduced in `97f254e`; before it the clause was a `const` and cached once.

## What changed

The cutoff is a bound parameter, so the clause goes back to being a `const` with stable text:

```rust
const HAS_CLAIMABLE_BOND_PAYOUT: &str = "EXISTS (SELECT 1 FROM bonds b \
     WHERE b.order_id = orders.id AND (\
       b.state = 'pending-payout' \
       OR (b.state = 'failed' AND b.slashed_at IS NOT NULL AND b.slashed_at > ?)))";
```

`has_claimable_bond_payout_at(now, window) -> String` becomes `claim_window_cutoff_at(now, window) -> i64`, keeping the injection point the boundary test needs. Bind order is preserved as you asked: one `?` in `find_active_trade_pubkeys`, and in `find_user_orders_by_master_key` the placeholders interleave — `master_key, cutoff, master_key, cutoff` — because SQLite binds positional `?` in order of appearance and each `UNION ALL` arm is `master_*_pubkey = ?` followed by the clause. The cutoff is read once so both arms filter on the same instant.

Side benefit: this removes the last runtime value interpolated into SQL in the file, so the doc-comment paragraph arguing why that interpolation was safe is gone. The remaining interpolations are the compile-time status constants.

Two new pins, since this is cheap to undo by accident:

- `claim_window_cutoff_is_a_bound_parameter` — asserts the clause ends in `> ?` and carries exactly one placeholder, with the cache-eviction reason in the doc comment so the next person does not "simplify" it back.
- The bind order is covered by the existing restore tests rather than by inspection — see the mutation table below.

## Verification

| | on `0d7c137` (this head) | on `main` 0.18.7 |
|---|---|---|
| `cargo test` | 1197 passed / 0 failed / 2 ignored | 1334 / 0 / 2 |
| `cargo clippy --all-targets --all-features -- -D warnings` | clean | clean |
| `cargo fmt --check` | clean | clean |

Mutation-checked, as with the previous round:

| Mutation | Result |
|---|---|
| bind `master_key, master_key, cutoff, cutoff` (the plausible wrong order) | 3 restore tests fail, 1194 pass |

@grunch — the round-3 findings are unchanged by this; it is the same predicate, only parameterised.

grunch (approved):
## Round 3 (`97f254e`) — approved

All four items are addressed the way they were asked, and the extra tests close gaps I had not asked about.

### Verified on the merged tree (`main` @ 0.18.7, no CI runs on fork PRs — #929)
- `cargo fmt --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean.
- `cargo test`: **1333 passed / 0 failed / 2 ignored**, the six new tests included.
- **Mutation check reproduced**: flipping `b.slashed_at > {cutoff}` to `>=` fails exactly one test, `claim_window_boundary_excludes_a_bond_slashed_exactly_one_window_ago`, and nothing else. The injected-`now` split is what makes that boundary deterministic; good call.
- **The bound is the exact complement of `apply_payout_invoice`**: it rejects on `now - slashed_at >= window`; the clause keeps on `slashed_at > now - window`. `NULL slashed_at` is excluded on both sides.
- **The asymmetry is now stated for the right reason** — the retained-invoice / `Indeterminate` path that ToRyVand traced — and pinned by the `pending-payout`-outside-the-window arm in both `*_bounds_only_the_failed_payout_arm` tests.
- `DEFAULT_CLAIM_WINDOW_SECONDS` matches `default_payout_claim_window_days()` (15), and the test pins it.
- P2: the restore test now asserts the terminal `status` of the rows it keeps.

### Notes, none blocking
- The `cutoff` is interpolated rather than bound. It is a computed `i64` behind `AssertSqlSafe`, so no injection surface, and the doc says so — fine as is.
- The one thing genuinely untested is the three-line `Settings::get_bond()` → seconds mapping, for the `OnceCell` reason given. It mirrors `process_one_bond` verbatim; acceptable.

Thanks to both of you — this is a much better change than the typo fix it started as.

Issue #750, opened 2026-05-25
Bond payout: winner is never re-prompted for an invoice after payout payment fails


Catrya:
When a slashed bond is being paid out to the dispute winner, if the LN payment to the winner's invoice fails all retry attempts, the bond is marked `Failed` in the database and **the winner is never asked for a new invoice again**. The payout silently stalls and requires manual operator intervention.

  ## Where it happens

  `src/app/bond/payout.rs`, `on_send_payment_failure()`: after `payout_attempts >= payout_max_retries` (default 5), the bond transitions to `Failed` and only an ERROR is logged — no message is sent to the user.

  ## Why the user is never re-prompted

  1. The scheduler (`run_bond_payout_cycle`) only enumerates bonds in `PendingPayout`. A `Failed` bond is invisible to it, so `AddBondInvoice` is never re-sent.
  2. During the retry phase the bond keeps its `payout_invoice` set, so `process_one_bond` always routes to `pay_counterparty`  the 5 retries all hit the **same** invoice, never requesting a new one.

  A recovery path exists (the "resurrection" branch in `apply_payout_invoice` flips `Failed` → `PendingPayout` when a new invoice arrives within the claim window), but it only fires if the client spontaneously resends. Mostro never prompts and never notifies the user that the payment failed, so in practice it is unreachable.

 ## Expected behavior

  After the final payment attempt fails, Mostro should re-request an invoice from the winner (re-send `AddBondInvoice`), **without** resetting the claim-window deadline (anchored on `slashed_at`), and stop prompting once a valid invoice  is submitted.

PR #883, opened 2026-08-17
fix: non-blocking buyer payout with bounded send_payment waits

Catrya:
- `do_payment` now performs only bounded work inline — resolve invoice/LNURL,
  connect to LND, persist the payout claim — and returns once the payout is
  dispatched. The payment itself runs in a background task, so the event loop
  never waits on payment resolution.
- The payment status watcher starts *before* `send_payment` and runs
  concurrently with it, so a busy status stream can no longer stall the
  payment on a full channel.
- `send_payment` for the buyer payout and for the bond payout is bounded by a
  shared 75s timeout (`PAYOUT_SEND_PAYMENT_TIMEOUT`). LND stops launching new
  route attempts at 60s; past that point the stream is only kept open by an
  HTLC that has not resolved yet, which the sender cannot cancel anyway.
- Hitting the timeout is **not** treated as a payment failure: the claim
  marker is kept and the reconciliation job resolves the real outcome by
  payment hash (finalize on success, re-arm retry on failure). Bond payout
  timeouts route through the existing `Indeterminate` handling for the same
  reason.

## Why

`send_payment` consumes LND's payment stream until a terminal state. A
payment that takes long to resolve therefore pinned its calling task for as
long as the payment stayed unresolved — and for the buyer payout that task is
the event loop, so message processing degraded until resolution.

This builds directly on the payout state tracking work (idempotent claim +
reconciler): the claim persisted *before* dispatch is what makes backgrounding
safe. Whatever happens to the task — RPC error, timeout, process restart —
the payout can always be finished or failed by hash, never lost and never
paid twice.

Semantics change: `do_payment` returning `Ok(())` now means *dispatched*, not
*settled*. All three callers (`release_action`, `admin_settle`, scheduler
retry job) already ignore or merely log the result, so no caller behavior
changes.

## Commits

1. Resolve the buyer pubkey before claiming — a malformed order fails fast
   without leaving a marker for a payout that was never dispatched.
2. Dispatch the buyer payout off the event loop with a bounded
   `send_payment` (watcher-first, 75s timeout, timeout keeps the claim).
3. Bound the bond payout `send_payment` with the same timeout.

## Testing

- `cargo test` (1195 passed), `cargo fmt`, `cargo clippy` clean. The existing
  `do_payment` unit tests cover the inline (pre-claim) paths, which are
  preserved unchanged.
- Regtest, end to end:
  - a payout that stays unresolved past the timeout: the daemon keeps
    processing messages throughout, the claim is kept, and a warn is logged;
  - restart while the payout is unresolved: no duplicate dispatch, the
    reconciler picks the claim up and defers while the payment is in flight;
  - payment failure: retry bookkeeping runs, the buyer is re-prompted
    (`payment-failed` → `add-invoice`), and a fresh invoice completes the
    order (`Success`, `purchase-completed` + `rate`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved payout reliability when payment processing takes longer than expected.
  * Added safeguards for concurrent background payment processing.
  * Preserved incomplete payouts and payment details when results are uncertain, allowing reconciliation.
  * Added payment status checks to reduce duplicate payments and incorrect retry decisions.
  * Revalidated payout requests before sending to prevent stale or replaced requests from being processed.
  * Continued finalizing successful orders and retrying failed payments as expected.

### DIFF
--- a/src/app/bond/payout.rs
+++ b/src/app/bond/payout.rs
@@ -683,7 +683,7 @@ async fn pay_counterparty(
     let (tx, mut rx) = channel(100);
 
     let send_fut = timeout(
-        crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT,
+        crate::lightning::PAYOUT_SEND_PAYMENT_TIMEOUT,
         ln_client.send_payment(invoice, counterparty_share, tx),
     );
 
@@ -794,7 +794,7 @@ fn classify_send_verdict(
             PaymentFailureKind::Indeterminate,
             format!(
                 "send_payment reached no terminal state after {}s",
-                crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
+                crate::lightning::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
             ),
         ),
         Ok(Err(e)) => {
@@ -2491,7 +2491,7 @@ mod tests {
                 PaymentFailureKind::Indeterminate,
                 format!(
                     "send_payment reached no terminal state after {}s",
-                    crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
+                    crate::lightning::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
                 )
             )
         );
--- a/src/app/bond/payout.rs
+++ b/src/app/bond/payout.rs
@@ -664,10 +664,44 @@ async fn pay_counterparty(
 
     // send_payment. The helper caps the fee via `routing_fee_cap_sats`,
     // the same value persisted above as `payout_routing_fee_sats`.
+    // `send_payment` consumes LND's payment stream until a terminal state,
+    // which a locked-in but unresolved HTLC (hold invoice as the winner's
+    // payout invoice, or an HTLC stuck in route) can delay indefinitely —
+    // unbounded, that pins this scheduler task forever and, because the
+    // status channel is drained only after `send_payment` returns, a chatty
+    // stream could also deadlock on the full channel. Bound it with the same
+    // timeout as the buyer payout; LND stops launching route attempts at 60s,
+    // so past 75s only an unresolved HTLC keeps the stream open.
     let (tx, mut rx) = channel(100);
-    let send_outcome = ln_client
-        .send_payment(invoice, counterparty_share, tx)
-        .await;
+    let send_outcome = timeout(
+        crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT,
+        ln_client.send_payment(invoice, counterparty_share, tx),
+    )
+    .await;
+    let send_outcome = match send_outcome {
+        Ok(outcome) => outcome,
+        Err(_) => {
+            // Timed out without a terminal state. Dropping the future closes
+            // our side of the gRPC stream but does NOT cancel the payment: a
+            // locked-in HTLC cannot be cancelled by the sender and may still
+            // settle. Indeterminate keeps the invoice + hash so the
+            // reconciliation branch above resolves the real outcome on the
+            // next tick — never re-prompt the winner against a payment that
+            // may still succeed.
+            return on_send_payment_failure(
+                pool,
+                bond,
+                max_retries,
+                claim_window_seconds,
+                PaymentFailureKind::Indeterminate,
+                &format!(
+                    "send_payment reached no terminal state after {}s",
+                    crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
+                ),
+            )
+            .await;
+        }
+    };
     if let Err(e) = send_outcome {
         // The RPC call itself errored. We cannot be sure the payment did
         // not partially enter LND, so treat it as indeterminate: keep
--- a/src/app/bond/payout.rs
+++ b/src/app/bond/payout.rs
@@ -740,32 +740,55 @@ async fn pay_counterparty(
 
     let (send_outcome, (succeeded, stream_failure)) = tokio::join!(send_fut, drain_fut);
 
-    // A terminal verdict from the stream is the payment's actual outcome and
-    // takes priority over however the send future ended: a `Succeeded`
-    // delivered just before the timeout finalizes the slash immediately
-    // instead of deferring to reconciliation, and an explicit `Failed` is
-    // safe to act on regardless of the send-side result.
-    if succeeded {
-        return slash_after_success(pool, bond, counterparty_share).await;
+    match classify_send_verdict(send_outcome, succeeded, stream_failure) {
+        SendVerdict::Settled => slash_after_success(pool, bond, counterparty_share).await,
+        SendVerdict::Failure(kind, msg) => {
+            on_send_payment_failure(pool, bond, max_retries, claim_window_seconds, kind, &msg).await
+        }
     }
-    if let Some((PaymentFailureKind::Terminal, msg)) = &stream_failure {
-        return on_send_payment_failure(
-            pool,
-            bond,
-            max_retries,
-            claim_window_seconds,
-            PaymentFailureKind::Terminal,
-            msg,
-        )
-        .await;
-    }
+}
 
-    // No terminal verdict from the stream: classify why, most specific cause
-    // first. Every branch is indeterminate — the payment may still settle, so
-    // keep the invoice + hash for the reconciliation branch above and never
-    // re-prompt the winner against a payment that may still succeed. (A
-    // locked-in HTLC cannot be cancelled by the sender; dropping the send
-    // future on timeout closes our side of the gRPC stream only.)
+/// Combined verdict of a bounded `send_payment` and its concurrent status
+/// drain (see `pay_counterparty`).
+#[derive(Debug, PartialEq)]
+enum SendVerdict {
+    /// The stream reported `Succeeded`: the payment settled — finalize the
+    /// slash.
+    Settled,
+    /// No settlement: route through `on_send_payment_failure` with this kind
+    /// and cause.
+    Failure(PaymentFailureKind, String),
+}
+
+/// Classify the joint outcome of the bounded send future and the status
+/// drain into a single verdict.
+///
+/// A terminal verdict from the stream is the payment's actual outcome and
+/// takes priority over however the send future ended: a `Succeeded`
+/// delivered just before the timeout finalizes the slash immediately instead
+/// of deferring to reconciliation, and an explicit `Failed` is safe to act
+/// on regardless of the send-side result.
+///
+/// With no terminal verdict, classify by the most specific cause. Every such
+/// branch is indeterminate — the payment may still settle, so the caller
+/// keeps the invoice + hash for reconciliation and never re-prompts the
+/// winner against a payment that may still succeed. (A locked-in HTLC cannot
+/// be cancelled by the sender; dropping the send future on timeout closes
+/// our side of the gRPC stream only.)
+fn classify_send_verdict(
+    send_outcome: Result<Result<(), MostroError>, tokio::time::error::Elapsed>,
+    succeeded: bool,
+    stream_failure: Option<(PaymentFailureKind, String)>,
+) -> SendVerdict {
+    if succeeded {
+        return SendVerdict::Settled;
+    }
+    let stream_failure = match stream_failure {
+        Some((PaymentFailureKind::Terminal, msg)) => {
+            return SendVerdict::Failure(PaymentFailureKind::Terminal, msg);
+        }
+        other => other,
+    };
     let (kind, msg) = match send_outcome {
         Err(_) => (
             PaymentFailureKind::Indeterminate,
@@ -786,7 +809,7 @@ async fn pay_counterparty(
             "payment stream ended without terminal status".to_string(),
         )),
     };
-    on_send_payment_failure(pool, bond, max_retries, claim_window_seconds, kind, &msg).await
+    SendVerdict::Failure(kind, msg)
 }
 
 /// Flip a `PendingPayout` row to `Slashed` after a confirmed payment.
@@ -2418,6 +2441,108 @@ mod tests {
         assert_eq!(after.payout_attempts, 3);
         assert_eq!(after.payout_invoice.as_deref(), Some("lnbc1pINFLIGHT"));
         assert_eq!(after.payout_payment_hash.as_deref(), Some("cafebabe"));
+    }
+
+    /// Produce a real `tokio::time::error::Elapsed` (it has no public
+    /// constructor): a zero-duration timeout over a pending future.
+    async fn elapsed() -> tokio::time::error::Elapsed {
+        timeout(std::time::Duration::ZERO, std::future::pending::<()>())
+            .await
+            .unwrap_err()
+    }
+
+    #[tokio::test]
+    async fn classify_stream_succeeded_wins_over_send_timeout() {
+        // A Succeeded delivered just before the 75s cutoff is the payment's
+        // real outcome: finalize the slash immediately instead of deferring
+        // to reconciliation, no matter how the send future ended.
+        let verdict = classify_send_verdict(Err(elapsed().await), true, None);
+        assert_eq!(verdict, SendVerdict::Settled);
+    }
+
+    #[tokio::test]
+    async fn classify_stream_terminal_failed_maps_to_terminal() {
+        let verdict = classify_send_verdict(
+            Ok(Ok(())),
+            false,
+            Some((
+                PaymentFailureKind::Terminal,
+                "payment failed: reason 1".to_string(),
+            )),
+        );
+        assert_eq!(
+            verdict,
+            SendVerdict::Failure(
+                PaymentFailureKind::Terminal,
+                "payment failed: reason 1".to_string()
+            )
+        );
+    }
+
+    #[tokio::test]
+    async fn classify_send_timeout_is_indeterminate() {
+        // The Elapsed branch: dropping the send future does not cancel a
+        // locked-in HTLC, so the verdict must be Indeterminate (keep the
+        // invoice + hash for reconciliation), never Terminal.
+        let verdict = classify_send_verdict(Err(elapsed().await), false, None);
+        assert_eq!(
+            verdict,
+            SendVerdict::Failure(
+                PaymentFailureKind::Indeterminate,
+                format!(
+                    "send_payment reached no terminal state after {}s",
+                    crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
+                )
+            )
+        );
+    }
+
+    #[tokio::test]
+    async fn classify_send_rpc_error_is_indeterminate() {
+        let rpc_err = MostroInternalErr(ServiceError::LnPaymentError("boom".to_string()));
+        let verdict = classify_send_verdict(Ok(Err(rpc_err)), false, None);
+        match verdict {
+            SendVerdict::Failure(PaymentFailureKind::Indeterminate, msg) => {
+                assert!(
+                    msg.contains("boom"),
+                    "cause must carry the RPC error: {msg}"
+                );
+            }
+            other => panic!("expected indeterminate failure, got {other:?}"),
+        }
+    }
+
+    #[tokio::test]
+    async fn classify_stream_recv_timeout_is_indeterminate() {
+        // The drain's own recv-timeout (second line of defense) surfaces its
+        // message when the send side ended cleanly.
+        let verdict = classify_send_verdict(
+            Ok(Ok(())),
+            false,
+            Some((
+                PaymentFailureKind::Indeterminate,
+                "payment status stream timed out after 120s without a terminal update".to_string(),
+            )),
+        );
+        assert_eq!(
+            verdict,
+            SendVerdict::Failure(
+                PaymentFailureKind::Indeterminate,
+                "payment status stream timed out after 120s without a terminal update".to_string()
+            )
+        );
+    }
+
+    #[tokio::test]
+    async fn classify_stream_eof_is_indeterminate() {
+        let verdict = classify_send_verdict(Ok(Ok(())), false, None);
+        assert_eq!(
+            verdict,
+            SendVerdict::Failure(
+                PaymentFailureKind::Indeterminate,
+                "payment stream ended without terminal status".to_string()
+            )
+        );
     }
 
     #[tokio::test]
--- a/src/app/bond/payout.rs
+++ b/src/app/bond/payout.rs
@@ -79,15 +79,6 @@ use super::db::{find_bond_by_id, find_bonds_by_state};
 use super::model::Bond;
 use super::types::{BondSlashReason, BondState};
 
-/// Per-message ceiling for the `send_payment` status stream. LND
-/// streams periodic InFlight updates while a payment is routing; if no
-/// update lands inside this window the channel is treated as dead and
-/// the attempt is routed through `on_send_payment_failure`. Picked to
-/// be longer than the typical InFlight cadence (a few seconds) but
-/// short enough to keep a single bond from blocking a scheduler task
-/// indefinitely.
-const PAYMENT_STATUS_RECV_TIMEOUT: Duration = Duration::from_secs(120);
-
 /// One full pass over every bond in [`BondState::PendingPayout`].
 ///
 /// Mirror of `dev_fee::run_dev_fee_cycle`: each tick walks the work
@@ -688,53 +679,40 @@ async fn pay_counterparty(
     );
 
     // Collect the first terminal status from the stream. Mirrors
-    // dev_fee::send_dev_fee_payment; each recv stays bounded by
-    // `PAYMENT_STATUS_RECV_TIMEOUT` as a second line of defense, although the
-    // send-side timeout above already bounds the whole exchange. We track
-    // *why* the stream ended: only an explicit `PaymentStatus::Failed` is
-    // terminal. A timeout or clean EOF leaves the payment outcome unknown
-    // (it may still be in flight), so it is routed as `Indeterminate` —
-    // `on_send_payment_failure` then keeps the invoice + hash for
-    // reconciliation instead of re-prompting.
-    let drain_fut = async {
+    // dev_fee::send_dev_fee_payment. The drain is bounded transitively by the
+    // send-side timeout above: when the send future ends — normal return, RPC
+    // error, or dropped at the 75s bound — `tx` drops and `recv()` yields
+    // `None`. We track *why* the stream ended: only an explicit
+    // `PaymentStatus::Failed` is terminal. A clean EOF leaves the payment
+    // outcome unknown (it may still be in flight), so it is routed as
+    // `Indeterminate` — `on_send_payment_failure` then keeps the invoice +
+    // hash for reconciliation instead of re-prompting.
+    let drain_fut = async move {
         let mut succeeded = false;
         let mut failure: Option<(PaymentFailureKind, String)> = None;
-        loop {
-            match timeout(PAYMENT_STATUS_RECV_TIMEOUT, rx.recv()).await {
-                Err(_) => {
-                    failure = Some((
-                        PaymentFailureKind::Indeterminate,
-                        format!(
-                            "payment status stream timed out after {}s without a terminal update",
-                            PAYMENT_STATUS_RECV_TIMEOUT.as_secs()
-                        ),
-                    ));
-                    break;
-                }
-                Ok(None) => break,
-                Ok(Some(msg)) => {
-                    if let Ok(status) = PaymentStatus::try_from(msg.payment.status) {
-                        match status {
-                            PaymentStatus::Succeeded => {
-                                succeeded = true;
-                                break;
-                            }
-                            PaymentStatus::Failed => {
-                                failure = Some((
-                                    PaymentFailureKind::Terminal,
-                                    format!(
-                                        "payment failed: reason {}",
-                                        msg.payment.failure_reason
-                                    ),
-                                ));
-                                break;
-                            }
-                            _ => {}
-                        }
+        while let Some(msg) = rx.recv().await {
+            if let Ok(status) = PaymentStatus::try_from(msg.payment.status) {
+                match status {
+                    PaymentStatus::Succeeded => {
+                        succeeded = true;
+                        break;
                     }
+                    PaymentStatus::Failed => {
+                        failure = Some((
+                            PaymentFailureKind::Terminal,
+                            format!("payment failed: reason {}", msg.payment.failure_reason),
+                        ));
+                        break;
+                    }
+                    _ => {}
                 }
             }
         }
+        // Unblock a send that is still pushing updates into a channel we are
+        // done reading: dropping `rx` fails its next `listener.send`, so the
+        // send future returns immediately instead of riding out the 75s
+        // bound for a payment whose verdict we already hold.
+        drop(rx);
         (succeeded, failure)
     };
 
@@ -802,8 +780,8 @@ fn classify_send_verdict(
             // not partially enter LND.
             (PaymentFailureKind::Indeterminate, format!("{e}"))
         }
-        // EOF (or recv timeout) with no terminal status: the stream closed
-        // without telling us the outcome.
+        // EOF with no terminal status: the stream closed without telling us
+        // the outcome.
         Ok(Ok(())) => stream_failure.unwrap_or((
             PaymentFailureKind::Indeterminate,
             "payment stream ended without terminal status".to_string(),
@@ -2513,24 +2491,13 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn classify_stream_recv_timeout_is_indeterminate() {
-        // The drain's own recv-timeout (second line of defense) surfaces its
-        // message when the send side ended cleanly.
-        let verdict = classify_send_verdict(
-            Ok(Ok(())),
-            false,
-            Some((
-                PaymentFailureKind::Indeterminate,
-                "payment status stream timed out after 120s without a terminal update".to_string(),
-            )),
-        );
-        assert_eq!(
-            verdict,
-            SendVerdict::Failure(
-                PaymentFailureKind::Indeterminate,
-                "payment status stream timed out after 120s without a terminal update".to_string()
-            )
-        );
+    async fn classify_stream_succeeded_wins_over_send_error() {
+        // Once the drain sees Succeeded it drops `rx`, which fails the send's
+        // next `listener.send` — the resulting Ok(Err(..)) from the send
+        // future must not shadow the settled verdict.
+        let send_err = MostroInternalErr(ServiceError::LnNodeError("channel closed".to_string()));
+        let verdict = classify_send_verdict(Ok(Err(send_err)), true, None);
+        assert_eq!(verdict, SendVerdict::Settled);
     }
 
     #[tokio::test]
--- a/src/app/bond/payout.rs
+++ b/src/app/bond/payout.rs
@@ -666,113 +666,126 @@ async fn pay_counterparty(
     // the same value persisted above as `payout_routing_fee_sats`.
     // `send_payment` consumes LND's payment stream until a terminal state,
     // which a locked-in but unresolved HTLC (hold invoice as the winner's
-    // payout invoice, or an HTLC stuck in route) can delay indefinitely —
-    // unbounded, that pins this scheduler task forever and, because the
-    // status channel is drained only after `send_payment` returns, a chatty
-    // stream could also deadlock on the full channel. Bound it with the same
-    // timeout as the buyer payout; LND stops launching route attempts at 60s,
-    // so past 75s only an unresolved HTLC keeps the stream open.
+    // payout invoice, or an HTLC stuck in route) can delay indefinitely, so
+    // it is bounded with the same timeout as the buyer payout; LND stops
+    // launching route attempts at 60s, so past 75s only an unresolved HTLC
+    // keeps the stream open.
+    //
+    // The bounded send and the status drain run CONCURRENTLY (`join!`):
+    // `send_payment` forwards every LND update through `tx` and blocks when
+    // the channel fills, so draining only after it returned could deadlock a
+    // chatty stream (>100 updates) until the timeout. Same watcher-before-send
+    // pattern as the buyer payout in `do_payment`, kept on this task (no
+    // spawn) because the scheduler job needs a single combined outcome. The
+    // drain always terminates: when the send future ends — normal return, RPC
+    // error, or dropped by the timeout — `tx` drops and `rx.recv()` yields
+    // `None`.
     let (tx, mut rx) = channel(100);
-    let send_outcome = timeout(
+
+    let send_fut = timeout(
         crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT,
         ln_client.send_payment(invoice, counterparty_share, tx),
-    )
-    .await;
-    let send_outcome = match send_outcome {
-        Ok(outcome) => outcome,
-        Err(_) => {
-            // Timed out without a terminal state. Dropping the future closes
-            // our side of the gRPC stream but does NOT cancel the payment: a
-            // locked-in HTLC cannot be cancelled by the sender and may still
-            // settle. Indeterminate keeps the invoice + hash so the
-            // reconciliation branch above resolves the real outcome on the
-            // next tick — never re-prompt the winner against a payment that
-            // may still succeed.
-            return on_send_payment_failure(
-                pool,
-                bond,
-                max_retries,
-                claim_window_seconds,
-                PaymentFailureKind::Indeterminate,
-                &format!(
-                    "send_payment reached no terminal state after {}s",
-                    crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT.as_secs()
-                ),
-            )
-            .await;
+    );
+
+    // Collect the first terminal status from the stream. Mirrors
+    // dev_fee::send_dev_fee_payment; each recv stays bounded by
+    // `PAYMENT_STATUS_RECV_TIMEOUT` as a second line of defense, although the
+    // send-side timeout above already bounds the whole exchange. We track
+    // *why* the stream ended: only an explicit `PaymentStatus::Failed` is
+    // terminal. A timeout or clean EOF leaves the payment outcome unknown
+    // (it may still be in flight), so it is routed as `Indeterminate` —
+    // `on_send_payment_failure` then keeps the invoice + hash for
+    // reconciliation instead of re-prompting.
+    let drain_fut = async {
+        let mut succeeded = false;
+        let mut failure: Option<(PaymentFailureKind, String)> = None;
+        loop {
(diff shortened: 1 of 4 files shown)


Catrya:
@coderabbitai review

src/app/release.rs
```diff
@@ -22,9 +22,22 @@ use nostr_sdk::prelude::*;
...
+/// Upper bound on how long the background payout task waits for
+/// `send_payment` to reach a terminal state. LND itself stops launching new
+/// route attempts after `timeout_seconds: 60`, so 75s leaves margin for LND to
+/// exhaust its own attempts; past that, the stream is only kept open by an
+/// HTLC that is locked-in but unresolved (hold invoice or stuck route), which
+/// the sender cannot cancel anyway. Hitting this timeout does NOT fail the
+/// payout: the payment may still settle in LND, so the claim marker is kept
+/// and `reconcile_inflight_payout` resolves the outcome by payment hash.
+/// Fixed for now; could become a settings knob later.
+pub(crate) const PAYOUT_SEND_PAYMENT_TIMEOUT: Duration = Duration::from_secs(75);
```
grunch: **Nit: this 75s is silently coupled to a literal in another module.**

The doc comment's justification is accurate today - I checked `src/lightning/mod.rs:299`, which sets `timeout_seconds: 60` - but the two values are unrelated literals in different files. If someone raises LND's `timeout_seconds` to, say, 120, this bound quietly becomes *tighter* than LND's own routing window and every slow-but-succeeding payout starts taking the timeout branch, with no compile error and no test to catch it.

Suggest deriving one from the other:

```rust
// in src/lightning/mod.rs
pub(crate) const LND_PAYMENT_TIMEOUT_SECS: u64 = 60;
// ... SendPaymentRequest { timeout_seconds: LND_PAYMENT_TIMEOUT_SECS as i32, .. }

// here
pub(crate) const PAYOUT_SEND_PAYMENT_TIMEOUT: Duration =
    Duration::from_secs(crate::lightning::LND_PAYMENT_TIMEOUT_SECS + 15);
```

src/app/release.rs
```diff
@@ -635,102 +662,152 @@ pub async fn do_payment(
...
+
+        match timeout(
+            PAYOUT_SEND_PAYMENT_TIMEOUT,
+            ln_client_payment.send_payment(&payment_request, amount as i64, tx),
+        )
+        .await
+        {
+            // The stream reached a terminal state; the watcher drains the
+            // remaining updates and finishes the bookkeeping.
+            Ok(Ok(())) => {}
```
grunch, line 819: **The comment overclaims, and the branch it describes is reachable without a terminal state.**

```rust
// The stream reached a terminal state; the watcher drains the
// remaining updates and finishes the bookkeeping.
Ok(Ok(())) => {}
```

`send_payment`'s stream loop is:

```rust
while let Ok(Some(payment)) = stream.message().await.map_err(...) { ... }
Ok(())
```

The `while let Ok(..)` swallows the error: a mid-stream gRPC failure (LND restart, connection reset, TLS drop) ends the loop and `send_payment` returns `Ok(())` **without any `PaymentStatus` having reached the watcher**. Same for a clean EOF with no terminal update. So `Ok(Ok(()))` does not mean "terminal state reached" - it means "the stream ended, for one of three reasons, one of which is a terminal state".

The outcome is still safe (the marker stays set and reconciliation owns it), so this isn't a correctness bug - but the branch as written parks the payout for a full reconciler cycle (60s grace + up to 60s tick) on what is often an immediately-resolvable condition, and the comment tells the next reader the opposite of what happens.

Two options, either is fine:

1. Minimal - fix the comment to state that `Ok(())` also covers a stream that died or EOF'd without a terminal update, and that the marker is deliberately left for reconciliation in that case.
2. Better - run the same `lookup_payment_status` reconcile-now logic the `Ok(Err(_))` branch already has. It is the identical situation (claim held, real outcome unknown, LND can answer immediately), so factoring that block into a small `resolve_claim_via_lookup(&mut ln_client_payment, &ctx, &order, &payout_hash, request_id)` helper and calling it from both branches removes the duplication *and* the reconciler-cycle delay.

src/app/bond/payout.rs
```diff
@@ -664,10 +664,44 @@ async fn pay_counterparty(
...
     // the same value persisted above as `payout_routing_fee_sats`.
+    // `send_payment` consumes LND's payment stream until a terminal state,
+    // which a locked-in but unresolved HTLC (hold invoice as the winner's
+    // payout invoice, or an HTLC stuck in route) can delay indefinitely —
+    // unbounded, that pins this scheduler task forever and, because the
+    // status channel is drained only after `send_payment` returns, a chatty
+    // stream could also deadlock on the full channel. Bound it with the same
+    // timeout as the buyer payout; LND stops launching route attempts at 60s,
+    // so past 75s only an unresolved HTLC keeps the stream open.
     let (tx, mut rx) = channel(100);
```
grunch, line 674: **The bond payout keeps the exact deadlock the buyer payout just had fixed.**

The new comment above is candid about it - *"because the status channel is drained only after `send_payment` returns, a chatty stream could also deadlock on the full channel"* - but the fix applied here only *bounds* that deadlock; it doesn't remove it. Commit 2 solved the same problem properly for the buyer payout by starting the watcher first and running it concurrently. This path is left with the old shape.

It is genuinely reachable: `send_payment` builds `SendPaymentRequest { timeout_seconds: 60, fee_limit_sat, ..Default::default() }`, so `no_inflight_updates` defaults to `false` and LND streams an update per HTLC attempt. A payout that probes many routes can exceed the `channel(100)` buffer, at which point `listener.send(msg).await` blocks forever with nothing draining `rx`. The result is now a 75s stall followed by `Indeterminate` - `payout_attempts` incremented and a full tick of delay - for a payment that may already have succeeded.

Since `pay_counterparty` needs the *first terminal status* rather than fire-and-forget bookkeeping, `tokio::join!` is the natural fit here (no second task, no restructuring of the `succeeded` / `failure` state machine):

```rust
let (tx, mut rx) = channel(100);
let send_fut = ln_client.send_payment(invoice, counterparty_share, tx);
let drain_fut = async {
    // existing `loop { match timeout(PAYMENT_STATUS_RECV_TIMEOUT, rx.recv()).await { ... } }`
};
let (send_outcome, (succeeded, failure)) =
    match timeout(PAYOUT_SEND_PAYMENT_TIMEOUT, async { tokio::join!(send_fut, drain_fut) }).await {
        Ok(pair) => pair,
        Err(_) => return on_send_payment_failure(..., PaymentFailureKind::Indeterminate, ...).await,
    };
```

If you'd rather keep this PR tightly scoped to "bound the wait", that's a defensible call - but then please say so explicitly in the commit body and open a follow-up issue, rather than leaving the deadlock described-but-unfixed in a comment.

grunch (changes requested):
## Review

The core of this change is right and the reasoning in the commit messages is unusually good. Moving everything past the idempotency claim off the event loop, and leaning on the persisted claim + `reconcile_inflight_payout` as the durable recovery path, is the correct shape for this bug. I verified on the branch head: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test` 1195 passed / 0 failed. I also confirmed `do_payment` is the last statement in both `release_action` and `admin_settle`, so the new *dispatched* semantics introduce no ordering race with the inline tail of those handlers, and that the bond payout `Indeterminate` path cannot double-pay (`pay_counterparty` step 1 reconciles on `payout_payment_hash` before any re-send).

Four things I'd like addressed before merge, then this is a clear approve.

---

### 1. The pre-send duplicate guard this PR's safety argument cites is querying the wrong hash

`src/lightning/mod.rs:269`:

```rust
```

`Bolt11Invoice::signable_hash()` is *"The hash of the `RawBolt11Invoice` that was signed"* - not the BOLT11 payment hash. Everywhere else in the codebase uses `invoice.payment_hash().as_ref()` (`release.rs:655`, `bond/payout.rs:571`, `dev_fee.rs:953`). So `send_payment`'s pre-send `track_payment_v2` check asks LND about a hash LND has never seen, always gets `NotFound`, and the guard never fires.

This is a pre-existing line, but it is not incidental to this PR - two comments explicitly name it as the backstop for exactly the windows this PR widens:

- `release.rs:~903` - *"`send_payment`'s own pre-send `track_payment_v2` check backstops any race where LND actually still has the payment"*
- `bond/payout.rs:~613` - *"Safe to attempt a fresh send; `send_payment`'s own pre-send check will reject any payment LND does already know about"*

The bond path is where it bites concretely: after the new 75s `Indeterminate` timeout the payment may be genuinely in flight; if a later tick's `lookup_payment_status` returns `Ok(None)` (LND restarted or pruned the record), step 1 falls through to a fresh `send_payment` **for the same invoice**, and the only thing standing between that and a double payout is this guard. Please fix it here (one line) or open a follow-up and drop the two comments that claim protection that doesn't exist.

### 2. The bond payout keeps the deadlock this PR fixes for the buyer payout

See inline comment on `src/app/bond/payout.rs`. The buyer payout got watcher-before-send; the bond payout is left draining `rx` only after `send_payment` returns, with `SendPaymentRequest` not setting `no_inflight_updates` (so LND streams every attempt update into a `channel(100)`). Bounding it at 75s turns an infinite hang into a bounded one, which is an improvement, but it leaves the same class of bug half-fixed in the same PR.

### 3. New fan-out in `job_retry_failed_payments`

`src/scheduler.rs:236`:

```rust
for payment_failed in payment_failed_list.into_iter() {
    if payment_failed.payment_attempts < retries_number {
        if let Err(e) = do_payment(&ctx, payment_failed.clone(), None).await { ... }
    }
}
```

This loop used to be serialized *because* `do_payment` awaited the payment to a terminal state. Now `do_payment` returns as soon as the claim is persisted, so a tick with a backlog of N failed payouts spawns N concurrent tasks, each holding its own `LndConnector::new()` (a fresh gRPC/TLS connection) for up to 75s. Correctness is fine - the claim CAS still prevents duplicates - but the resource profile changed materially and silently. A `Semaphore` around the dispatch, or reusing a single connector for the tick, would keep the fix without the burst.

### 4. No test covers any of the new behaviour

`PAYOUT_SEND_PAYMENT_TIMEOUT` appears in zero tests. None of the three new behaviours - `do_payment` returning `Ok(())` before settlement, the timeout keeping the claim marker, watcher-started-before-send - is exercised. The existing `do_payment` tests only reach pre-claim failures, and I accept that `LndConnector` being a concrete type (no trait seam, unlike `EscrowBackend`) makes the buyer path genuinely hard to test today. But the bond timeout branch *is* testable the same way `send_payment_indeterminate_failure_keeps_invoice` already is - please add at least that one, asserting the `Indeterminate` route keeps `payout_invoice` + `payout_payment_hash`.

---

### Non-blocking notes

- Base branch is `payout-state-tracking`, not `main` - stacked PR, must land after its base.
- `crate::app::release::PAYOUT_SEND_PAYMENT_TIMEOUT` is now consumed from `app::bond::payout`. A payment-layer bound living in `app::release` is odd layering once it has two consumers; consider moving it next to `send_payment` in `lightning/`.

Everything else - the pre-claim `get_buyer_pubkey()` move, keeping the marker on timeout rather than re-arming, the inline `lookup_payment_status` reconciliation on RPC error, spawning the watcher as its own task so it survives the outer future being dropped - reads correct to me.

Catrya:
All four points are addressed, plus both non-blocking notes. The branch was also rebased onto `main` after #881's squash-merge, so the PR now shows only this work. Point by point:

**1. Duplicate guard queried the wrong hash** → fixed in `fix(lightning): key the duplicate-payment guard on the real payment hash`.

You're right, and it's worse than a stale comment: `signable_hash()` is the invoice's *signature* digest, which LND never indexes, so the guard has never fired in the code's lifetime. It now keys on `invoice.payment_hash()` — but note the fix couldn't be just swapping the hash: with the real hash, the old existence-based logic ("LND tracks it → abort") would have aborted legitimate retries of a *failed* payment. The guard now decides by **status**: only `InFlight` or `Succeeded` aborts; `Failed`/`Unknown`/no-record proceeds, and a lookup transport error proceeds too (LND's own duplicate rejection for an in-flight/settled hash remains the hard backstop, and an unreachable LND fails the subsequent send anyway). It reuses `lookup_payment_status`, the same primitive the payout and bond reconciliation paths already use. Side effect: `send_payment`'s log lines now print the real payment hash.

**2. Bond payout kept its drain-after-send deadlock** → fixed in `fix(bond): drain the payment status stream concurrently with send_payment`.

Agreed (CodeRabbit flagged it too). The bounded send and the status drain now run concurrently via `tokio::join!` — same watcher-before-send pattern as `do_payment`, kept on the task (no spawn) because the scheduler job needs a single combined outcome. The drain always terminates: when the send future ends (return, RPC error, or timeout drop), `tx` drops and `recv()` yields `None`. Bonus semantics improvement: a terminal verdict from the stream now takes priority over how the send future ended, so a `Succeeded` delivered just before the cutoff finalizes the slash immediately instead of deferring to reconciliation. The per-recv `PAYMENT_STATUS_RECV_TIMEOUT` stays as a second line of defense.

**3. Retry-loop fan-out** → fixed in `fix: bound concurrent buyer-payout dispatch with a semaphore`.

A static 8-permit semaphore is acquired at the top of the spawned task and held (RAII) through send, watcher drain, and RPC-error reconcile, so a backlog queues instead of fanning out into N concurrent connections/streams. One interplay worth noting: a task queued past the reconcile grace window can lose its claim to re-arm-and-redispatch; that is safe precisely because of the point-1 fix — the now-working pre-send guard (plus LND's duplicate rejection) stops the late sender from double-paying.

**4. Test coverage** → partially already there, rest added in `test(bond): cover the payout send-outcome classification`.

The minimum you asked for already exists: `send_payment_indeterminate_failure_keeps_invoice` asserts the `Indeterminate` branch preserves `payout_invoice` + `payout_payment_hash` through budget exhaustion (it dates from the #750 review), and `pay_counterparty_fresh_send_persists_hash_then_fails_indeterminate` drives the full RPC-failure path against a dead endpoint — both pass unchanged against the new `join!` wiring.

What genuinely lacked coverage was the **timeout branch**, which is unreachable with a dead endpoint (it fails fast, it never hangs). The new commit extracts the post-`join!` verdict logic into a pure `classify_send_verdict` and covers its full matrix (6 tests): stream `Succeeded` wins over a send timeout, explicit `Failed` maps to `Terminal`, and the three no-verdict cases (send timeout, RPC error, stream EOF/recv-timeout) each classify as `Indeterminate` with their specific message — the `Elapsed` value is produced with a zero-duration timeout, no mock needed.

The remaining gap is `do_payment`'s spawned path itself (timeout keeps the claim marker, watcher-before-send): unit-testing it requires a mockable `LndConnector` (trait), which I'd rather do as its own refactor than smuggle into this PR. That behavior is verified end to end on regtest: an unresolved payout leaves the daemon processing messages, logs the timeout warn at exactly 75s, survives a restart without duplicate dispatch, and resolves through reconciliation (finalize on settle, re-arm + re-prompt on failure).

**Non-blocking notes** → both taken, in `refactor(lightning): derive the payout timeout from LND's route timeout`. The constant now lives in the lightning layer next to `send_payment`, and it is *derived*: `LND_PAYMENT_ROUTE_TIMEOUT_SECS (60) + 15s margin`, so raising LND's window can never silently undercut its own route attempts — and the bond module no longer reaches into `app::release` for a lightning-layer constant.

grunch (changes requested):
Re-review after the five follow-up commits. Three of my earlier points are addressed well: the timeout is now derived from `LND_PAYMENT_ROUTE_TIMEOUT_SECS` (47ea73c), the bond payout drains concurrently with the send (9883aad), and `classify_send_verdict` is a genuinely nice refactor with real tests behind it (05a1665). `d74c111` is also a good catch on its own — the old `track_payment_v2(signable_hash())` guard could never match, so this is the first version of that check that can actually fire.

Blocking on one new issue introduced by `bc1e1f9`.

**Blocking**

- The dispatch semaphore re-opens a double-payout window: an unbounded wait now sits *between* the idempotency claim and the send, so reconciliation can re-arm the claim (and prompt the buyer for a fresh invoice) while a task still sits in the permit queue. Neither the new hash-keyed guard nor LND's duplicate rejection covers that case, because the re-armed payout uses a different invoice. Details inline on `src/app/release.rs:687`. The claim marker is the whole safety property this PR leans on for backgrounding, so this needs closing before merge.

**Should fix**

- `src/app/bond/payout.rs:741` — `join!` keeps awaiting the send future after the drain already saw `Succeeded`, with nothing draining the channel any more; and `PAYMENT_STATUS_RECV_TIMEOUT` (120s) is now unreachable behind the 75s send bound.
- `src/lightning/mod.rs:306` — the guard's `lookup_payment_status` is an unbounded await now on every `send_payment`, including `dev_fee::send_dev_fee_payment`'s 5s budget.
- My earlier note on `src/app/release.rs:775` is still open with no reply. `Ok(Ok(()))` does not mean "the stream reached a terminal state" — `send_payment`'s `while let Ok(Some(..))` swallows a mid-stream gRPC error and returns `Ok(())`, so that branch also covers a stream that died or EOF'd with no terminal update, and the comment tells the next reader the opposite. Either fix the comment or run the same `lookup_payment_status` reconcile-now block the `Ok(Err(_))` branch already has.

**Testing**

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test` (1209 passed) are all clean on `47ea73c` here — no complaints there. But the riskiest new code has no test at all: the `do_payment` tests still only cover the pre-claim paths, and nothing exercises the background dispatch, the timeout-keeps-the-claim invariant, or the semaphore queueing. The bond side got `classify_send_verdict` extracted specifically so it could be unit-tested; the buyer side deserves the same treatment — a small `dispatch_payout_task`-shaped seam (or at least a test that the claim survives a timed-out send) would have caught the issue above.

src/app/release.rs
```diff
@@ -637,115 +664,172 @@ pub async fn do_payment(
...
+    // this task (RPC error, timeout, process restart), reconciliation can
+    // always finish or fail the payout by hash, so it is never lost or paid
+    // twice.
+    tokio::spawn(async move {
+        // Bound concurrent sends (see PAYOUT_DISPATCH_SEMAPHORE). The
+        // semaphore is static and never closed, so acquire() only errs if it
+        // were closed; proceeding unpermitted in that impossible case beats
+        // silently dropping a claimed payout. The permit is held for the
+        // whole task (send, watcher drain, RPC-error reconcile) via RAII.
+        let _permit = PAYOUT_DISPATCH_SEMAPHORE.acquire().await;
```
grunch, line 692: **Blocking: a permit-queued task can double-pay the buyer.**

The doc on `PAYOUT_DISPATCH_SEMAPHORE` argues the queue-past-grace case is safe because "the pre-send duplicate guard in `send_payment` (keyed on the real payment hash) and LND's own duplicate rejection stop the late sender from double-paying". Both of those are keyed on *this* invoice's hash — and the re-arm path does not reuse this invoice.

The sequence:

1. Backlog: the retry job fans more than 8 failed payouts into background tasks, so task #9 blocks on this `acquire()`. Each wave holds its permit for up to `PAYOUT_SEND_PAYMENT_TIMEOUT` (75s), so the wait is unbounded in principle and ~75s per wave in practice.
2. The reconciler's grace window is `max(payment_retries_interval, MIN_GRACE_SECS = 30)` (`scheduler.rs:267`) and it ticks every 60s. A task that has queued for more than that is already visible to `find_inflight_payouts` — while it has not called `send_payment` even once.
3. `reconcile_inflight_payout` looks the hash up in LND. Nothing was sent, so LND has no record → `Ok(None)` → the `Failed | Unknown | None` arm at `release.rs:1019` → `fail_order_payout` + `check_failure_retries_or_log`. The claim is cleared and the buyer gets `payment-failed` → `add-invoice`.
4. The buyer submits a **new** invoice. `pay_new_invoice` is no longer blocked (marker gone), a new claim is created under a **different** payment hash, and a second `do_payment` dispatches it.
5. Task #9 finally gets its permit and sends the **old** invoice. The guard in `send_payment` looks up the old hash, LND still has no record of it, so it proceeds. LND has nothing to reject either — two distinct hashes, two distinct payments.

Both invoices settle against one settled escrow. That is precisely what the claim marker exists to prevent; the semaphore re-opens the window by putting an unbounded wait *between* the claim and the send. Before `bc1e1f9` the gap was just the send call itself, which is what the comment at `release.rs:652` still promises.

Cheapest fix — re-validate ownership of the claim after acquiring the permit:

```rust
let _permit = PAYOUT_DISPATCH_SEMAPHORE.acquire().await;
// Reconciliation may have re-armed this claim while we queued; if it did,
// a newer payout owns the order and this invoice must not be sent.
if !crate::db::payout_claim_is_current(ctx.pool(), order.id, &payout_hash, payout_claimed_at)
    .await
    .unwrap_or(false)
{
    warn!(
        "Order {}: payout claim was re-armed while queued; dropping stale dispatch",
        order.id
    );
    return;
}
```

`fail_order_payout` / `clear_order_payout` already scope on `(order_id, hash, claimed_at)`, so that predicate is a one-liner over the existing schema and is worth a unit test of its own.

If you'd rather avoid the extra DB round trip, either acquire the permit *before* `claim_order_payout` (then the claim is never older than the send by more than the send itself), or bound the queue wait with `try_acquire_owned` plus a wait strictly shorter than `MIN_GRACE_SECS` and fail the claim locally if the permit never arrives.

src/app/bond/payout.rs
```diff
@@ -664,82 +664,152 @@ async fn pay_counterparty(
...
                         }
-                        _ => {}
                     }
                 }
             }
         }
+        (succeeded, failure)
+    };
+
+    let (send_outcome, (succeeded, stream_failure)) = tokio::join!(send_fut, drain_fut);
```
grunch, line 719: **`join!` keeps waiting on the send future after the drain already saw `Succeeded`.**

`drain_fut` breaks out of its loop on `PaymentStatus::Succeeded`, but `rx` lives in the enclosing scope and `join!` holds both futures alive until *both* complete — so breaking neither closes the channel nor ends the wait. `send_payment` is still `listener.send(msg).await`-ing every remaining LND update into a channel nobody drains any more, and the job pays up to the full 75s for a payment it already knows settled.

In practice LND closes the stream right after the terminal update and the 100-slot buffer absorbs the tail, so this is a latency/shape issue rather than a live deadlock — but it is the same "nobody is draining" shape this commit set out to remove, so it is worth closing properly. Moving `rx` into the drain and dropping it on the way out makes the send future finish immediately (`listener.send` errors → `Ok(Err(..))`, which `classify_send_verdict` already subordinates to `succeeded`):

```rust
let drain_fut = async move {          // take ownership of `rx`
    let mut succeeded = false;
    let mut failure: Option<(PaymentFailureKind, String)> = None;
    loop { /* unchanged */ }
    // Unblock a send that is still pushing updates into a channel we are
    // done reading.
    drop(rx);
    (succeeded, failure)
};
```

Separately: `PAYMENT_STATUS_RECV_TIMEOUT` (120s) is now unreachable. The send side is bounded at 75s, and once its future is dropped `tx` drops and `rx.recv()` returns `None` on the spot, so the recv timeout can never fire. The comment calls it "a second line of defense"; in this shape it is dead code. Either drop it below 75s so it can actually fire first, or say plainly that it is vestigial — `classify_stream_recv_timeout_is_indeterminate` currently tests a branch that production can no longer reach.

src/lightning/mod.rs
```diff
@@ -266,37 +284,42 @@ impl LndConnector {
...
-                "Track error".to_string(),
-            )));
+        // Duplicate-dispatch guard: refuse to send only when LND reports this
+        // hash as already in flight or settled. A Failed/Unknown/absent record
+        // must NOT abort — the retry flow legitimately re-sends the same
+        // invoice after a failure. A lookup transport error also proceeds:
+        // LND itself rejects a duplicate SendPaymentV2 for an in-flight or
+        // settled hash (the hard backstop behind this check), and if LND is
+        // truly unreachable the send below fails anyway.
+        match self.lookup_payment_status(&payment_hash).await {
```
grunch: **The guard adds an unbounded RPC to every `send_payment`, including the 5s dev-fee path.**

Keying on `payment_hash()` instead of `signable_hash()` is right, and worth stating in the commit body that this is a behavior change rather than a pure bug fix: the old `track_payment_v2(signable_hash())` check returned `NotFound` for every invoice, so `send_payment` effectively had **no** duplicate guard. This is the first version that can actually abort a send.

The cost is that `lookup_payment_status` opens a `track_payment_v2` stream and awaits `stream.message()` with no timeout of its own, and it now runs before every payment. The two payout paths cover it with the new 75s outer bound, but `dev_fee::send_dev_fee_payment` (`src/app/dev_fee.rs:992`) wraps `send_payment` in a **5s** timeout that previously only had to cover `send_payment_v2` — it now has to cover a full track round trip first. A slow LND turns dev-fee payments into `send_payment timeout` errors that used to go through.

Worth bounding the lookup itself here (it is the only unbounded await left inside `send_payment`), so callers get a guard that degrades to "proceed" — which the comment above already says is the safe direction, since LND rejects a genuine duplicate anyway — instead of eating the caller's whole budget.

grunch (approved):
Re-reviewed `5b8b4a9` + `49190c4`. The blocking issue is properly closed and the other three notes are addressed.

**On the double-payout fix (`5b8b4a9`)**

`touch_order_payout_claim` is the right shape, and the choice to *refresh* the token rather than merely check it is a better answer than what I suggested — a plain ownership check would still have left reconciliation free to re-arm in the gap between the check and LND registering the payment, and refreshing closes that too. I traced the CAS end to end:

- `payout_claimed_at IS ?4` (null-safe) + `payout_payment_hash = ?3` + `status = 'settled-hold-invoice'` matches the scoping in `claim_order_payout` / `fail_order_payout` / `clear_order_payout`, so ownership, invoice identity and order state are all pinned in one statement.
- The refreshed token shadows the outer binding *before* the watcher closure and the send are constructed (`release.rs:703`), so `clear_order_payout` in the `Succeeded` path, `fail_order_payout` in both the watcher's `Failed` path and the outer `Ok(Err(..))` path all carry the new token. No path was left on the stale one.
- The `Ok(None)` and `Err(..)` branches both drop the send and keep the marker, which is the recoverable direction: `find_inflight_payouts` still sees the row under the *old* token whose grace has long elapsed, so the reconciler resolves it on the next tick.
- The five new `db.rs` tests cover exactly the sequence I described — `test_touch_payout_claim_loses_to_replacement_claim` is the double-payout scenario verbatim, and `test_touch_payout_claim_invalidates_pre_touch_snapshot` pins the reconciler-snapshot half that I had not asked for.

**On `49190c4`**

- `drop(rx)` at the end of the drain makes the `Succeeded` case return immediately instead of riding out the 75s bound, and `classify_stream_succeeded_wins_over_send_error` pins the resulting `Ok(Err(channel closed))` not shadowing the settled verdict. Removing `PAYMENT_STATUS_RECV_TIMEOUT` outright rather than keeping a bound that could never fire is the honest call: the drain is now bounded transitively and airtight, since `tx` only lives inside `send_payment`'s frame and the 75s `timeout` drops that frame.
- `DUPLICATE_GUARD_LOOKUP_TIMEOUT` bounds the only unbounded await left in `send_payment`, and degrading to "proceed" on timeout is right given LND's own duplicate rejection is the hard backstop.
- The `Ok(Ok(()))` comment now says what the branch actually covers.

**Verification**

On `49190c4`: `cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test` (1214 passed, +5) all clean locally.

**Non-blocking, for whenever**

Two small notes inline (`release.rs:42`, `lightning/mod.rs:333`) — both are doc/observability, neither is worth another round.

One last thought rather than a request: the dispatch task itself still has no test at the `release.rs` level. The db layer is now well covered, and the 20 lines of wiring read correctly, so this is fine to merge — but if the buyer payout ever grows a third failure mode, extracting the task body the way `classify_send_verdict` was extracted on the bond side would pay for itself.

Good work on the turnaround. LGTM.

src/app/release.rs
```diff
@@ -23,8 +23,24 @@ use sqlx::{Pool, Sqlite};
...
+/// connection and payment stream for up to [`PAYOUT_SEND_PAYMENT_TIMEOUT`];
+/// this semaphore makes a backlog queue instead of fanning out. The queue puts
+/// an unbounded wait between the claim and the send, during which
+/// reconciliation may re-arm the claim — and the buyer may then supply a fresh
+/// invoice with a *different* hash, a case neither the pre-send duplicate
+/// guard nor LND's duplicate rejection can catch; that window is closed by
+/// `touch_order_payout_claim` right after the permit: a task whose claim was
+/// re-armed or replaced while it queued drops its send. Fixed for now; could
+/// become a settings knob later.
+static PAYOUT_DISPATCH_SEMAPHORE: Semaphore = Semaphore::const_new(8);
```
grunch, line 42: Non-blocking, doc only: the semaphore does not actually bound LND connections.

`LndConnector::new()` runs at `release.rs:645`, before the claim and therefore before the queue — deliberately, so a connect blip never leaves a marker set. So a backlog of N payouts still opens N gRPC connections; each one then sits idle for the whole queue wait and only the *sends* are capped at 8. What this semaphore bounds is concurrent payment streams, not connections.

The doc says "each holding its own LND gRPC connection and payment stream ... this semaphore makes a backlog queue instead of fanning out", which reads as though both are capped. Worth narrowing to the payment stream, since the connection fan-out is a deliberate trade for the fail-fast-before-claim property and someone tuning the 8 later will want to know which resource it governs.

src/lightning/mod.rs
```diff
@@ -266,37 +293,49 @@ impl LndConnector {
...
+            | Ok(Ok(Some(payment::PaymentStatus::Succeeded))) => {
+                info!(
+                    "Aborting payment for hash {}: already in flight or settled",
+                    hash
+                );
+                return Err(MostroInternalErr(ServiceError::LnPaymentError(
+                    "payment already dispatched for this hash".to_string(),
+                )));
+            }
+            _ => {}
```
grunch, line 333: Non-blocking: the catch-all silently absorbs the two cases worth knowing about.

Bounding the lookup is right, and proceeding on timeout is the correct default given LND's own duplicate rejection. But `_ => {}` now covers `Err(Elapsed)` and `Ok(Err(transport))` with no trace, so an LND whose `track_payment_v2` consistently exceeds 2s leaves the guard permanently disabled and nothing in the logs says so — the abort path logs, the degraded path does not.

A line on those two arms would make it visible:

```rust
Err(_) => info!(
    "Duplicate guard lookup for hash {} timed out after {}s; proceeding (LND rejects real duplicates)",
    hash,
    DUPLICATE_GUARD_LOOKUP_TIMEOUT.as_secs()
),
Ok(Err(e)) => info!("Duplicate guard lookup for hash {} failed ({e}); proceeding", hash),
_ => {}
```

src/app/release.rs
```diff
@@ -637,115 +669,211 @@ pub async fn do_payment(
...
+    // this task (RPC error, timeout, process restart), reconciliation can
+    // always finish or fail the payout by hash, so it is never lost or paid
+    // twice.
+    tokio::spawn(async move {
+        // Bound concurrent sends (see PAYOUT_DISPATCH_SEMAPHORE). The
+        // semaphore is static and never closed, so acquire() only errs if it
+        // were closed; proceeding unpermitted in that impossible case beats
+        // silently dropping a claimed payout. The permit is held for the
+        // whole task (send, watcher drain, RPC-error reconcile) via RAII.
+        let _permit = PAYOUT_DISPATCH_SEMAPHORE.acquire().await;
```
grunch, line 692: **A permit-queued task burns the buyer's retry budget for a payout that was never sent.**

`touch_order_payout_claim` closes the **double-pay** window from the earlier review. It does not close the **spurious-failure** window that the same queue opens, and the doc on `PAYOUT_DISPATCH_SEMAPHORE` (L33-41) only reasons about the former.

Reachable on stock settings — `payment_attempts = 3`, `payment_retries_interval = 60`, so `grace_secs = max(60, 30) = 60` and the reconcile tick is 60s:

1. `job_retry_failed_payments` (`scheduler.rs:233-240`) walks `find_failed_payment` **sequentially**, and `do_payment` now returns right after claiming. N failed payouts therefore become N spawned tasks in one tight loop.
2. Eight acquire permits; the rest block here. Wave 1 holds its permits for up to `PAYOUT_SEND_PAYMENT_TIMEOUT` (75s) — which is precisely the stuck-payout case this PR exists to survive. So wave 2 queues for **more than the 60s grace**.
3. Its claim is now older than `grace_secs`, so `find_inflight_payouts` returns it and `reconcile_inflight_payout` looks the hash up. **Nothing was ever sent**, so LND has no record → `Ok(None)` → the `Failed | Unknown | None` arm → `fail_order_payout` + `check_failure_retries_or_log`.
4. `check_failure_retries` (L79-101) bumps `payment_attempts`, and once `payment_attempts >= retries_number` it enqueues `Action::AddInvoice` — asking the buyer for a **fresh invoice for a payment that was never attempted**. At `payment_attempts = 3`, two such waves exhaust the budget.
5. The task finally gets its permit, `touch` returns `None`, the send is dropped. Correct — but the damage is already committed, and the payout now depends on the buyer re-submitting.

No funds are at risk; that part `touch` does fix. But a transient backlog now converts into user-visible payment failures and retry exhaustion, on orders where nothing was ever attempted.

Keeping the claim fresh while queued is the minimal fix, and the token is already refreshable:

```rust
let _permit = loop {
    tokio::select! {
        p = PAYOUT_DISPATCH_SEMAPHORE.acquire() => break p,
        // HEARTBEAT must stay below MIN_GRACE_SECS
        _ = tokio::time::sleep(HEARTBEAT) => {
            match crate::db::touch_order_payout_claim(
                ctx.pool(), order.id, &payout_hash, Some(payout_claimed_at),
            ).await {
                Ok(Some(t)) => payout_claimed_at = t,
                // lost the claim, or a DB error: drop the dispatch either way
                _ => return,
            }
        }
    }
};
```

The alternative already raised in the previous round — acquire the permit **before** `claim_order_payout` — also works and needs no heartbeat, at the cost of letting `find_failed_payment` re-pick the order while it queues.

src/app/bond/payout.rs
```diff
@@ -664,82 +655,139 @@ async fn pay_counterparty(
...
     }
-
-    // EOF with no terminal status (the `Ok(None)` break above) is also
-    // indeterminate: the stream closed without telling us the outcome.
-    let (kind, msg) = failure.unwrap_or((
-        PaymentFailureKind::Indeterminate,
-        "payment stream ended without terminal status".to_string(),
-    ));
-    on_send_payment_failure(pool, bond, max_retries, claim_window_seconds, kind, &msg).await
+    let stream_failure = match stream_failure {
```
grunch, line 764: **`classify_send_verdict` admits a state its only caller can no longer produce.**

`49190c4` removed `PAYMENT_STATUS_RECV_TIMEOUT`, and with it the only producer of an `Indeterminate` stream failure. `drain_fut` can now yield exactly two things: `None`, or `Some((Terminal, _))`. The signature still takes the wider `Option<(PaymentFailureKind, String)>`, so:

```rust
let stream_failure = match stream_failure {
    Some((PaymentFailureKind::Terminal, msg)) => return SendVerdict::Failure(Terminal, msg),
    other => other,     // provably None from here on
};
// ...
Ok(Ok(())) => stream_failure.unwrap_or((Indeterminate, "payment stream ended…")),
```

Both the rebinding and the `unwrap_or` at L785 are dead: for every input the caller can construct, `stream_failure` is `None` by the time it is read. `Some((Indeterminate, _))` is untestable and untested, and the next reader has to go two functions away and re-derive that from the drain body to know it.

The `succeeded: bool` + `Option<…>` pair has the same problem from the other direction — it can encode `(true, Some(Terminal))`, an impossible pairing the function nonetheless has to pick a winner for. Collapsing both into what the drain actually returns removes the guesswork:

```rust
enum StreamOutcome { Succeeded, Failed(String), Ended }

fn classify_send_verdict(
    send_outcome: Result<Result<(), MostroError>, Elapsed>,
    stream: StreamOutcome,
) -> SendVerdict
```

The six new tests port over almost unchanged, and the matrix they cover becomes total rather than "total over the reachable subset".

src/app/release.rs
```diff
@@ -637,115 +669,211 @@ pub async fn do_payment(
...
+                            _ => {}
                         }
-                        _ => {}
                     }
                 }
             }
+        };
+        tokio::spawn(watcher);
+
+        match timeout(
```
grunch, line 806: **The new dispatch path has no test coverage above the DB layer.**

The four `do_payment` tests (L2126-2260) all assert on **pre-claim** failures — missing invoice, fee-consumed amount, unreachable LND, LNURL resolve-and-validate. None of them reaches `tokio::spawn`, so everything this PR added inside the task is covered only indirectly:

- the `touch` gate → only through the five DB-level tests at `db.rs:3224-3355`. Those are good tests, but they pin the SQL; nothing asserts that the task actually honours `Ok(None)` by dropping the send.
- the semaphore, this 75s `Err(_)` branch, and the watcher-before-send ordering → nothing at all.

Worth noting the bond side of this same PR shows the move: `classify_send_verdict` got extracted into a pure function with six tests *precisely because* the surrounding I/O could not be exercised. The identical extraction is available here — this `match timeout(...)` is three branches over `Result<Result<(), MostroError>, Elapsed>` plus a `keep_marker` decision, all pure once the lookup result is a parameter. That would put the "timeout must keep the marker" invariant, which is the whole safety argument of the PR, under test instead of under a comment.

src/app/release.rs
```diff
@@ -637,115 +669,211 @@ pub async fn do_payment(
...
+    // persisted above is what makes backgrounding safe: whatever happens to
+    // this task (RPC error, timeout, process restart), reconciliation can
+    // always finish or fail the payout by hash, so it is never lost or paid
+    // twice.
+    tokio::spawn(async move {
+        // Bound concurrent sends (see PAYOUT_DISPATCH_SEMAPHORE). The
+        // semaphore is static and never closed, so acquire() only errs if it
+        // were closed; proceeding unpermitted in that impossible case beats
+        // silently dropping a claimed payout. The permit is held for the
+        // whole task (send, watcher drain, RPC-error reconcile) via RAII.
```
grunch, line 691: **The permit doc overstates what RAII covers.** (non-blocking, doc only)

> The permit is held for the whole task (send, watcher drain, RPC-error reconcile) via RAII.

The watcher is `tokio::spawn(watcher)` at L804 — a **sibling** task, not part of the permit-holding one. Its drain is transitively bounded (it ends when `tx` drops along with the send), but the work it does on the final message — `payment_success`, `update_order_event`, the nostr publish, `clear_order_payout` — runs after this permit has already been released.

That is fine as behavior; the semaphore is meant to bound concurrent *payment streams*, and it does. But the comments in this module are load-bearing — they are how the next reader reconstructs the safety argument — so it is worth narrowing to "the send and the RPC-error reconcile" rather than implying the watcher is inside the bound.

src/db.rs
```diff
@@ -1229,6 +1229,49 @@ pub async fn claim_order_payout(
...
+/// Returns `Some(refreshed_at)` — the new per-claim token every later release
+/// must be scoped to — when the caller still owned the claim, or `None` when
+/// the claim was re-armed or replaced while queued (drop the send).
+pub async fn touch_order_payout_claim(
+    pool: &SqlitePool,
+    order_id: Uuid,
+    payment_hash: &str,
+    claimed_at: Option<i64>,
+) -> Result<Option<i64>, MostroError> {
+    let refreshed_at = chrono::Utc::now().timestamp();
```
grunch, line 1259: **The claim token is wall-clock, so it is not monotonic.** (non-blocking; pre-existing at `claim_order_payout`, L1217, but `touch` now leans on it harder)

`touch_order_payout_claim` refreshes with `chrono::Utc::now().timestamp()`. Two consequences — neither exploitable, both worth writing down since the doc at L1232-1252 presents the token as an identity:

- **One-second resolution.** A claim released and re-claimed inside the same second produces an identical token, so a stale task's `touch` can match a *newer* claim. This is not a double-pay: `touch` also matches on `payout_payment_hash`, so a same-token match implies the same invoice, which LND's own duplicate rejection catches. But the CAS is weaker than "the new per-claim token every later release must be scoped to" suggests.
- **Backward clock step.** `find_inflight_payouts` filters on `payout_claimed_at <= now - grace_secs`, so an NTP step backwards makes a just-refreshed claim immediately reconcilable — the same shape as the queue finding above, without needing a backlog to trigger it.

A monotonic counter (or a `rowid`-style sequence) would make the token an identity rather than a timestamp. The grace window can keep using a separate wall-clock column, since it genuinely wants elapsed time.

grunch (changes requested):
**Post-merge audit.** Reviewed the merged state (`2f2b813`) rather than the branch, since #883 landed before this pass. The 8 inline notes from the earlier thread were all addressed by `47ea73c` / `5b8b4a9` / `49190c4` — **nothing below repeats them.**

The core design holds up: claim-before-dispatch is the right idempotency anchor, backgrounding is safe *because* of that claim, and extracting `classify_send_verdict` was the correct move for a branch that could not otherwise be exercised. The findings are almost entirely about what the dispatch semaphore (`bc1e1f9`) added on top of it.

### Findings

| # | Severity | Where | What |
|---|---|---|---|
| 1 | 🔴 High | `release.rs:692` | A permit-queued task burns the buyer's retry budget for a payout that was never sent |
| 2 | 🟡 Medium | `payout.rs:764` | `classify_send_verdict` admits a state its only caller can no longer produce |
| 3 | 🟡 Medium | `release.rs:806` | The new dispatch path has no test coverage above the DB layer |
| 4 | 🔵 Low | `release.rs:691` | The permit doc overstates what RAII covers |
| 5 | 🔵 Low | `db.rs:1259` | The claim token is wall-clock, so it is not monotonic |

<details>
<summary>✅ Verified correct — spot-checked, no action needed</summary>

- `lookup_payment_status` uses `no_inflight_updates: false` (`lightning/mod.rs:413`), so the 2s-bounded duplicate guard **can** actually observe `InFlight`. It does not silently degrade into an always-proceed check, which was my first suspicion when reading the 2s bound.
- `tokio::join!` uses `maybe_done`, which drops a future once it completes — so the 75s `Elapsed` genuinely drops `send_payment`, closes `tx`, and terminates `drain_fut`. The bond drain's `drop(rx)` is symmetric: it fails `listener.send(..)?` at `lightning/mod.rs:383-387` into `Ok(Err(LnNodeError))`, which `classify_send_verdict` correctly subordinates to `succeeded`. `classify_stream_succeeded_wins_over_send_error` pins exactly that.
- Re-keying the duplicate guard onto `payment_hash()` is not a regression for `dev_fee::send_dev_fee_payment`: LND already rejected a duplicate `SendPaymentV2` for an in-flight/settled hash, so the failure mode is unchanged — only reached sooner and with a clearer error.
- `Duration` is still used at `payout.rs:813` after `PAYMENT_STATUS_RECV_TIMEOUT` was removed — no dead import left behind.
- `touch_order_payout_claim`'s `payout_claimed_at IS ?4` is the correct SQLite form for the `Option<i64>` bind, and its `status = 'settled-hold-invoice'` scoping matches `claim_order_payout` and `find_inflight_payouts`.

</details>

**Verdict: REQUEST_CHANGES**, as follow-up work on the merged state. Finding 1 is a real behavioral regression: `touch_order_payout_claim` stopped the queue from *double-paying*, but not from *failing payouts that were never dispatched*. Findings 2-5 are quality, not correctness.

PR #884, opened 2026-08-17
feat(rpc): require bearer-token auth on admin gRPC service

### LINKED ISSUE #807: [HIGH] Admin gRPC service has no code-level authentication

ToRyVand:
Closes #807 (HIGH). The admin gRPC service (`cancel_order`, `settle_order`, `add_solver`, `take_dispute`, `get_version`, `validate_db_password`) had zero code-level authentication — anyone reaching `listen_address:port` had full admin control of the daemon. `validate_db_password`'s rate limiter is anti-brute-force, not authentication (the docs already say it "does not validate any password; it always succeeds," kept only for backward compat). A comment in `call_admin_cancel` falsely claimed "gRPC transport authenticates the operator."

I claimed this issue 8 days ago and posted a design proposal (shared-token over mTLS, since the stated threat model is an operator widening the bind address or fronting it with a reverse proxy that already terminates TLS). Two design questions haven't gotten a reply yet, so I'm proceeding on explicit decisions below rather than blocking further — easy to revisit if you'd rather go a different way.

## What this adds

A shared bearer token (`[rpc].auth_token` / `MOSTRO_RPC_AUTH_TOKEN`), mirroring the existing `nsec_privkey` secret-handling pattern (`secrecy::SecretString`, env-var override, never round-tripped into serialized config). Checked by a new `TokenAuthInterceptor` (`src/rpc/auth.rs`) via constant-time comparison (`subtle::ConstantTimeEq`), wired onto the server with tonic-build's generated `with_interceptor` so every call is rejected with `UNAUTHENTICATED` before it reaches any handler — confirmed with an end-to-end test that binds a real server and connects a real client without the header.

The token travels as gRPC metadata (`authorization: Bearer <token>`), not a proto field, so no `.proto` changes were needed.

## The two decisions I made without a reply

**1. Fail-closed at startup, not a deprecation window.** `[rpc].enabled` defaults to `false`, so untouched deployments are unaffected. Any deployment with `enabled = true` today is running an authless admin surface — that's the exact bug being fixed, so silently continuing on missing config would defeat the fix. `validate_mostro_settings` now rejects `enabled = true` with an empty token, and the daemon refuses to start. Flagged as a `BREAKING CHANGE:` footer on the commit for that specific subset of deployments.

**2. Kept `validate_db_password`'s existing `RateLimiter` gate as-is**, rather than removing it as redundant. Once the interceptor covers the whole service, unauthenticated callers never reach it at all — but it's cheap, already-reviewed code that adds a second layer against a leaked/brute-forced token specifically. No changes to `rate_limiter.rs` or that handler; its existing test (`validate_db_password_succeeds_with_remote_addr`) is untouched since it calls the handler directly, bypassing transport/interceptor like every other handler-level test in that file.

## Also fixed

- The stale `service.rs` comment claiming transport already authenticates.
- Docs: `docs/RPC.md` (Security Considerations + client example), `README.md` (grpcurl examples now show the header), `docs/ADMIN_RPC_AND_DISPUTES.md` (Audit and Safety), `docs/RPC_RATE_LIMITING.md` (the "Strong auth: out of scope" row now points at this), `settings.tpl.toml`.

## Verification

`cargo build`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, `cargo test --bin mostrod` (1199 passed), `markdownlint-cli2` on the touched docs — all clean.

## Test plan

- [x] Interceptor unit tests: accepts valid token, rejects missing/malformed header, rejects wrong token, rejects wrong-length token
- [x] End-to-end test: real server + real client, call without header → `UNAUTHENTICATED`
- [x] Fail-closed tests: `enabled=true` + no token → rejected; `enabled=true` + token → accepted; `enabled=false` + no token (default) → accepted
- [x] Env-var override tests mirroring the existing `nsec_privkey` suite
- [ ] Manual smoke test with `grpcurl` against a running daemon (will do before merge if useful, or happy to have a maintainer confirm independently)

## Summary by CodeRabbit

* **New Features**
  * Added Bearer-token authentication for all administrative RPC calls.
  * Added configuration through TOML settings or the `MOSTRO_RPC_AUTH_TOKEN` environment variable.
  * RPC startup now fails when enabled without a valid authentication token.
  * Added secure token validation and protection against malformed or invalid credentials.

* **Documentation**
  * Updated RPC, administration, rate-limiting, and configuration guidance with authentication requirements and client examples.
  * Clarified compatibility behavior for `ValidateDbPassword` and credential-handling recommendations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2372,6 +2372,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sqlx",
+ "subtle",
  "tokio",
  "toml",
  "tonic 0.14.6",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,7 @@ cdk = { version = "0.17.2", default-features = false, features = ["wallet"] }
 secp256k1 = { version = "0.30", features = ["serde"] }
 secrecy = { version = "0.10", features = ["serde"] }
 zeroize = "1.8"
+subtle = "2.6"
 
 [dev-dependencies]
 tokio = { version = "1.47.1", features = ["full", "test-util", "macros"] }
--- a/README.md
+++ b/README.md
@@ -794,17 +794,19 @@ For client documentation, see the respective client repositories.
 
 ### Admin Operations (RPC Interface)
 
-If RPC is enabled, use admin tools for dispute resolution:
+If RPC is enabled, use admin tools for dispute resolution. Every call requires the `authorization: Bearer <token>` header set to your `[rpc].auth_token` / `MOSTRO_RPC_AUTH_TOKEN` value:
 
 ```bash
+export MOSTRO_RPC_AUTH_TOKEN="your-configured-token"
+
 # Cancel an order (admin override)
-grpcurl -plaintext -d '{"order_id": "abc123"}' localhost:50051 mostro.admin.v1.AdminService/CancelOrder
+grpcurl -plaintext -H "authorization: Bearer $MOSTRO_RPC_AUTH_TOKEN" -d '{"order_id": "abc123"}' localhost:50051 mostro.admin.v1.AdminService/CancelOrder
 
 # Settle disputed order
-grpcurl -plaintext -d '{"order_id": "abc123"}' localhost:50051 mostro.admin.v1.AdminService/SettleOrder
+grpcurl -plaintext -H "authorization: Bearer $MOSTRO_RPC_AUTH_TOKEN" -d '{"order_id": "abc123"}' localhost:50051 mostro.admin.v1.AdminService/SettleOrder
 
 # Add dispute solver
-grpcurl -plaintext -d '{"solver_pubkey": "npub1..."}' localhost:50051 mostro.admin.v1.AdminService/AddSolver
+grpcurl -plaintext -H "authorization: Bearer $MOSTRO_RPC_AUTH_TOKEN" -d '{"solver_pubkey": "npub1..."}' localhost:50051 mostro.admin.v1.AdminService/AddSolver
 ```
 
 ---
--- a/docs/ADMIN_RPC_AND_DISPUTES.md
+++ b/docs/ADMIN_RPC_AND_DISPUTES.md
@@ -41,7 +41,7 @@ sequenceDiagram
 ```
 
 ## Audit and Safety
-- Require admin authentication/authorization at message level.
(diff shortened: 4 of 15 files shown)


grunch:
Hi @ToRyVand this was already addressed here https://github.com/MostroP2P/mostro/pull/877

I'm sorry that @AndreaDiazCorreia and you worked on the same issue, I'm looking on the issue I see you asked for assignment but nobody assinged it to you, let's see which of both PRs are more solid to merged it.

We also need code reviewers, we are not enough people doing reviews of PRs right now, this is a bottle neck for Mostro, if you could help us with that it would be highly appreciate it

ToRyVand:
Thanks for the pointer @grunch — closing this in favor of #877.

I checked: it's already `APPROVED`, and it goes further than mine in a way that matters — it adds optional TLS and an `allow_remote` startup guard that forces an explicit opt-in before binding to a non-loopback address. That's exactly the gap CodeRabbit flagged on this PR as the highest-risk item (a remote listener could otherwise transmit the bearer token in clear, open to capture/replay). No reason to carry two competing fixes for the same issue forward, especially with #877 already reviewed.

Happy to help with review load going forward — will look at open PRs when I have bandwidth, per your note about the bottleneck.

PR #887, opened 2026-08-17
fix: write settings.toml with owner-only permissions

arkanoider:
## Summary

`settings.toml` contains the `nsec_privkey`. Previously, the daemon created it with default permissions (`0o644`) on the headless/Docker/CI path and the wizard manual-setup path, leaving the private key readable by any user on the host.

This PR introduces a shared `util::write_private_file` helper that creates files with `0o600` on Unix and tightens permissions on pre-existing files, then routes all three `settings.toml` creation paths through it.

## Changes

- Add `util::write_private_file` in `src/config/util.rs` (Unix: `OpenOptions::mode(0o600)` + explicit `set_permissions(0o600)`; non-Unix: plain write).
- Replace `std::fs::write` with `write_private_file` in the non-interactive/headless template copy.
- Replace the manual-setup and guided-wizard writes in `src/config/wizard.rs` with `write_private_file`.
- Add unit tests for content writes, new-file permissions, and tightening an existing `0o644` file.

## Summary by CodeRabbit

* **Security Improvements**
  * Configuration files created or updated by the application now use owner-only permissions.
  * Existing configuration files have their permissions tightened automatically before content is written.

* **Bug Fixes**
  * Improved handling and reporting of configuration file write errors.
  * Added coverage for file contents and permission behavior.

### DIFF
--- a/src/config/util.rs
+++ b/src/config/util.rs
@@ -9,8 +9,8 @@ use crate::config::{init_mostro_settings, Settings};
 use mostro_core::error::MostroError::{self, *};
 use mostro_core::error::ServiceError;
 use std::fs;
-use std::io::IsTerminal;
-use std::path::PathBuf;
+use std::io::{IsTerminal, Write};
+use std::path::{Path, PathBuf};
 use zeroize::Zeroizing;
 
 const DB_FILENAME: &str = "mostro.db";
@@ -129,6 +129,47 @@ fn validate_cashu_settings(
     Ok(())
 }
 
+/// Write `contents` to `path` with owner-only permissions (0o600).
+///
+/// On Unix the file is created with mode 0o600 and permissions are tightened
+/// after opening so that pre-existing files are also restricted. This matches
+/// how the `.env` file and the SQLite database are protected.
+pub fn write_private_file(path: &Path, contents: &[u8]) -> Result<(), MostroError> {
+    #[cfg(unix)]
+    let mut file = {
+        use std::os::unix::fs::OpenOptionsExt;
+        std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .mode(0o600)
+            .open(path)
+            .map_err(|e| MostroInternalErr(ServiceError::IOError(e.to_string())))?
+    };
+
+    #[cfg(not(unix))]
+    let mut file = {
+        std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(path)
+            .map_err(|e| MostroInternalErr(ServiceError::IOError(e.to_string())))?
+    };
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        let permissions = std::fs::Permissions::from_mode(0o600);
+        file.set_permissions(permissions)
+            .map_err(|e| MostroInternalErr(ServiceError::IOError(e.to_string())))?;
+    }
+
+    file.write_all(contents)
+        .map_err(|e| MostroInternalErr(ServiceError::IOError(e.to_string())))?;
+    Ok(())
+}
+
 /// Initialize the default settings directory and create a settings file from the template if it doesn't exist.
 /// Checks if the directory already exists, and if not, creates it and writes the template file.
 /// If a custom config path is provided, it uses that instead of the default `~/.mostro` directory.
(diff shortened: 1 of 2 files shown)


AndreaDiazCorreia (commented):
@arkanoider  apologies, I opened #890 a couple of hours after this one without noticing it was already covering the same ground on `settings.toml`. That's on me for not checking first.

The two overlap directly: both replace the same `std::fs::write` in `config::util` and the same inline `OpenOptions` block in `config::wizard`, so whichever lands first will conflict with the other. They also differ on one design point. The helper here tightens an existing file, while mine refuses to touch a path that already exists (`O_EXCL`, so a planted symlink can't be written through).

Would you mind taking a look at #890 so we can decide together which one to keep? Happy either way, including reworking mine on top of yours if you prefer this shape.

arkanoider:
> @arkanoider apologies, I opened #890 a couple of hours after this one without noticing it was already covering the same ground on `settings.toml`. That's on me for not checking first.
> 
> The two overlap directly: both replace the same `std::fs::write` in `config::util` and the same inline `OpenOptions` block in `config::wizard`, so whichever lands first will conflict with the other. They also differ on one design point. The helper here tightens an existing file, while mine refuses to touch a path that already exists (`O_EXCL`, so a planted symlink can't be written through).
> 
> Would you mind taking a look at #890 so we can decide together which one to keep? Happy either way, including reworking mine on top of yours if you prefer this shape.

@AndreaDiazCorreia don't worry it happens in these days...if your pr covers also this one, go and close this in favor of yours!

AndreaDiazCorreia:
I'll leave this open until #890 is actually merged, just so we don't end up with nothing fixing the `settings.toml` permissions if that one takes a while, and close it right after.

Thanks for catching this first.

ToRyVand (commented):
Audited this rather than read it, since the file in question holds `nsec_privkey`.

**Two things are right and worth confirming so nobody "fixes" them backwards:** the order is `open(truncate)` → `set_permissions` → `write_all`, so the mode is tightened *before* any secret byte is written — there is no window where the key exists in a 0o644 file. And `set_permissions` is called on the `File` handle (`fchmod`), not on the path, so there is no TOCTOU between open and chmod. Both are easy to get wrong and both are correct here.

Three findings.

## 1. The fix does not reach the deployments that have the problem

All three call sites sit inside the same guard in `init_configuration_file`:

```rust
if !config_file_path.exists() {
    //  ... interactive wizard / manual template / headless template
}
```

So `settings.toml` is only ever written **when it does not exist**. Which means an operator who already ran an earlier Mostro has a 0o644 `settings.toml` on disk **today**, containing their nsec — and after this PR it stays 0o644 forever, because nothing will ever rewrite it. The population described in the summary ("the daemon created it with default permissions (0o644) … leaving the private key readable by any user on the host") is exactly the population this PR does not help.

This also makes the `set_permissions` call dead for these callers: a freshly `create`d file already has 0o600 from `OpenOptionsExt::mode`. The rustdoc says *"permissions are tightened after opening so that pre-existing files are also restricted"*, and the test `preexisting_world_readable_file_is_tightened` pins that behaviour — but no caller can reach it. The capability is real and the test is honest; it is the wiring that is missing.

I checked whether anything else closes the gap at startup: nothing does. The only `metadata().permissions()` in `src/config/` is inside the `wizard.rs` test module.

The one-liner that would actually remediate it: on load, if `settings.toml` exists and `mode & 0o077 != 0`, tighten it to 0o600 and log that it did — or at minimum warn. That turns this from "new installs are safe" into "the vulnerability is closed".

## 2. `write_env_file` is the same function, one file away, untouched

`src/config/wizard.rs` already contains `write_env_file`, and it is the same implementation as the new `util::write_private_file`: the same `#[cfg(unix)]` `OpenOptions` with `.mode(0o600)`, the same `#[cfg(not(unix))]` fallback, the same post-open `set_permissions(0o600)`, the same error mapping. The only difference is `writeln!(file, "{}={}", NSEC_ENV_VAR, nsec)` where the new one takes `&[u8]`.

Its doc comment even states the rationale this PR's helper restates:

> `OpenOptionsExt::mode(0o600)` only applies when the file is created, so for preexisting files we must explicitly tighten permissions after opening.

The summary says this PR "introduces a **shared** `util::write_private_file` helper" — it shares it across three paths and leaves the fourth copy in the module it edits. And `.env` holds the same nsec, so these are not incidentally similar functions, they are the same requirement.

It collapses to:

```rust
fn write_env_file(path: &Path, nsec: &str) -> Result<(), MostroError> {
    util::write_private_file(path, format!("{}={}\n", NSEC_ENV_VAR, nsec).as_bytes())
}
```

which deletes roughly 30 lines and leaves one place to fix if the permission policy ever changes. The rustdoc line *"This matches how the `.env` file … [is] protected"* would then be true by construction rather than by duplication.

## 3. Windows is a shipped target, and there the function is a plain write

`archs` lists `x86_64-pc-windows-gnu`, so this is a platform you publish binaries for — not a hypothetical. On it, `write_private_file` falls through to `OpenOptions` with no mode and no ACL work, and says nothing about it: the rustdoc describes only the Unix behaviour, and there is no log line at runtime. A Windows operator ends up with the same exposure the PR exists to remove, from a function named `write_private_file`, with no signal that anything is different.

Not necessarily code — but the rustdoc should state the non-Unix behaviour explicitly, and a one-line `warn!` when the config is created on a platform where permissions are not applied would mean the operator finds out from the daemon rather than from an incident.

---

None of this touches the Unix mechanics, which are correct. 1 is the one I would want before merge, since without it the reported vulnerability stays open on every existing install.

PR #888, opened 2026-08-17
fix: stop re-publishing the pending orderbook hourly and at startup

Catrya:
The orderbook reconciler introduced in #872 re-asserted every live `pending` /
`waiting-taker-bond` order on relays once per hour, with the first sweep right
at daemon startup. Per the author of #872, this full re-publication was never
the intended behavior: an order's kind-38383 event should be published when the
order actually transitions, and then left alone.

Beyond being unintended, the sweep was actively harmful:

- **Every re-assert stamped a fresh `created_at`.** The monotonic timestamp
  registry is in-memory, so after a restart nothing bounds the new stamp. No
  pending order ever appeared older than ~1 hour on clients, and sorting the
  book by date was meaningless. (Verified locally: an order's public
  `created_at` jumped forward on every daemon restart.)
- **Every pending order was re-sent to every relay 24 times a day** — pointless
  traffic that relays may well treat as spam.
- **`orders.event_id` drifted**: the sweep deliberately did not persist the
  re-published event id, so the DB column stopped matching the event actually
  live on relays.

The sweep also never healed the "ghost order" class the reconciler exists for:
its query only listed DB-pending rows, so an order that is terminal in the DB
while a relay still advertises it as `pending` was never touched by it.

## What this PR does

Removes the hourly + startup full sweep (`ORDERBOOK_FULL_SWEEP_EVERY_TICKS`,
the `full_sweep` branch of `reconcile_orderbook_once`, and the now-dead
`find_pending_orders_for_reconcile` query). Everything else from #872 stays
intact: the failed-publish queue (drained every 60 s), the NIP-40 expiration
tag capping any stale pending event at the order's real take window, the
monotonic `created_at` registry, and the take-window gate on takes.

A second commit makes the queue the safety net this PR relies on: in
nostr-sdk 0.45 `send_event` returns `Err` only when there was no relay to
send to at all — a per-relay rejection or timeout resolves to `Ok` with the
refusing relays in `output.failed`, which the publish paths never read (and
the `Ok` arm even cleared the order's existing queue entry). All three queue
feeders (`update_order_event_stamped`, the initial pending publish, and the
range child-order publish) now treat a non-empty `failed` map as a failed
publish and queue the order for republish.

## Accepted trade-off

The queue is process-local and only records failures the daemon observed: an
entry queued right before a restart is lost, and a relay that silently drops
an event (or restores from a backup) is never healed. NIP-40 expiration
bounds any such divergence to the order's real take window. This is stated in
the reconciler's doc comment as well.

An idempotent re-assert (re-publishing with the last published stamp so
relays dedupe by event id) could keep the healing without the churn, but it
requires persisting per-order publish state, and the stated intent is that an
order's event is published when the order transitions and then left alone —
so it is left as a possible follow-up rather than done here.

## Tests

- The three reconciler tests (queue drain, requeue while relays unreachable,
  quiescence guard) are kept; only the removed sweep's listing test goes away
  with its query.
- `cargo test --bin mostrod`: 1198 passed, 0 failed. `clippy` and `fmt` clean.
- Verified live: after this change a daemon restart no longer re-publishes the
  pending book, and order `created_at` values stay stable.

## Bug Fixes

* **Orderbook Publishing**
  * Improved recovery when order updates fail to publish or are rejected by individual relays.
  * Failed publications are now automatically queued for retry without unnecessarily republishing healthy pending orders.
  * Retry progress is preserved across temporary interruptions, with retries capped to prevent repeated unsuccessful attempts.
  * Child-order updates that encounter delivery failures are also scheduled for republishing.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -675,34 +675,6 @@ pub async fn find_order_by_date(pool: &SqlitePool) -> Result<Vec<Order>, MostroE
     Ok(order)
 }
 
-/// All orders currently advertised (or advertisable) as `pending` on the
-/// book. Includes `waiting-taker-bond`, which publishes as `pending` on the
-/// wire (Phase 1.5), and excludes `waiting-maker-bond`, which was never
-/// published. Used by the orderbook reconciler to re-assert the book on
-/// relays; only orders whose take window is still open are returned — an
-/// expired row is the expiry job's business, not the reconciler's. Legacy
-/// rows with `expires_at = 0` have no take-window TTL (mirroring
-/// `is_order_take_window_closed`) and stay re-assertable.
-pub async fn find_pending_orders_for_reconcile(
-    pool: &SqlitePool,
-) -> Result<Vec<Order>, MostroError> {
-    let now = Timestamp::now();
-    let orders = sqlx::query_as::<_, Order>(
-        r#"
-          SELECT *
-          FROM orders
-          WHERE (expires_at >= ?1 OR expires_at = 0)
-            AND status IN ('pending', 'waiting-taker-bond')
-        "#,
-    )
-    .bind(now.as_secs() as i64)
-    .fetch_all(pool)
-    .await
-    .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
-
-    Ok(orders)
-}
-
 pub async fn find_order_by_seconds(pool: &SqlitePool) -> Result<Vec<Order>, MostroError> {
     let mostro_settings = Settings::get_mostro();
     let exp_seconds = mostro_settings.expiration_seconds as u64;
@@ -2059,61 +2031,6 @@ mod tests {
         .execute(pool)
         .await
         .unwrap();
-    }
-
-    /// Insert a minimal order row with an explicit status and expiry, for
-    /// the reconciler full-sweep query.
-    async fn insert_book_order(pool: &SqlitePool, status: &str, expires_at: i64) {
-        sqlx::query(
-            r#"
-            INSERT INTO orders (id, kind, event_id, status, premium, payment_method,
-                                amount, fiat_code, fiat_amount, created_at, expires_at)
-            VALUES (?1, 'sell', 'event123', ?2, 0, 'bank',
-                    100000, 'USD', 100, 1700000000, ?3)
-            "#,
-        )
-        .bind(uuid::Uuid::new_v4())
-        .bind(status)
-        .bind(expires_at)
-        .execute(pool)
-        .await
-        .unwrap();
-    }
-
-    /// The full sweep re-asserts only live pending orders: expired rows are
-    /// the expiry job's business and post-trade rows are not book entries.
-    /// Legacy rows with `expires_at = 0` have no take-window TTL and must
-    /// stay re-assertable (mirroring `is_order_take_window_closed`).
-    #[tokio::test]
-    async fn find_pending_orders_for_reconcile_lists_only_live_pending_orders() {
-        let pool = setup_orders_db().await.unwrap();
-        let now = nostr_sdk::prelude::Timestamp::now().as_secs() as i64;
-
-        for (status, expires_at) in [
-            ("pending", now + 3_600),          // live → listed
-            ("waiting-taker-bond", now + 600), // publishes as pending → listed
-            ("pending", 0),                    // legacy, no TTL → listed
-            ("pending", now - 60),             // expired → expiry job's business
-            ("active", now + 3_600),           // post-trade → not a book entry
-            ("waiting-maker-bond", now + 600), // never published → skipped
-        ] {
-            insert_book_order(&pool, status, expires_at).await;
-        }
-
-        let listed = super::find_pending_orders_for_reconcile(&pool)
-            .await
-            .unwrap();
-        let statuses: Vec<String> = listed.iter().map(|o| o.status.clone()).collect();
-
-        assert_eq!(
-            listed.len(),
-            3,
-            "live pending-published rows plus legacy no-TTL: {statuses:?}"
-        );
-        assert!(statuses.contains(&"waiting-taker-bond".to_string()));
-        assert!(listed
-            .iter()
-            .any(|o| o.status == "pending" && o.expires_at == 0));
     }
 
     async fn setup_disputes_table(pool: &SqlitePool) {
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -1085,14 +1085,36 @@ const ORDERBOOK_QUIESCENT_SECS: u64 = 120;
 /// entries are kept for the next pass), so a republish can never supersede
 /// a transition that published before persisting its CAS. On publish
 /// failure the order re-queues itself, so a relay outage self-heals on a
-/// later pass. The returned order (fresh `event_id`) is deliberately not
-/// persisted, matching the CAS-miss repair path: the DB row's status is
-/// the source of truth being re-advertised, not mutated.
+/// later pass — up to [`util::MAX_ORDERBOOK_REPUBLISH_ATTEMPTS`] failed
+/// attempts, after which the entry is dropped: a persistently-failing
+/// relay cannot be converged by republishing (each retry only rewrites
+/// the healthy relays' copies with a fresh `created_at`), so the cap
+/// trades the retry loop for a residual divergence bounded by NIP-40.
+/// The returned order (fresh `event_id`) is deliberately not persisted,
+/// matching the CAS-miss repair path: the DB row's status is the source
+/// of truth being re-advertised, not mutated.
 async fn reconcile_orderbook_once(pool: &sqlx::SqlitePool, keys: &Keys) {
-    for (order_id, generation) in util::take_failed_orderbook_publishes() {
+    for (order_id, generation, attempts) in util::take_failed_orderbook_publishes() {
+        if attempts >= util::MAX_ORDERBOOK_REPUBLISH_ATTEMPTS {
+            warn!(
+                "orderbook reconciler: giving up on order {order_id} after {attempts} failed \
+                 publish attempts — check the relay list; the residual divergence is bounded \
+                 by the event's NIP-40 expiration"
+            );
+            continue;
+        }
         match Order::by_id(pool, order_id).await {
             Ok(Some(order)) => match order.get_order_status() {
                 Ok(status) => {
+                    // Re-seed the drained entry *before* attempting the
+                    // republish: the send path records its own failures via
+                    // `mark_orderbook_publish_failed_at`, which must
+                    // increment on top of the consumed attempts — against
+                    // an absent entry it would restart the count at 1 and
+                    // the attempt cap would never trip. A full success
+                    // clears the entry (its stamp generation is newer), a
+                    // quiescent skip leaves it untouched for the next pass.
+                    util::requeue_orderbook_publish_failure(order_id, generation, attempts);
                     match util::update_order_event_if_quiescent(
                         keys,
                         status,
@@ -1103,8 +1125,12 @@ async fn reconcile_orderbook_once(pool: &sqlx::SqlitePool, keys: &Keys) {
                     {
                         Ok(Some(_)) => {}
                         // Stamped too recently — another publication may be
-                        // in flight; keep the failure queued for later.
-                        Ok(None) => util::mark_orderbook_publish_failed_at(order_id, generation),
+                        // in flight; the re-seeded entry keeps the failure
+                        // queued without consuming an attempt.
+                        Ok(None) => {}
+                        // Pre-send failure (tags/ratings/event build):
+                        // consume an attempt so a permanently broken order
+                        // cannot occupy the queue forever.
                         Err(e) => {
                             warn!(
                                 "orderbook reconciler: republish of order {order_id} failed: {e}"
@@ -1119,8 +1145,9 @@ async fn reconcile_orderbook_once(pool: &sqlx::SqlitePool, keys: &Keys) {
             Ok(None) => {}
             Err(e) => {
                 warn!("orderbook reconciler: could not reload order {order_id}: {e}");
-                // Transient DB error: keep it queued for the next pass.
-                util::mark_orderbook_publish_failed_at(order_id, generation);
+                // Transient DB error: keep it queued for the next pass
+                // without consuming an attempt — nothing was published.
+                util::requeue_orderbook_publish_failure(order_id, generation, attempts);
             }
         }
     }
@@ -1728,6 +1755,96 @@ mod tests {
             probe.created_at.as_secs(),
             1_700_000_000,
             "a reconciler pass must not stamp an order that has no failed publish"
+        );
+    }
+
+    /// The drained entry is re-seeded before the republish attempt, so a
+    /// failed send increments the consumed-attempt count instead of
+    /// restarting it at 1 — and a quiescence skip consumes nothing.
+    #[tokio::test]
+    async fn reconciler_failed_retry_consumes_exactly_one_attempt() {
+        let ctx = migrated_ctx().await;
+        let keys = ctx.keys().clone();
+        let _guard = crate::util::ORDERBOOK_QUEUE_TEST_LOCK.lock().await;
+
+        let order = Order {
+            id: Uuid::new_v4(),
+            kind: Kind::Sell.to_string(),
+            status: Status::Canceled.to_string(),
+            fiat_code: "USD".to_string(),
+            payment_method: "bank".to_string(),
+            expires_at: nostr_sdk::prelude::Timestamp::now().as_secs() as i64 + 3_600,
+            ..Default::default()
+        };
+        let order = order.create(ctx.pool()).await.unwrap();
+        crate::util::mark_orderbook_publish_failed(order.id);
+        assert_eq!(crate::util::orderbook_publish_attempts(order.id), Some(1));
+
+        // No reachable relays in tests: the republish is attempted and
+        // fails, which must cost exactly one more attempt.
+        reconcile_orderbook_once(ctx.pool(), &keys).await;
+        assert_eq!(
+            crate::util::orderbook_publish_attempts(order.id),
+            Some(2),
+            "a failed republish must increment the consumed attempts, not reset them"
+        );
+
+        // The republish just stamped the order, so the next pass hits the
+        // quiescence guard: requeued without consuming an attempt.
+        reconcile_orderbook_once(ctx.pool(), &keys).await;
+        assert_eq!(
+            crate::util::orderbook_publish_attempts(order.id),
+            Some(2),
+            "a quiescence skip must not consume an attempt"
+        );
+
+        // Leave the shared queue clean for other tests.
+        let _ = crate::util::take_failed_orderbook_publishes();
+    }
+
+    /// Once an entry has consumed its attempt budget the reconciler drops
+    /// it instead of republishing again: the retry cannot converge a relay
+    /// that keeps failing — each extra cycle would only rewrite the healthy
+    /// relays' copies with a fresh `created_at`. The residual divergence is
+    /// bounded by the event's NIP-40 expiration.
+    #[tokio::test]
+    async fn reconciler_drops_entry_after_max_attempts() {
+        let ctx = migrated_ctx().await;
+        let keys = ctx.keys().clone();
+        let _guard = crate::util::ORDERBOOK_QUEUE_TEST_LOCK.lock().await;
+
+        let order = Order {
+            id: Uuid::new_v4(),
+            kind: Kind::Sell.to_string(),
+            status: Status::Canceled.to_string(),
+            fiat_code: "USD".to_string(),
+            payment_method: "bank".to_string(),
+            expires_at: nostr_sdk::prelude::Timestamp::now().as_secs() as i64 + 3_600,
+            ..Default::default()
+        };
+        let order = order.create(ctx.pool()).await.unwrap();
+        crate::util::requeue_orderbook_publish_failure(
+            order.id,
+            1,
+            crate::util::MAX_ORDERBOOK_REPUBLISH_ATTEMPTS,
+        );
+
+        reconcile_orderbook_once(ctx.pool(), &keys).await;
+
+        assert!(
+            !crate::util::is_orderbook_publish_queued(order.id),
+            "an entry at the attempt cap must be dropped, not retried"
+        );
+        // No stamp may have been created for the dropped entry: probing
+        // with a past candidate must return it unchanged.
+        let probe = crate::util::stamp_orderbook_event(
+            order.id,
+            nostr_sdk::prelude::Timestamp::from(1_700_000_000),
+        );
+        assert_eq!(
+            probe.created_at.as_secs(),
+            1_700_000_000,
+            "a dropped entry must not be republished"
         );
     }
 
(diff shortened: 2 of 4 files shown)


grunch:
@coderabbitai review

src/scheduler.rs
```diff
@@ -1083,18 +1078,17 @@ const ORDERBOOK_FULL_SWEEP_EVERY_TICKS: u64 = 60;
 const ORDERBOOK_QUIESCENT_SECS: u64 = 120;
 
 /// One reconciler pass: republish every order whose last kind-38383
-/// publish failed (`util::take_failed_orderbook_publishes`), and — when
-/// `full_sweep` — re-assert every live `pending` order from the DB.
+/// publish failed (`util::take_failed_orderbook_publishes`).
```
grunch, line 1081: **The failed-publish queue only fills on `Err`, so it cannot be the safety net this PR credits it as.**

This doc line is now the whole convergence story, so it is worth checking what actually feeds that queue. At `util.rs:1244-1258`:

```rust
match client.send_event(&event).await {
    Ok(_) => clear_orderbook_publish_failure_up_to(order.id, stamp.generation),
    Err(e) => { warn!(...); mark_orderbook_publish_failed_at(order.id, stamp.generation); }
}
```

In nostr-sdk 0.45.1 (`Cargo.toml:49`), `send_event` resolves to `Result<SendEventOutput, Error>` where `SendEventOutput = Output<EventId, EventSendStatus, String>` (`client/api/send_event.rs:19`), and `Output` carries **both** `success: HashMap<RelayUrl, S>` and `failed: HashMap<RelayUrl, F>` (`client/api/output.rs:14-21`). `Err` is reserved for the "no relays to send to" class — `Error::not_found` and `gossip_not_configured` (`client/api/send_event.rs:301,413`), plus the pool's `NoRelaysSpecified` / `NoRelays` / `RelayNotFound`.

So a per-relay rejection or timeout lands in `Ok`, inside a `failed` map this code never reads — and that same `Ok(_)` arm additionally *clears* whatever queue entry the order already had. Relay A accepts, relay B rejects: nothing is queued, and the book stays divergent on B indefinitely.

That is precisely the gap the hourly sweep was papering over by brute force. Removing the sweep promotes it to the primary divergence path, with nothing left to heal it. The fix is short:

```rust
Ok(output) if output.failed.is_empty() =>
    clear_orderbook_publish_failure_up_to(order.id, stamp.generation),
Ok(output) => {
    tracing::warn!(
        "orderbook publish rejected by {} relay(s) for order {} (status {}): {:?}; queued for republish",
        output.failed.len(), order_updated.id, status, output.failed,
    );
    mark_orderbook_publish_failed_at(order.id, stamp.generation);
}
Err(e) => { /* unchanged */ }
```

This is pre-existing from #872, so it can land as a companion PR — but it should land with this one or before it, because this PR's risk assessment depends on it.

<sub>One caveat I will state rather than overclaim: I verified the `failed` map and the `Err` conditions in the SDK directly. The "**every** relay fails" case very likely also returns `Ok` — that is how `nostr-relay-pool` 0.44.3 behaves, which is the version I have vendored — but I could not confirm it against the exact pool 0.45.1 pins. The patch above covers both readings either way.</sub>

src/scheduler.rs
```diff
@@ -1130,62 +1124,20 @@ async fn reconcile_orderbook_once(pool: &sqlx::SqlitePool, keys: &Keys, full_swe
...
-        }
-    }
 }
 
 /// Keeps the public NIP-33 orderbook converged with the DB: drains the
-/// failed-publish queue every minute and re-asserts the whole pending book
-/// hourly (first sweep right at startup, healing relay state lost across
-/// daemon restarts). Without this job a single dropped publish
-/// leaves a dead order advertised as `pending` until its NIP-40 expiration.
+/// failed-publish queue every minute. Without this job a single dropped
```
grunch, line 1157: **This removes the only recovery for publishes lost across a restart, and the PR body does not say so.**

`PENDING_ORDERBOOK_REPUBLISH` (`util.rs:1070`) is a `LazyLock<Mutex<HashMap<Uuid, u64>>>` — process-local, and empty again after every restart. The doc line this PR deletes named exactly that case: *"first sweep right at startup, healing relay state lost across daemon restarts"*.

Two classes lose all coverage:

- **Queue lost across a restart.** A publish fails at T and is queued in memory; the daemon restarts before the next 60s drain. The entry is gone, and the order stays divergent on relays for its entire take window.
- **Silent relay loss.** A relay ACKs and later drops the event, or restores from a backup. The daemon never observed a failure, so nothing was ever queued, and nothing ever heals it.

The body states *"Everything else from #872 stays intact, including the two layers that actually keep the book converged"*. Those two layers do not cover these: the queue only covers **observed** failures, and NIP-40 caps how long a stale event lives but cannot **restore a missing one**.

To be clear, I think this is a defensible trade — the author of #872 confirms the re-publication was never intended, the harms you list are real, and NIP-40 bounds the blast radius to the order's actual take window. My ask is narrow: state the trade explicitly here and in the body, instead of leaving it implicit behind "everything else stays intact". The deleted comment is the only place it was written down.

src/db.rs
```diff
@@ -675,34 +675,6 @@ pub async fn find_order_by_date(pool: &SqlitePool) -> Result<Vec<Order>, MostroE
...
 
-/// All orders currently advertised (or advertisable) as `pending` on the
-/// book. Includes `waiting-taker-bond`, which publishes as `pending` on the
-/// wire (Phase 1.5), and excludes `waiting-maker-bond`, which was never
-/// published. Used by the orderbook reconciler to re-assert the book on
-/// relays; only orders whose take window is still open are returned — an
-/// expired row is the expiry job's business, not the reconciler's. Legacy
-/// rows with `expires_at = 0` have no take-window TTL (mirroring
-/// `is_order_take_window_closed`) and stay re-assertable.
-pub async fn find_pending_orders_for_reconcile(
```
grunch, line 686: **The `created_at` churn has a smaller root cause than the sweep.**

The first and strongest justification in the body is that every re-assert stamps a fresh `created_at` *because the monotonic registry is in-memory*. That is accurate, but the sweep is the symptom rather than the cause:

- `orders.created_at` **is a persisted column** — visible in the `INSERT` of the very test helper this PR removes;
- `ORDERBOOK_REGISTRY.last_ts` (`util.rs:960-976`) simply starts empty, and its own doc admits it: *"Across a restart the map starts empty"*;
- `update_order_event_stamped` (`util.rs:1219-1223`) always passes `Timestamp::now()` as the candidate, in **both** `StampPolicy` arms, so nothing ever reuses the persisted value.

Seeding the registry from `orders.created_at` at startup — or reusing the persisted `created_at` when the published status has not changed — makes a re-assert **idempotent**: same `d` tag, same content, same `created_at` -> same event id -> relays dedupe it outright. That kills all three harms the body lists (the churn, the 24x/day traffic, and the `orders.event_id` drift, which disappears because the id no longer changes) *without* giving up the healing.

Caveat I will name myself: `order_to_tags` includes reputation data via `get_ratings_for_pending_order`, which can legitimately change between re-asserts and would produce a new id. That is a real update, not churn, so it does not undermine the approach.

Even if the deletion is still the preferred outcome, this deserves a sentence in the body on why the idempotent-re-assert path was rejected — right now the body reads as though re-publication is inherently churn-producing, and it is not.

src/scheduler.rs
```diff
@@ -1649,7 +1601,7 @@ mod tests {
         let ghost = Uuid::new_v4();
         crate::util::mark_orderbook_publish_failed(ghost);
 
-        reconcile_orderbook_once(ctx.pool(), &keys, false).await;
+        reconcile_orderbook_once(ctx.pool(), &keys).await;
```
grunch, line 1640: **No test pins the newly documented invariant.**

The three retained tests all exercise the queue drain — ghost order, requeue while relays are unreachable, quiescence guard. None of them asserts the guarantee this PR newly documents at `scheduler.rs:1132`: *"Orders whose publishes succeeded are never re-sent"*.

A cheap regression test closes that: seed a live `pending` order that is **not** in the failed-publish queue, run `reconcile_orderbook_once`, and assert no stamp was created for it. The existing test helpers already reach far enough — `is_orderbook_publish_queued` plus the registry are enough to observe it.

Deletions are the changes that regress most quietly. Without a test, nothing stops someone from re-adding a periodic sweep in six months for what will look like a perfectly good reason.

src/scheduler.rs
```diff
@@ -1130,62 +1124,20 @@ async fn reconcile_orderbook_once(pool: &sqlx::SqlitePool, keys: &Keys, full_swe
...
 async fn job_orderbook_reconciler(ctx: AppContext) {
     let keys = ctx.keys().clone();
     tokio::spawn(async move {
         let pool = ctx.pool();
-        let mut tick: u64 = 0;
         loop {
-            let full_sweep = tick.is_multiple_of(ORDERBOOK_FULL_SWEEP_EVERY_TICKS);
-            reconcile_orderbook_once(pool, &keys, full_sweep).await;
-            tick = tick.wrapping_add(1);
+            reconcile_orderbook_once(pool, &keys).await;
```
grunch: **The first pass at startup is now a guaranteed no-op.** (nit)

The loop calls `reconcile_orderbook_once` *before* its first sleep. That ordering was meaningful while `tick == 0` triggered the startup full sweep; with the sweep gone, the in-memory queue is always empty at process start, so this first call can only ever do nothing.

Harmless either way — but moving the sleep to the top of the loop, or leaving a one-line comment, saves the next reader from working out what the startup pass is supposed to accomplish.

grunch (changes requested):
Reviewed the removal against what it leaves behind, rather than the diff in isolation.

**The removal hygiene is clean.** I grepped the head commit for every symbol this PR deletes — `find_pending_orders_for_reconcile`, `ORDERBOOK_FULL_SWEEP_EVERY_TICKS`, `insert_book_order`, `full_sweep` — and there is not one dangling reference. `Timestamp` is still used in `db.rs` (`647`, `681`, `1367`), so no unused import is left behind. The doc comments were updated consistently, down to renaming the test prose from *sweep* to *republish*. The `created_at` churn diagnosis is correct and well argued.

The findings are about **the premise**, not the deletion. The PR justifies removing the sweep on the grounds that the failed-publish queue is "what heals ghost orders". That claim does not hold.

### Findings

| # | Severity | Where | What |
|---|---|---|---|
| 1 | 🟠 High | `util.rs:1244` (via `scheduler.rs:1081`) | The failed-publish queue only fills on `Err`, so it cannot be the safety net this PR credits it as |
| 2 | 🟡 Medium | `scheduler.rs:1130` | Removes the only recovery for publishes lost across a restart; the body does not say so |
| 3 | 🟡 Medium | `db.rs:686` (removed) | The `created_at` churn has a smaller root cause than the sweep |
| 4 | 🔵 Low | `scheduler.rs:1604` | No test pins the newly documented invariant |
| 5 | 🔵 Low | `scheduler.rs:1140` | The first pass at startup is a guaranteed no-op |

<details>
<summary>✅ Verified correct — spot-checked, no action needed</summary>

- **The NIP-40 argument holds for the whole published book.** `nip40_expiration_for_order` (`nip33.rs:445`) only shortens the expiration when the *published* status is `Pending` — and `create_status_tags` (`nip33.rs:330`) maps `WaitingTakerBond` to wire `Status::Pending`, so `waiting-taker-bond` orders are capped at their real `expires_at` too, not at the 30-day retention. There is no hole there.
- **The queue re-arm on a failed republish is correct.** `take_failed_orderbook_publishes()` drains the map, but `update_order_event_stamped` re-queues internally on send failure before returning `Ok(Some(..))`, so the reconciler's `Ok(Some(_)) => {}` is not dropping the entry. The "requeue while relays unreachable" test covers exactly this.
- **The `Err(e)` branch in `reconcile_orderbook_once` cannot double-mark.** An `Err` out of `update_order_event_if_quiescent` can only come from tag building (`order_to_tags` / nostr error), which runs *before* the send, so the inner `mark_orderbook_publish_failed_at` never fired for that pass.
- **The quiescence guard is untouched and still load-bearing.** `ORDERBOOK_QUIESCENT_SECS` and `try_stamp_orderbook_event_quiescent` remain wired into the queue-drain path, so the publish-before-CAS race #872 closed stays closed.

</details>

**Verdict: REQUEST_CHANGES** — on finding 1, not on the deletion. The removal itself is well executed and well argued; what does not hold is the risk assessment around it, because the queue this PR names as the safety net does not capture per-relay rejections. With those ~4 lines of `output.failed` handling fixed (here or as a companion PR) and the finding-2 trade-off stated in the body, this is ready to approve.

src/util.rs
```diff
@@ -498,7 +498,22 @@ async fn finalize_order_publication(
...
+            if !output.failed.is_empty() {
+                tracing::warn!(
+                    "initial orderbook publish rejected by {} relay(s) for order {}: {:?}; queued for republish",
+                    output.failed.len(),
+                    order_id,
+                    output.failed
+                );
+                mark_orderbook_publish_failed(order_id);
+            }
+        })
```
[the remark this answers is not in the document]
Catrya: Intentional and documented — this is the accepted trade-off stated in the PR
body and in the reconciler's doc comment, per the maintainer's review ("a
defensible trade; my ask is narrow: state it explicitly").

Two precisions: the removed startup sweep only re-asserted *pending* rows, so
terminal-status publish failures lost across a restart were never recovered by
it either — that class is unchanged by this PR and bounded by NIP-40, as
before. And durable queue state alone is the wrong shape for the follow-up: if
a migration is ever paid for this, it should persist the last published stamp
instead, which yields idempotent re-asserts (relays dedupe by event id) and
covers this case as a side effect. That is the follow-up already named in the
PR body.

src/util.rs
```diff
@@ -1245,7 +1260,25 @@ async fn update_order_event_stamped(
...
+                Ok(output) if output.failed.is_empty() => {
+                    clear_orderbook_publish_failure_up_to(order.id, stamp.generation)
+                }
+                // A per-relay rejection or timeout resolves to `Ok` with the
+                // refusing relays in `output.failed` — `send_event` returns
+                // `Err` only when there was no relay to send to at all. Any
+                // rejected relay means the book is divergent there, so the
+                // publish counts as failed and stays queued until a
+                // republish converges it.
+                Ok(output) => {
```
grunch, line 1320: **A single persistently-failing relay turns the queue into an unbounded republish loop — reintroducing this PR's own headline harm at 30x the rate.**

This is the fix to my own earlier comment, and it is correct as far as it goes: all three feeders now read `output.failed`. But `failed` is populated by *any* per-relay error, and the queue has no attempt cap, no backoff, and no terminal state. That closes a loop I did not flag and should have.

`nostr-relay-pool` (`pool/mod.rs:789-798`) inserts into `failed` on every `Err` from `relay.send_event` — a relay that is offline, timing out, rate-limiting, or refusing kind 38383 all land there. The overall result is `Ok` regardless, even when *every* relay failed.

Trace one offline relay through the new code:

1. Any transition publishes -> `output.failed` non-empty -> `mark_orderbook_publish_failed_at` (`util.rs:1281`).
2. The reconciler drains 60s later (`scheduler.rs:1092`) and calls `update_order_event_if_quiescent`.
3. `try_stamp_orderbook_event_quiescent` (`util.rs:1050-1060`) stamps a **fresh `Timestamp::now()`** as soon as the order has been quiet for 120s.
4. New `created_at` -> new event id -> the healthy relays replace their copy. The offline relay fails again -> requeued at step 1.

The loop ends only when the bad relay returns. While it runs:

- **`created_at` churns every 120s per queued order** — 720 revisions/day, against the 24/day this PR removes. Harm #1 of the PR body, amplified 30x.
- **`orders.event_id` drifts** — the reconciler deliberately does not persist the republished id (`scheduler.rs:1085-1088`). Harm #3, back.
- **The republish converges nothing.** It cannot reach the relay that failed; it only rewrites the copies that were already correct.
- **`PENDING_ORDERBOOK_REPUBLISH` (`util.rs:1085`) never shrinks** during the outage. It has no pruning, and an entry clears only on a fully-successful publish, so the working set grows with every order touched while the relay is down.

Before this commit the loop was reachable only through `Err` — no relay at all — where nothing is published and nothing churns. Treating partial failure as failure is the right call, but it moves the loop into the case where events *do* go out.

Bounding it is cheap. Carry an attempt count in the queue value and drop the entry after a few tries with a loud warning; NIP-40 already bounds the residual divergence, which is exactly the trade-off this PR accepts elsewhere:

```rust
// util.rs
static PENDING_ORDERBOOK_REPUBLISH:
    LazyLock<Mutex<HashMap<Uuid, (u64, u8)>>> = ...;   // (generation, attempts)
const MAX_ORDERBOOK_REPUBLISH_ATTEMPTS: u8 = 5;
```

with the reconciler dropping — and warning about — an order that exceeds it. Per-order exponential backoff would work equally well.

A narrower alternative that also closes it: queue only when `output.success.is_empty()`, i.e. the event reached no relay at all. That gives up the partial-divergence healing I asked for, but it is strictly better than the current unbounded state and rests on the same NIP-40 argument.

grunch (changes requested):
## Follow-up review — all five prior comments verified against `2b68eab`

Re-reviewed at `2b68eab0`, with `86f00b43` as the baseline my earlier review targeted. Every one of the five points is addressed; one of the fixes opens a new issue, which is the single item below.

### Prior comments

| # | Comment | Status | Evidence |
|---|---------|--------|----------|
| 1 | Queue only fills on `Err`, so it cannot be the safety net | **Fixed** | All three feeders now branch on `output.failed`: `util.rs:1263-1281` (status publish), `util.rs:501-516` (initial publish), `release.rs:284-300` (range child). The `Ok(_)` arm no longer clears an entry on a partially-rejected send. |
| 2 | Restart/silent-loss trade-off left implicit | **Fixed** | Reconciler doc gains an explicit *"Accepted trade-off"* paragraph (`scheduler.rs:1135-1141`) naming process-locality, observed-failures-only, and the NIP-40 bound; the PR body gains a matching section. |
| 3 | `created_at` churn has a smaller root cause than the sweep | **Fixed** (as asked) | The body now states why the idempotent-re-assert path was rejected — it needs persisted per-order publish state — and names it as a follow-up rather than leaving re-publication implied to be inherently churn-producing. Consistent with the reply to CodeRabbit's durability finding. |
| 4 | No test pins the newly documented invariant | **Fixed** | `reconciler_never_republishes_order_without_failed_publish` (`scheduler.rs:1691-1731`). The probe is sound: it re-stamps with a 2023 candidate and asserts it comes back unchanged, which can only hold if the pass created no registry entry for the order. The order id is freshly generated, so no cross-test contamination. |
| 5 | First pass at startup is a guaranteed no-op (nit) | **Fixed** | Loop now sleeps before its first pass, with a comment saying why (`scheduler.rs:1146-1153`). |

Also confirmed: no dangling references to `find_pending_orders_for_reconcile`, `ORDERBOOK_FULL_SWEEP_EVERY_TICKS`, or the `full_sweep` branch anywhere in `src/` or the docs.

### Local verification

- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --bin mostrod` — **1199 passed, 0 failed, 2 ignored**

### New finding

One blocking item, inline on `src/util.rs`. It is a consequence of the fix to comment #1 — my suggested patch was right about reading `failed` and wrong to stop there, since the queue it feeds has no attempt cap. With `failed` now populated by any per-relay error, one offline or rate-limiting relay puts every touched order into a permanent 120s republish cycle, each cycle stamping a fresh `created_at` and drifting `orders.event_id`. That is harm #1 and harm #3 from this PR's own body, at 30x the rate of the sweep being removed, and the republish cannot converge the relay that failed. A retry cap or backoff closes it in a few lines.

The rest of the PR is in good shape and the direction is right — removing the sweep is well argued and now well documented.

grunch (approved):
## Follow-up review at `cd55800` — all prior comments verified fixed

Re-reviewed the full thread against the head commit: both of @grunch's rounds (5 findings + the 1 new blocking finding) and CodeRabbit's durability finding. Everything is addressed; local checks pass.

### Prior comments

| # | Comment | Status | Evidence |
|---|---------|--------|----------|
| 1 | Queue only fills on `Err`, so it cannot be the safety net | **Fixed** (`2b68eab`) | All three feeders now branch on `output.failed`: `update_order_event_stamped`, `finalize_order_publication`, and the range child-order publish in `release.rs`. The `Ok` arm no longer clears an existing entry on a partially-rejected send. |
| 2 | Restart/silent-loss trade-off left implicit | **Fixed** | Explicit *Accepted trade-off* paragraph in the `job_orderbook_reconciler` doc comment plus a matching PR body section. |
| 3 | `created_at` churn has a smaller root cause than the sweep | **Fixed** (as asked) | The body now states why the idempotent re-assert path was rejected (requires persisted per-order publish state) and names it a follow-up. |
| 4 | No test pins the newly documented invariant | **Fixed** | `reconciler_never_republishes_order_without_failed_publish`. The past-candidate probe is sound: it can only come back unchanged if the pass created no registry entry for the order. |
| 5 | First pass at startup is a guaranteed no-op | **Fixed** | The loop sleeps before its first pass, with a comment saying why. |
| CR | Persist failed-publish state across restarts (CodeRabbit) | **Withdrawn** | Withdrawn by CodeRabbit after the author's reply; consistent with the trade-off accepted in finding 2. |
| 6 | Unbounded republish loop via `output.failed` — no attempt cap | **Fixed** (`cd55800`) | See below. |

### The attempt cap (`cd55800`)

Traced the accounting through the queue:

- Entries are now `(generation, attempts)`: `mark_orderbook_publish_failed_at` increments (saturating), `requeue_orderbook_publish_failure` merges without downgrading either field.
- The reconciler re-seeds the drained entry **before** attempting the republish, so the send path's own failure marking increments on top of the consumed attempts — the cap can actually trip. A quiescent skip or a transient DB reload error requeues without consuming an attempt; a pre-send (tags/ratings/event build) error consumes one, so a permanently broken order cannot occupy the queue forever.
- At `MAX_ORDERBOOK_REPUBLISH_ATTEMPTS` (5) the entry is dropped with a loud warning naming the relay list and the NIP-40 bound. Since the drain already removed it, the warning fires exactly once per episode — no 60 s log spam.
- A full success still clears generation-aware, and the next failure episode starts with a fresh budget. All three behaviors are pinned by tests: `reconciler_failed_retry_consumes_exactly_one_attempt`, `reconciler_drops_entry_after_max_attempts`, `republish_attempts_count_failures_not_requeues`.

The churn traced in finding 6 is now bounded to at most 4 republish revisions per failure episode (each gated by the 120 s quiescence window), with the residual divergence capped by NIP-40 — the same trade-off this PR already accepts for silent relay loss.

### Also verified

- No dangling references to `find_pending_orders_for_reconcile`, `ORDERBOOK_FULL_SWEEP_EVERY_TICKS`, `insert_book_order`, or the `full_sweep` branch anywhere in `src/` or `docs/`.
- Every queue-touching test serializes on `ORDERBOOK_QUEUE_TEST_LOCK`.
- Edge case considered and dismissed: a queued order reaching a non-publishable status (`create_event == false`) would leave its entry as a 60 s no-op loop — practically unreachable, since the transition's own publish resolves the entry first, and harmless if reached (no stamp, no relay traffic).

### Local verification (at `cd55800`)

- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test --bin mostrod` — **1202 passed, 0 failed, 2 ignored**

**Verdict: approve.** The removal is well argued, both review rounds are fully addressed, and the attempt cap closes the one regression the `output.failed` fix had opened.

PR #890, opened 2026-08-17
fix: set explicit permissions on the LND admin macaroon

AndreaDiazCorreia:
## Summary

The `docker-build` target and every documented deployment flow copied the LND admin macaroon with plain `cp`, which keeps whatever mode the source file (or an already existing destination file) happens to have. Nothing set the mode explicitly and nothing checked it, so a spend-capable credential could sit world-readable with no output saying so.

## Changes

- `Makefile`: `install -m 600` for the macaroon, `install -m 644` for the cert, `install -d -m 700` for `config/lnd`, plus a note about the uid/gid 1000 the container runs as.
- Docs (`docker/README.md`, `docker/ENV_VARIABLES.md`, `README.md`, `INSTALL.md`, `AGENTS.md`): `install -m` instead of `cp` in all three flows, with the ownership each one needs so tightening the mode does not lock mostrod out of the file.
- `src/config/permissions.rs` (new): startup check that warns when the macaroon is reachable by users outside its owner and group. Advisory only, and tolerates `0640`, the mode LND itself writes the macaroon with.

## Verification

- `cargo test`: 1207 passed, 0 failed, 2 ignored (8 new tests).
- `cargo clippy --all-targets --all-features`, `cargo fmt --check`, `markdownlint-cli2@0.23.2`: clean.
- `make docker-build` against a `0644` source and against a preexisting `0644` destination: both end at `0600`, directory at `0700`.

No config keys added or changed.

## Summary by CodeRabbit

- **Security Enhancements**
  - Added startup warnings when LND admin credentials are accessible to other local accounts.
  - Configuration directories and files now use owner-only permissions and reject unsafe existing files or symlinks.
  - Docker builds exclude staged credentials and environment files.
  - Docker setup supports configuring container user and secure file ownership.

- **Documentation**
  - Expanded guidance for protecting settings, databases, and spend-capable Lightning credentials.
  - Documented secure installation commands, permissions, ownership, and deployment options.


AndreaDiazCorreia:
@codex review

src/config/permissions.rs
```diff
@@ -0,0 +1,362 @@
...
+    builder.recursive(true);
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::DirBuilderExt;
+        builder.mode(0o700);
+    }
+    builder
+        .create(settings_dir)
+        .map_err(|e| MostroInternalErr(ServiceError::IOError(e.to_string())))
+}
```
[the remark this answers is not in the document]
AndreaDiazCorreia: Checked this against the actual behaviour before deciding. Three cases, only one of which is a real gap:

- **Settings dir is a symlink to an existing directory** — `settings_dir.exists()` returns true, so `create_settings_dir` is never called. The caller uses that directory, as it did before this PR.
- **Settings dir is a dangling symlink** — `DirBuilder::create` already fails with `AlreadyExists`. Nothing is created through the link.
- **An intermediate component is a symlink to an existing directory** — confirmed, the recursive builder resolves it and the settings directory lands at the link's target.

So the finding reduces to the third case, and we are leaving it as is.

Exploiting it requires the ability to create a symlink at a component of the settings directory path. For the default `~/.mostro` that component is `$HOME`: an account that can write there replaces `.bashrc` or `.ssh/authorized_keys` and owns the session outright, so this path adds nothing to what it already has. For an operator-supplied `-d <path>`, the path is the operator's own choice. That is a materially weaker threat model than the `settings.toml` TOCTOU fixed earlier in this PR, where write access to the settings directory alone was enough — plausible in group-shared deployments — and the fix cost nothing.

The fix asked for is not free either. `openat`/`mkdirat` with `O_NOFOLLOW | O_DIRECTORY` are not in Rust's std, so this means a new dependency (`libc`, `rustix` or `cap-std`) for a permissions helper, and rejecting symlinked components breaks a legitimate and common setup: pointing a config directory at another volume through a symlink.

What we did instead is scope the documented guarantee honestly, so nobody reads more into it than it provides — `docs/STARTUP_AND_CONFIG.md` and the doc comment on `create_settings_dir` now state that the no-follow guarantee covers the final `settings.toml` entry and not the directory path leading to it, and why.

grunch (commented):
## Strict review — PR #890

Traje la rama localmente, la compilé y verifiqué los checks antes de opinar.

**Verificación local (rama `pr890`, commit `38085ad`)**

| Check | Resultado |
|---|---|
| `cargo test --bin mostrod permissions` | 18 passed, 0 failed |
| `cargo test --bin mostrod save_settings` | 4 passed, 0 failed |
| `cargo clippy --all-targets --all-features` | limpio |
| `cargo fmt --check` | limpio |

También verifiqué dos cosas que parecían hallazgos y **no lo son**: (a) el chequeo va después del `return run_cashu(ctx)` en `main.rs:222`, pero `run_cashu` no instancia `LndConnector` (`src/app.rs:503`), así que saltárselo es correcto; (b) `config` a `0700` no rompe al servicio `nostr-relay`, porque sus bind mounts (`./config/relay/config.toml`, `./config/relay/data`) se montan directamente en el namespace del contenedor y no requieren atravesar el directorio padre del host.

---

## 🔴 Alta

### 1. `src/config/wizard.rs:252` — el `.env` con el mismo `nsec` sigue siendo vulnerable al symlink que la PR arregla

`create_owner_only` cierra el hueco para `settings.toml` con `O_CREAT|O_EXCL`, y `docs/STARTUP_AND_CONFIG.md` declara la invariante ("never through a symlink" para el nsec). Pero `write_env_file` sigue con `create(true).truncate(true)`: sigue symlinks y trunca el destino, reseteándolo a `0600`. Y es peor de lo que parece por el orden de ejecución: en el wizard, `prompt_nostr_settings` → `write_env_file` corre **antes** de `save_settings`, así que el escenario exacto que la PR describe (otra cuenta local con escritura en el directorio de settings) sigue abierto por la ruta del `.env` — que contiene el mismo secreto.

```rust
// src/config/wizard.rs:252
std::fs::OpenOptions::new()
    .write(true)
    .create(true)
    .truncate(true)
    .mode(0o600)
+   .custom_flags(libc::O_NOFOLLOW)   // o migrar a create_owner_only con manejo del caso "ya existe"
    .open(path)
```

No se puede usar `create_owner_only` tal cual porque `write_env_file` sí necesita reescribir un `.env` existente (hay un test que lo cubre: `test_write_env_file_tightens_preexisting_broader_permissions`). `O_NOFOLLOW` es el mínimo. Si no se arregla en esta PR, hay que quitar la afirmación absoluta de `docs/STARTUP_AND_CONFIG.md`, porque hoy documenta una garantía que el `.env` no cumple.

---

## 🟠 Media

### 2. `Makefile:43,53` — `install -d -m 700 config` revierte permisos deliberados en cada `make docker-up`

GNU/BSD `install -d` aplica el modo también a directorios existentes (la propia PR lo usa como feature en `docker/README.md`). El efecto en `docker-up`/`docker-relay-up` es que un `config` que el operador dejó en `0750` con un grupo se le vuelve `0700` en cada arranque, en silencio. Eso contradice la política que la propia PR declara para el lado Rust: *"An existing settings directory is left as it is, so a deliberately group-readable deployment keeps working"*. Dos reglas opuestas para el mismo directorio.

```diff
-	install -d -m 700 config && \
-	mkdir -p config/relay && \
+	[ -d config ] || install -d -m 700 config; \
+	mkdir -p config/relay && \
```

### 3. `src/config/permissions.rs:90` — `0700` se propaga a todos los ancestros creados, no solo a la hoja

`DirBuilder::mode(0o700)` + `recursive(true)` aplica el modo a cada directorio que crea. El propio test lo fija: `settings_dir_creation_is_recursive_and_owner_only` afirma `mode_of(&root.join("nested")) == 0o700`. Con `mostrod -d /srv/apps/mostro/conf` sobre un árbol nuevo, `/srv/apps` y `/srv/apps/mostro` quedan `0700` propiedad del daemon, lo que puede dejar fuera a otros consumidores del árbol. Para `~/.mostro` da igual; para el flag `-d` no. O se crean los padres con el umask normal y se aplica `0700` solo a la hoja, o hay que documentarlo explícitamente (hoy `STARTUP_AND_CONFIG.md` solo habla de symlinks en el path, no de esto).

### 4. `src/main.rs:230` — el aviso de arranque cubre el macaroon pero no lo que las guías te hacen crear a mano

`warn_if_other_accessible` ya es genérica (recibe `label`). `settings.toml` y `<dir>/.env` llevan `nsec_privkey` en claro y son precisamente los archivos que los flujos documentados crean con `cp`/`curl`/editor — el caso que motiva toda la PR. Un `settings.toml` en `0644` arranca sin decir nada. Son dos líneas:

```diff
 permissions::warn_if_other_accessible(&Settings::get_ln().lnd_macaroon_file, "LND admin macaroon");
+permissions::warn_if_other_accessible(&config_file_path, "Mostro settings file");
```

(La ruta hay que exponerla desde `init_configuration_file`, que hoy no la devuelve.)

### 5. `Makefile` / `docker/compose.yml:13` — `MOSTRO_CONTAINER_USER` se avisa solo en `docker-build`, y falla tarde

El aviso se imprime en `docker-build`, pero la variable la consume `docker-up`. Un operador con uid distinto de 1000 que corra `make docker-up` en otra shell obtiene un `EACCES` al leer el macaroon `0600`, que aflora como fallo de conexión a LND — un síntoma que no apunta a permisos. Y el nuevo chequeo de arranque tampoco lo cubre (el warning solo mira bits `other`, no si *este* proceso puede leer). Lo más simple es derivarlo en el Makefile en vez de documentarlo:

```make
docker-up:
	@set -o pipefail; \
	cd docker && \
	export MOSTRO_CONTAINER_USER="$${MOSTRO_CONTAINER_USER:-$$(id -u):$$(id -g)}" && \
```

---

## 🟡 Baja / nits

6. **`Makefile:31` — ventana de exposición en el propio `install`.** Si `config/lnd/admin.macaroon` ya existe en `0644` (de un build previo con el flujo `cp` viejo), `install` trunca y escribe el contenido nuevo y solo hace el `chmod` al final: el macaroon nuevo pasa unos milisegundos world-readable. Borrar el destino antes del `install`, o instalar a nombre temporal + `mv`, lo cierra.

7. **`docker/README.md` (2 sitios) — `curl … -o file` + `chmod 600` tiene la misma ventana**, y la PR incluso la señala en prosa ("`curl` creates the file under your umask… so the mode has to be set afterwards"). Se elimina, no se documenta: `(umask 077 && curl -fsSL … -o ~/mostro-config/settings.toml)`.

8. **`src/config/permissions.rs:120` — escritura no atómica.** Si `write_all` falla a mitad (disco lleno), queda un `settings.toml` truncado en `0600`; el siguiente arranque lo parsea y muere con "malformed TOML" en vez de recrearlo. Temporal en el mismo directorio (con `create_new`, que conserva la garantía anti-symlink) + `rename` lo hace atómico.

9. **`docker/start.sh:9`** — `[ -f /config/settings.toml ]` es falso para un symlink colgante y el `install` escribiría a través de él. Misma clase de bug que la PR arregla en Rust; marginal dentro del contenedor, pero es incoherente dejarlo.

10. **Tests: `std::env::temp_dir()` + PID como nombre predecible**, en tres módulos, con `temp_root`/`mode_of` duplicados tres veces. En un CI compartido `/tmp` es escribible por todos, así que los tests de una PR sobre permisos son ellos mismos plantables por symlink. `tempfile` no está en `[dev-dependencies]`; añadirlo y extraer los helpers a un módulo común resuelve las dos cosas.

11. **`#[cfg(all(test, unix))]` en ambos módulos de test** — `create_owner_only` tiene rama non-unix y cero cobertura. Al menos un test portable que verifique creación + rechazo de preexistente.

12. **`src/config/permissions.rs:61`** — mezcla `use std::fs;` con `std::fs::metadata(...)` totalmente cualificado en el mismo archivo.

13. **`README.md`** afirma el warning de arranque sin decir que solo ocurre en modo Lightning; `STARTUP_AND_CONFIG.md` sí lo aclara.

14. **Alcance vs. título.** "set explicit permissions on the LND admin macaroon" no anticipa el `.dockerignore` nuevo, el `user:` en compose (cambio de contrato de despliegue) ni la refactorización de la creación de `settings.toml`. Vale la pena reflejarlo en el título o en las release notes.

---

## Lo que está bien hecho

- El razonamiento de `O_CREAT|O_EXCL` es correcto y está bien justificado en el comentario, incluido el `fchmod` posterior para derrotar al umask (`OpenOptionsExt::mode` sí se enmascara).
- Tolerar `0640` en vez de exigir `0600` es la decisión acertada: es el modo con el que LND escribe el macaroon, y exigir `0600` habría roto el patrón legítimo de acceso por grupo.
- Advertir en vez de abortar evita convertir una brecha de hardening en un outage en la siguiente actualización.
- El `.dockerignore` es un hallazgo real y no obvio: `COPY . .` con contexto en la raíz metía el macaroon staged en la capa del builder y su caché, y `.gitignore` no aplica ahí.
- Los tests de symlink/preexistente verifican que la víctima queda intacta en contenido **y** modo, no solo que la llamada falla.

**Veredicto: aprobar con cambios.** El único bloqueante real es el #1 — o se arregla `write_env_file` o se corrige la garantía documentada, porque tal como está la PR afirma una invariante sobre el nsec que su propio código no cumple.

Makefile
```diff
@@ -27,24 +42,32 @@ docker-build:
...
 	docker compose build
 
 docker-up:
 	@set -o pipefail; \
 	cd docker && \
 	echo "Copying Nostr relay config" && \
+	{ [ -d config ] || install -d -m 700 config; } && \
 	mkdir -p config/relay && \
 	cp -v ./relay_config.toml config/relay/config.toml && \
+	export MOSTRO_CONTAINER_USER="$${MOSTRO_CONTAINER_USER:-$$(stat -c '%u:%g' config 2>/dev/null || stat -f '%u:%g' config)}" && \
```
grunch, line 86: _🔒 Security & Privacy_ | _🟠 Major_

**A root-owned `docker/config` makes the container run as root.**

Before this PR the container always ran as `mostrouser` (`USER mostrouser` in the Dockerfile). With this line, whoever owns `docker/config` becomes the container user — and on a host where Docker needs root, `sudo make docker-build && sudo make docker-up` is the common path (the VPS section of `docker/README.md` itself assumes root for every step). `install -d -m 700 config` then creates the directory as `0:0`, `stat` yields `0:0`, and `mostrod` runs as **root inside the container**. The line prints it, but nothing refuses it, so a hardening PR silently drops the one privilege boundary the image had.

A bare `docker compose up` in that scenario fails instead (uid 1000 cannot enter a root-owned `0700` dir), so the two documented entry points now diverge on exactly the sensitive case.

<details>
<summary>Proposed fix — refuse uid 0 rather than run with it</summary>

```make
	export MOSTRO_CONTAINER_USER="$${MOSTRO_CONTAINER_USER:-$$(stat -c '%u:%g' config 2>/dev/null || stat -f '%u:%g' config)}" && \
	if [ "$${MOSTRO_CONTAINER_USER%%:*}" = 0 ]; then \
		echo "docker/config is owned by root; the container will not run as root." >&2; \
		echo "Run: chown -R 1000:1000 docker/config   (or export MOSTRO_CONTAINER_USER=uid:gid)" >&2; \
		exit 1; \
	fi && \
```

Alternatively fall back to `1000:1000` and `chown -R 1000:1000 config` when running as root, which is what the VPS flow documents by hand. Either way `docker/README.md` / `ENV_VARIABLES.md` should say what happens when the directory is root-owned.
</details>

src/config/permissions.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    #[test]
+    fn atomic_write_steps_over_a_stale_temporary() {
+        let root = temp_dir("owner-only", "atomic-stale");
+        let env_file = root.join(".env");
+        // What a run killed mid-write leaves behind. Writing through it would
+        // be harmless here, but the same name could just as well be a symlink
+        // another account planted, so the first free suffix is used instead.
+        let stale = root.join(format!(".env.tmp-{}-0", std::process::id()));
```
grunch: _🧪 Test Coverage_ | _🟡 Minor_

**This test never exercises the retry it claims to cover.**

`create_temp_sibling` builds the candidate as `"." + file_name + ".tmp-{pid}-{n}"`, which for `.env` is `..env.tmp-<pid>-0` (two leading dots). The seed here is `.env.tmp-<pid>-0` — a different name — so attempt 0 never collides and the "step over" path is never taken.

Verified by mutation: with `TEMP_NAME_ATTEMPTS = 1` (retry disabled) this test still passes; with the seed corrected to `..env.tmp-…` it fails under `TEMP_NAME_ATTEMPTS = 1` and passes under `16`, which is the behaviour the test is meant to pin.

<details>
<summary>Suggested fix</summary>

Either fix the seed:

```diff
-        let stale = root.join(format!(".env.tmp-{}-0", std::process::id()));
+        let stale = root.join(format!("..env.tmp-{}-0", std::process::id()));
```

or, better, extract the name construction into a `fn temp_sibling_name(file_name: &OsStr, attempt: u32) -> OsString` used by both `create_temp_sibling` and the test, so the two cannot drift apart again.
</details>

src/config/permissions.rs
```diff
@@ -0,0 +1,623 @@
...
+
+    for attempt in 0..TEMP_NAME_ATTEMPTS {
+        let mut name = OsString::from(".");
+        name.push(file_name);
+        name.push(format!(".tmp-{}-{attempt}", std::process::id()));
+        let candidate = dir.join(name);
+
+        match open_owner_only_new(&candidate) {
+            Ok(file) => return Ok((candidate, file)),
+            Err(e) => last_error = Some(e),
```
grunch: _🎯 Functional Correctness_ | _🟡 Minor_

**Retry only on `AlreadyExists`.**

The loop retries on every error kind. On an unwritable settings directory (`EACCES`), a missing one (`ENOENT`) or a full disk it performs 16 identical failing `open`s before reporting — and the eventual message says "after 16 attempts", which points the operator at name collisions rather than at the real cause.

```diff
         match open_owner_only_new(&candidate) {
             Ok(file) => return Ok((candidate, file)),
-            Err(e) => last_error = Some(e),
+            Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => last_error = Some(e),
+            Err(e) => {
+                return Err(MostroInternalErr(ServiceError::IOError(format!(
+                    "Could not create a temporary file in {}: {e}",
+                    dir.display()
+                ))))
+            }
         }
```

Makefile
```diff
@@ -1,10 +1,25 @@
...
+# - The mostro container has to run as whoever owns `config`: the macaroon
+#   there is 0600 and the daemon also writes mostro.db beside it. Deriving the
+#   uid:gid from the directory covers both the default (docker-build created it
+#   as you) and the documented `chown -R 1000:1000` handover.
+
 docker-build:
 	@set -o pipefail; \
 	cd docker && \
-	mkdir -p config/lnd && \
+	install -d -m 700 config config/lnd && \
```
grunch: _📐 Maintainability & Code Quality_ | _🟡 Minor_

**`docker-build` still re-tightens an existing `config` on every rebuild.**

The comment block above states the policy — "only the run that first creates the directory gets to decide its mode" — and `docker-up`/`docker-relay-up` follow it, but this line applies `0700` unconditionally. An operator who opened `config` to a group and later runs `make docker-build` for a version bump gets it closed again, silently. `docker/README.md` step 2 also promises "`make docker-up` will not decide it again" without mentioning that `docker-build` will.

```diff
-	install -d -m 700 config config/lnd && \
+	{ [ -d config ] || install -d -m 700 config; } && \
+	install -d -m 700 config/lnd && \
```

(`config/lnd` is owned by this target, so tightening it every time is fine.)

docker/README.md
```diff
@@ -123,28 +139,36 @@ Steps to run the plain Mostro image on a VPS (no repo clone; image from Docker H
...
+4. **Put LND files** in the config dir so they appear at `/config/lnd/` in the container. Use `install -m` rather than `cp`, which would keep whatever mode the source file (or an already existing destination file) happens to have:
 
    ```sh
-   cp /path/to/lnd/tls.cert /opt/mostro/lnd/tls.cert
-   cp /path/to/lnd/admin.macaroon /opt/mostro/lnd/admin.macaroon
+   install -m 644 /path/to/lnd/tls.cert /opt/mostro/lnd/tls.cert
+   install -m 600 -o 1000 -g 1000 /path/to/lnd/admin.macaroon /opt/mostro/lnd/admin.macaroon
    ```
 
+   The admin macaroon grants full control of your LND node — anyone who reads it can move the funds escrowed in Mostro's hold invoices — so it is installed owner-readable only, and `-o 1000 -g 1000` hands it to the container's user (as the `0700` directory from step 2 already was). Without that ownership, mode `0600` would leave mostrod unable to read the macaroon. (`-o`/`-g` require root; as a non-root user, drop them and run the steps as the account that owns the config dir.)
```
grunch: _🎯 Functional Correctness_ | _🟡 Minor_

**The non-root parenthetical still leads to `EACCES`.**

CodeRabbit raised this in round 2 (outside-diff, lines 164-165) and it is only half-addressed: step 2 now says the VPS steps run as root, but this sentence still tells a non-root operator to "drop `-o`/`-g` and run the steps as the account that owns the config dir". That account is rarely uid 1000, and both `docker run` examples in step 6 still run the container as the image default `1000:1000` — so the `0700` directory and `0600` macaroon it just created are unreadable to the daemon. The plain-image section a few paragraphs up already has the right remedy; mirror it here:

```diff
-(`-o`/`-g` require root; as a non-root user, drop them and run the steps as the account that owns the config dir.)
+(`-o`/`-g` require root; as a non-root user, drop them, run the steps as the account that owns the config dir, and add `--user $(id -u):$(id -g)` to the `docker run` commands in step 6 so the container runs as that account.)
```

grunch (changes requested):
## Strict re-review — head `dc24e11`

**Actionable comments posted: 5**

This round checks the follow-up commit against the 14 points of my previous review and looks for anything new it introduced. Local verification on `dc24e11`: `cargo test --bin mostrod config::permissions` (all green), plus a mutation probe on the new stale-temporary test (details inline).

<details>
<summary>✅ Previous findings resolved (13/14)</summary>

| # | Finding | Status |
|---|---|---|
| 1 | `.env` symlink-follow in `write_env_file` | ✅ `write_owner_only_atomic` (O_EXCL temp + `rename`), regression test added |
| 2 | `install -d -m 700 config` re-tightening on every `docker-up` | ✅ guarded with `[ -d config ]` in `docker-up`/`docker-relay-up` |
| 3 | `0700` propagated to every created ancestor | ✅ parents via `create_dir_all`, mode only on the leaf, test compares against a reference tree |
| 4 | No startup warning on `settings.toml`/`.env` | ✅ both checked in `main.rs`, in Lightning and Cashu mode |
| 5 | `MOSTRO_CONTAINER_USER` failing late | ✅ derived from the owner of `docker/config` in `docker-up` — but see the new Major below |
| 6 | `install` exposure window | ✅ analysed and documented in the Makefile |
| 7 | `curl -o` + `chmod` window | ✅ `(umask 077 && curl …)` |
| 8 | Truncated `settings.toml` on partial write | ✅ file removed on write failure |
| 9 | `start.sh` `-f` vs dangling symlink | ✅ `! -e && ! -L` |
| 10 | Predictable `/tmp` names, triplicated helpers | ✅ `test_support::temp_dir` with non-recursive `mkdir` (no `tempfile`, but O_EXCL semantics achieve the same) |
| 11 | Non-Unix branch uncovered | ✅ `owner_only_tests` now portable |
| 12 | `use std::fs` inconsistency | ✅ |
| 13 | README omitted "Lightning mode only" | ✅ |
| 14 | Title vs scope | ⏳ title unchanged — still worth a mention in the release notes |

</details>

<details>
<summary>🧹 Nitpick (not posted inline)</summary>

- `docs/STARTUP_AND_CONFIG.md` says the parents are made "with a recursive `DirBuilder`"; the code now uses `fs::create_dir_all`. Same semantics, but the citation is stale.

</details>

**Verdict:** REQUEST_CHANGES — the container-user derivation turns a root-run `make docker-up` into a root-run `mostrod`, which is a regression the pre-PR `USER mostrouser` did not have. Everything else is Minor.

Catrya (changes requested):
Request changes on one thing; the rest of the PR I'd take as is. Verified
locally: `install` does unlink the destination before creating it (hardlink
test), `-m` applies from creation, `install -d -m 700` chmods an existing dir
(so the `[ -d config ] ||` guards matter), symlink defences behave as
documented, 1228 tests pass, clippy and fmt clean.

**`sudo make docker-build` leaves a macaroon the container cannot read.**
`cp` used to keep the source mode, so a 0644 macaroon was readable by uid 1000
whoever owned it. `install -m 600` unlinks and recreates it as the invoking
user, so under `sudo` it lands `root:root 0600` in a `config/lnd` whose
ownership `docker-build` never touches — while `docker-up` derives the
container user from `config`, not from the credential:

1. `sudo make docker-build` → `config` is root → `docker-up` refuses on uid 0
   (that guard works).
2. Operator follows the doc: `sudo chown -R 1000:1000 config`.
3. Version bump: `sudo make docker-build` → `config` stays 1000, the macaroon
   is reinstalled `root:root 0600`.
4. `sudo make docker-up` → derives 1000:1000 → starts → mostrod can't read the
   macaroon, failing at LND connect with no hint about ownership.

Either derive `-o/-g` from `config`'s owner in `docker-build` when running as
root, or have `docker-up` verify the derived uid can read
`config/lnd/admin.macaroon` and fail there with a message.

Same file: `[ "${MOSTRO_CONTAINER_USER%%:*}" = 0 ]` only catches the numeric
form, so `MOSTRO_CONTAINER_USER=root:root` runs the container as root.

Non-blocking:
- `compose.yml` and three docs hardcode `1000:1000`, but the Dockerfile only
  says `useradd -m mostrouser`. Pin it with `-u 1000`.
- `mostro.db` is outside the startup check, though it sits in the same
  directory, is created under the umask, and holds trade history, disputes and
  hold-invoice preimages. Pre-existing 0755 installs are exactly the ones that
  need the warning.
- `write_owner_only_atomic` doesn't fsync the parent dir: the rename is atomic
  but not durable. Either fsync or narrow the doc comment.
- `README.md` says `docker-build` "sets `docker/config` to mode 0700"; it only
  does when it creates it.

Catrya (approved):
ACK

Matobi98:
@grunch friendly nudge on this one.

Your five round-3 comments are all addressed on the current head — `0868bcb` and `9d22652` landed after your review. What I ran against `9d22652`:

- **Points 1 and 4** (`Makefile`): `MOSTRO_CONTAINER_USER=0:0 make docker-up` refuses with "refusing to run the mostro container as root", exits 2 and starts nothing. A `config` at mode 775 is still 775 after `make docker-build`.
- **Points 2 and 3** (`permissions.rs`): `cargo test --bin mostrod config::permissions`, 23 passing. Your mutation still discriminates — with `TEMP_NAME_ATTEMPTS = 1` the stale-temporary test fails, so it does exercise the retry now.
- **Point 5** is the VPS path in `docker/README.md`: documentation, nothing to run.

Full suite on `9d22652`: 1228 passing.

Your review is what still has it blocked, and #887 is waiting on this one to land.

PR #891, opened 2026-08-17
fix: clear next-trade fields on sell child orders

### LINKED ISSUE #817: [LOW] Stale next-trade fields leak into sell child orders (buy path clears them, sell doesn't)
### Severity: LOW

Stale "next trade" fields leak into sell child orders. The buy path clears them; the sell path does not.

### Where
`src/app/release.rs` (range-order child creation) — buy vs sell child-order construction.

### Problem
When a range order spawns its next child, the buy branch resets the next-trade index/pubkey fields, but the sell branch carries them over, so a sell child can be created carrying stale next-trade metadata from the parent.

### Suggested fix
Clear the next-trade fields symmetrically on the sell path (mirror the buy path).

21Mill:
Closes #817

## What

`handle_sell_child_order` gains the two lines its buy-path sibling already has:

```rust
// Clear next trade fields for sell order
child_order.next_trade_index = None;
child_order.next_trade_pubkey = None;
```

## Why the child inherits them at all

`create_base_order` derives the child from the parent by cloning it:

```rust
// src/app/release.rs:1001
fn create_base_order(order: &Order) -> Result<Order, MostroError> {
    let mut new_order = order.clone();
    new_order.id = uuid::Uuid::new_v4();
    ...
```

It then resets the fields that must not carry over — `status`, `amount`, `hash`, `preimage`, `buyer_invoice`, `taken_at`, `invoice_held_at` — and blanks the counterparty side per order kind. The next-trade pair is not in that list, so it reaches both helpers still populated, and clearing it is left to them.

`handle_buy_child_order` does it (`src/app/release.rs:379-381`).
`handle_sell_child_order` returns without doing it.

The fields describe who takes over *after* the current trade. By the time a child is spawned that rotation has already been consumed — it is precisely how the child got its seller. Carrying the pair forward leaves the new order advertising a hand-off that already happened.

## That the path is reachable, not theoretical

Those fields are written in exactly one place:

```rust
// src/app/fiat_sent.rs:77-85
// If this is a range order, we need to update next trade fields
if order.is_range_order() {
    // Update next trade fields only when the buyer is the maker of a range order
    // These fields will be used to create the next child order in the range
    if let Some((pubkey, index)) = next_trade {
        order_updated.next_trade_pubkey = Some(pubkey);
        order_updated.next_trade_index = Some(index as i64);
    }
}
```

`fiat-sent` is a buyer action, and the guard is `is_range_order()` alone — the comment says the update is meant for buyer-maker ranges, but nothing checks it. So on a **sell** range order, where the maker is the seller, a buyer who includes a next-trade key still writes it to the parent row. Release then spawns the sell child, and the child was persisted carrying a key supplied by the counterparty of the finished trade.

**Persisted, not published.** `mostro-core-0.14.5/src/db/order.rs:53,102` binds `next_trade_pubkey` on insert, so the stale value did reach the child row. It never went over the wire, though: `SmallOrder` carries no next-trade fields and `as_new_order()` does not populate them, so the `NewOrder` message never contained the pair; and `order_to_tags` emits no next-trade tag, so neither did the 38383 event. The honest severity is hygiene — dead state that looks like live state — not disclosure.

**The parent row is out of scope here.** `fiat_sent.rs:77-85` wrote the pair to the **parent**, and nothing clears it there; this PR only stops it propagating to the child. Whether that guard should also check makership is the actual root cause and a separate question from the symmetry #817 reports, so I have deliberately not touched `fiat_sent.rs`. There is no issue tracking it yet — happy to open one, or to tighten the guard in a follow-up PR, if you want it addressed.

## Testing

New test `handle_sell_child_order_clears_next_trade_fields`, mirroring the assertions `handle_buy_child_order_uses_identity_key_in_normal_mode` already makes on the buy side (`src/app/release.rs:1660-1661`).

Confirmed it fails before the fix and passes after — in that order:

```
---- app::release::tests::handle_sell_child_order_clears_next_trade_fields stdout ----
assertion `left == right` failed
  left: Some("4efdb55da63145f8445006a0d431342c62bb33bf7370d20f9342828d9ba607b6")
 right: None
```

With the fix applied:

```
cargo test --bin mostrod        1200 passed; 0 failed; 2 ignored
cargo clippy --all-targets -- -D warnings   clean
cargo fmt --check                           clean
```

## Notes for the reviewer

- The comment wording copies the buy path (`// Clear next trade fields for
  sell order`) so the two branches read identically side by side. Say the word if you would rather it explained the reasoning instead.
- The existing end-to-end test `release_action_creates_child_order_for_range_order` is left as-is: it builds a parent with no next-trade fields set, so it never exercised this. The new unit test targets the helper directly, which is where the asymmetry lives.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed child orders retaining inherited next-trade rotation details when first created.
  * Ensured persisted child orders start with cleared next-trade fields for more reliable order processing.

### DIFF
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -416,6 +416,10 @@ fn handle_sell_child_order(
         }
     }
 
+    // Clear next trade fields for sell order
+    child_order.next_trade_index = None;
+    child_order.next_trade_pubkey = None;
+
     Ok((
         child_order.seller_pubkey.clone(),
         child_order.trade_index_seller,
@@ -1729,6 +1733,23 @@ mod tests {
         let mut child = fiat_sent_sell_order(seller, buyer);
         handle_sell_child_order(&mut child, Some((next_seller.to_string(), 3)), None).unwrap();
         assert_eq!(child.master_seller_pubkey, Some(next_seller.to_string()));
+    }
+
+    #[test]
+    fn handle_sell_child_order_clears_next_trade_fields() {
+        let seller = Keys::generate().public_key();
+        let buyer = Keys::generate().public_key();
+        let next_seller = Keys::generate().public_key();
+
+        // The child is cloned from the parent, next-trade fields included.
+        let mut child = fiat_sent_sell_order(seller, buyer);
+        child.next_trade_pubkey = Some(next_seller.to_string());
+        child.next_trade_index = Some(9);
+
+        handle_sell_child_order(&mut child, Some((next_seller.to_string(), 3)), None).unwrap();
+
+        assert_eq!(child.next_trade_pubkey, None);
+        assert_eq!(child.next_trade_index, None);
     }
 
     // ---------------------------------------------------------------
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -377,10 +377,6 @@ fn handle_buy_child_order(
         }
     }
 
-    // Clear next trade fields for buy order
-    child_order.next_trade_index = None;
-    child_order.next_trade_pubkey = None;
-
     Ok((
         child_order.buyer_pubkey.clone(),
         child_order.trade_index_buyer,
@@ -415,10 +411,6 @@ fn handle_sell_child_order(
             child_order.master_seller_pubkey = Some(next_trade_pubkey.to_string());
         }
     }
-
-    // Clear next trade fields for sell order
-    child_order.next_trade_index = None;
-    child_order.next_trade_pubkey = None;
 
     Ok((
         child_order.seller_pubkey.clone(),
@@ -1012,6 +1004,10 @@ fn create_base_order(order: &Order) -> Result<Order, MostroError> {
     new_order.taken_at = 0;
     new_order.invoice_held_at = 0;
     new_order.range_parent_id = Some(order.id);
+    // The next-trade rotation is consumed by the time a child is spawned —
+    // it is how the child got its counterparty. Never carry it forward.
+    new_order.next_trade_index = None;
+    new_order.next_trade_pubkey = None;
 
     match new_order.get_order_kind().map_err(MostroInternalErr)? {
         mostro_core::order::Kind::Sell => {
@@ -1522,6 +1518,8 @@ mod tests {
         order.max_amount = Some(100);
         order.min_amount = Some(10);
         order.fiat_amount = 40;
+        order.next_trade_pubkey = Some(Keys::generate().public_key().to_string());
+        order.next_trade_index = Some(9);
         let order = order.create(&pool).await.unwrap();
         let event = create_unwrapped_message_with_pubkey(seller);
         let msg = release_message(
@@ -1547,6 +1545,8 @@ mod tests {
         assert_eq!(child.fiat_amount, 0);
         assert_eq!(child.seller_pubkey, Some(next_seller.to_string()));
         assert_eq!(child.trade_index_seller, Some(2));
+        assert_eq!(child.next_trade_pubkey, None);
+        assert_eq!(child.next_trade_index, None);
         assert!(queued_actions_for(child.id)
             .await
             .contains(&Action::NewOrder));
@@ -1661,8 +1661,6 @@ mod tests {
         assert_eq!(trade_index, Some(5));
         assert_eq!(child.master_buyer_pubkey, Some(idkey));
         assert_eq!(child.creator_pubkey, next_buyer.to_string());
-        assert_eq!(child.next_trade_pubkey, None);
-        assert_eq!(child.next_trade_index, None);
     }
 
     #[test]
@@ -1735,23 +1733,6 @@ mod tests {
         assert_eq!(child.master_seller_pubkey, Some(next_seller.to_string()));
     }
 
-    #[test]
-    fn handle_sell_child_order_clears_next_trade_fields() {
-        let seller = Keys::generate().public_key();
-        let buyer = Keys::generate().public_key();
-        let next_seller = Keys::generate().public_key();
-
-        // The child is cloned from the parent, next-trade fields included.
-        let mut child = fiat_sent_sell_order(seller, buyer);
-        child.next_trade_pubkey = Some(next_seller.to_string());
-        child.next_trade_index = Some(9);
-
-        handle_sell_child_order(&mut child, Some((next_seller.to_string(), 3)), None).unwrap();
-
-        assert_eq!(child.next_trade_pubkey, None);
-        assert_eq!(child.next_trade_index, None);
-    }
-
     // ---------------------------------------------------------------
     // handle_child_order
     // ---------------------------------------------------------------
@@ -1768,9 +1749,7 @@ mod tests {
         order.creator_pubkey = buyer.to_string();
         order.next_trade_pubkey = Some(next_buyer.to_string());
         order.next_trade_index = Some(4);
-        let mut child = order.clone();
-        child.id = uuid::Uuid::new_v4();
-        child.status = Status::Pending.to_string();
+        let child = create_base_order(&order).unwrap();
 
         // Act
         let result = handle_child_order(child.clone(), &order, None, &pool, None).await;
@@ -1780,6 +1759,8 @@ mod tests {
         let db_child = Order::by_id(&pool, child.id).await.unwrap().unwrap();
         assert_eq!(db_child.buyer_pubkey, Some(next_buyer.to_string()));
         assert_eq!(db_child.trade_index_buyer, Some(4));
+        assert_eq!(db_child.next_trade_pubkey, None);
+        assert_eq!(db_child.next_trade_index, None);
         assert!(queued_actions_for(child.id)
             .await
             .contains(&Action::NewOrder));
@@ -1792,10 +1773,12 @@ mod tests {
         let seller = Keys::generate().public_key();
         let buyer = Keys::generate().public_key();
         let next_seller = Keys::generate().public_key();
-        let order = fiat_sent_sell_order(seller, buyer);
-        let mut child = order.clone();
-        child.id = uuid::Uuid::new_v4();
-        child.status = Status::Pending.to_string();
+        let mut order = fiat_sent_sell_order(seller, buyer);
+        // Seed the rotation on the parent. `create_base_order` is the only
+        // path that builds a child, so this is the shape production sees.
+        order.next_trade_pubkey = Some(Keys::generate().public_key().to_string());
+        order.next_trade_index = Some(9);
+        let child = create_base_order(&order).unwrap();
 
         // Act
         let result = handle_child_order(
@@ -1812,6 +1795,8 @@ mod tests {
         let db_child = Order::by_id(&pool, child.id).await.unwrap().unwrap();
         assert_eq!(db_child.seller_pubkey, Some(next_seller.to_string()));
         assert_eq!(db_child.trade_index_seller, Some(6));
+        assert_eq!(db_child.next_trade_pubkey, None);
+        assert_eq!(db_child.next_trade_index, None);
     }
 
     #[tokio::test]
@@ -1861,7 +1846,9 @@ mod tests {
     fn create_base_order_resets_trade_state_for_sell_orders() {
         let seller = Keys::generate().public_key();
         let buyer = Keys::generate().public_key();
-        let order = fiat_sent_sell_order(seller, buyer);
+        let mut order = fiat_sent_sell_order(seller, buyer);
+        order.next_trade_pubkey = Some(Keys::generate().public_key().to_string());
+        order.next_trade_index = Some(9);
 
         let base = create_base_order(&order).unwrap();
 
@@ -1877,6 +1864,9 @@ mod tests {
         assert_eq!(base.trade_index_buyer, None);
         // ... and keep the seller side.
         assert_eq!(base.seller_pubkey, order.seller_pubkey);
+        // ... and drop the consumed next-trade rotation.
+        assert_eq!(base.next_trade_pubkey, None);
+        assert_eq!(base.next_trade_index, None);
     }
 
     #[test]


src/app/release.rs
```diff
@@ -416,6 +416,10 @@ fn handle_sell_child_order(
         }
     }
 
+    // Clear next trade fields for sell order
+    child_order.next_trade_index = None;
+    child_order.next_trade_pubkey = None;
```
grunch: _♻️ Refactor_ | _🔵 Minor_ | non-blocking

**The clear has a better home: `create_base_order`.**

`create_base_order` (`src/app/release.rs:1250-1276`) is already the single place that lists every field that must not carry over from parent to child — `status`, `amount`, `hash`, `preimage`, `buyer_invoice`, `taken_at`, `invoice_held_at`, plus the per-kind counterparty blanking. The next-trade pair belongs in exactly that list. Putting it there collapses the four duplicated lines (buy path + sell path) into two and makes the next field-that-must-not-carry-over impossible to get asymmetric again.

I checked that the move is safe: `handle_buy_child_order` reads the next-trade pair off the **parent** `order` (`release.rs:411,418`), never off `child_order`, so clearing it on the base order cannot starve the buy path.

It would also close a gap this PR leaves open. `get_child_order` calls `create_base_order` and then `order_for_equal` / `order_for_greater`, which build **and sign the 38383 child event**, all before `handle_child_order` runs. The stale pair is live across that whole stretch today. It happens not to matter right now (`order_to_tags` does not emit next-trade tags), but that is a coincidence the current placement depends on, and the earlier placement does not.

```rust
// create_base_order, alongside the other resets
new_order.range_parent_id = Some(order.id);
// The next-trade rotation is consumed by the time a child is spawned —
// it is how the child got its counterparty. Never carry it forward.
new_order.next_trade_index = None;
new_order.next_trade_pubkey = None;
```

Then drop both `// Clear next trade fields for …` blocks, and extend the existing `create_base_order_resets_trade_state_for_sell_orders` test instead of adding a helper-level one.

Entirely your call — I understand the PR is deliberately scoped to the one-line symmetry fix #817 asks for.

src/app/release.rs
```diff
@@ -416,6 +416,10 @@ fn handle_sell_child_order(
         }
     }
 
+    // Clear next trade fields for sell order
+    child_order.next_trade_index = None;
```
grunch: _📝 Documentation_ | _🔵 Minor_ | non-blocking

**Two corrections to the PR description's severity framing** — the code is right, the write-up overstates the blast radius, and that wording will outlive the PR in `git log`.

**1. It is persisted, but not published.** The description says the child is *"persisted and published carrying a key supplied by the counterparty"*. Persisted, yes — `mostro-core-0.14.5/src/db/order.rs:53,102` binds `next_trade_pubkey` on insert. Published, no:

- `as_new_order()` → `SmallOrder` (`mostro-core-0.14.5/src/order.rs:315-333`) does not carry the next-trade pair, so the `NewOrder` message to the next taker never contained it.
- `order_to_tags` emits no next-trade tag, so the 38383 event never contained it either.

So the real impact is a stale DB column with no reader, not an over-the-wire leak. Still worth fixing — dead state that looks like live state is how the next bug gets written — but the honest severity is hygiene, not disclosure.

**2. The parent row keeps the same key.** If the concern is *"a key supplied by the counterparty of the finished trade is stored"*, this PR only clears the child. `fiat_sent.rs:77-85` wrote it to the **parent** row and nothing ever clears it there. That is the same `is_range_order()`-only guard you flagged, and it is the actual root cause — agreed it belongs in a follow-up, not here. Worth linking that follow-up from this description so the residual is tracked rather than assumed closed.

src/app/release.rs
```diff
@@ -1731,6 +1735,23 @@ mod tests {
...
+    #[test]
+    fn handle_sell_child_order_clears_next_trade_fields() {
+        let seller = Keys::generate().public_key();
+        let buyer = Keys::generate().public_key();
+        let next_seller = Keys::generate().public_key();
+
+        // The child is cloned from the parent, next-trade fields included.
+        let mut child = fiat_sent_sell_order(seller, buyer);
+        child.next_trade_pubkey = Some(next_seller.to_string());
+        child.next_trade_index = Some(9);
```
grunch: _🧪 Test Quality_ | _🔵 Minor_ | non-blocking

**Use a distinct key for the stale value.**

The stale pubkey and the incoming next-trade pubkey are both `next_seller`, so the assertion cannot distinguish *"the field was cleared"* from *"the field was overwritten with the new value and then cleared"*. The `assert_eq!(…, None)` makes the test sound either way, but the intent reads muddier than it needs to — and it would stay green under a hypothetical implementation that assigned rather than cleared.

Note the index already does this right (`Some(9)` stale vs `3` incoming); only the pubkey doesn't.

```rust
let stale_next_trade = Keys::generate().public_key();

// The child is cloned from the parent, next-trade fields included.
let mut child = fiat_sent_sell_order(seller, buyer);
child.next_trade_pubkey = Some(stale_next_trade.to_string());
child.next_trade_index = Some(9);
```

src/app/release.rs
```diff
@@ -1731,6 +1735,23 @@ mod tests {
...
+
+        handle_sell_child_order(&mut child, Some((next_seller.to_string(), 3)), None).unwrap();
+
+        assert_eq!(child.next_trade_pubkey, None);
+        assert_eq!(child.next_trade_index, None);
```
grunch: _🧪 Test Quality_ | _🔵 Minor_ | non-blocking

**Consider pinning the property at the DB layer too.**

The thing this PR actually fixes is what lands in the `orders` row. This helper-level test would keep passing if someone later re-populated the pair between `handle_sell_child_order` and `child_order.create(pool)` in `handle_child_order` — the exact shape of the regression, one layer up.

`handle_child_order_creates_sell_child_when_creator_is_seller` already reads the row back via `Order::by_id`, so it is two lines:

```rust
let db_child = Order::by_id(&pool, child.id).await.unwrap().unwrap();
assert_eq!(db_child.seller_pubkey, Some(next_seller.to_string()));
assert_eq!(db_child.trade_index_seller, Some(6));
assert_eq!(db_child.next_trade_pubkey, None);
assert_eq!(db_child.next_trade_index, None);
```

For that to be meaningful the arranged `child` there needs the fields pre-set (it is cloned from `order`, which leaves them `None` today), so seed them in the Arrange block first. Same reasoning applies to `release_action_creates_child_order_for_range_order`, which you correctly noted never exercised this — seeding `order.next_trade_pubkey` before the release would make it an end-to-end guard.

grunch (dismissed):
**Actionable comments posted: 4** (all non-blocking: 1 refactor, 1 documentation accuracy, 2 test quality)

The fix itself is correct, minimal and genuinely regression-tested. I re-derived the reachability claim independently and it holds: `create_base_order` (`src/app/release.rs:1250-1276`) clones the parent and resets `status`, `amount`, `hash`, `preimage`, `buyer_invoice`, `taken_at`, `invoice_held_at` and the counterparty side, but not the next-trade pair; `handle_buy_child_order` clears it, `handle_sell_child_order` did not; and `mostro-core-0.14.5/src/db/order.rs:53,102` binds `next_trade_pubkey` on insert, so the stale value really did reach the child row.

<details>
<summary>🔍 Local verification (merged onto current <code>main</code> <code>a7249cc</code>, not the PR base)</summary>

The branch is based on `09390da`; `src/app/release.rs` has moved +463/-104 on `main` since. It merges cleanly and is green on top of current `main`:

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | clean |
| `cargo test --bin mostrod` | `1223 passed; 0 failed; 2 ignored` |
| Regression test with the two production lines deleted | `FAILED` — `left: Some("1daac2ca…") right: None` |

The last row is the important one: the test is a true regression test, not a tautology.

</details>

<details>
<summary>⚠️ No CI signal on this PR (repo issue, not the author's)</summary>

`gh api .../check-runs` for `e6b0915` returns only three **skipped** jobs (`mutation-pr`, `mutation-baseline`, `mint-integration`). `.github/workflows/ci.yml` triggers on `push:` only — it has no `pull_request:` trigger — so fork PRs get zero fmt/clippy/test signal. Worth fixing separately; my local run above stands in for it here.

</details>

<details>
<summary>✅ Confirmed no behavioural regression from clearing the fields</summary>

`next_trade_pubkey` / `next_trade_index` are read in exactly one place — `handle_buy_child_order` (`release.rs:411,418`), and it reads them off the **parent** `order`, never off `child_order`. There is no take/authorization path, no DB query and no orderbook reconciler read that consults them. Clearing them on the sell child is inert for every consumer.

</details>

21Mill:
Thanks — took all four. Pushed as a separate commit rather than an amend so the delta is reviewable on its own.

**On hoisting the clear.** Agreed, and the gap you point out is the stronger half of the argument: `get_child_order` signs the child event before `handle_child_order` ever runs, so the stale pair was live across event construction. Depending on `order_to_tags` not emitting the tag is not a property I want the fix to rest on.

One thing the move surfaced that is worth recording, because it cuts against the DB-layer pinning suggestion as literally written. `handle_child_order` receives an already-built child and never calls `create_base_order`, so hoisting the clear moves the invariant out of it. The guarantee still holds — `get_child_order` is the only production caller — but it is now positional rather than enforced at the helper. Two existing helper-level tests failed as a result:

- `handle_sell_child_order_clears_next_trade_fields` (mine, added in this PR) — dropped; the helper no longer owns the behaviour.
- `handle_buy_child_order_uses_identity_key_in_normal_mode` (pre-existing) — lost its two next-trade assertions for the same reason.

So rather than seeding the hand-rolled clone in the sell-child test, I build the child through `create_base_order` there, which is the shape production actually produces and keeps your DB-layer assertion meaningful.

**On the two test-quality points.** Distinct stale key, and the DB-layer pin. I also seeded `release_action_creates_child_order_for_range_order` as you suggested.

CodeRabbit then caught a gap I had left on the buy side: dropping the helper-level assertion removed the buy path's only coverage of the clear, and `handle_child_order_creates_buy_child_when_creator_is_buyer` already seeded the parent's next-trade pair but never asserted on it — so it cloned past the reset and passed silently. Adding the two assertions against the unchanged clone reproduces it (`left: Some("d8cd8862…") right: None`); building the child through `create_base_order`, as the sell-child test now does, fixes it. Valid catch, taken.

The property is now held at four points:

| Test | Layer |
|---|---|
| `create_base_order_resets_trade_state_for_sell_orders` | unit |
| `handle_child_order_creates_buy_child_when_creator_is_buyer` | persisted row |
| `handle_child_order_creates_sell_child_when_creator_is_seller` | persisted row |
| `release_action_creates_child_order_for_range_order` | end to end |

All four fail with the two production lines removed — I checked in that order rather than assuming.

**On the description.** You are right and the original wording overstated it: persisted, not published. I verified both halves — `SmallOrder` has no next-trade fields and `as_new_order()` does not populate them, and `order_to_tags` emits no next-trade tag. Reworded to "hygiene, not disclosure", and added an explicit note that the parent row still carries the pair and that the `is_range_order()`-only guard in `fiat_sent.rs` is the root cause. No issue tracks that yet — say the word and I will open one.

Local verification on the new head: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --bin mostrod` 1199 passed / 0 failed / 2 ignored.

Note the push dismissed the two standing approvals, so I am re-requesting review.

src/app/release.rs
21Mill: Taken, with one adjustment: seeding the hand-rolled clone would not exercise the moved reset, since handle_child_order never calls create_base_order. The test now builds its child through create_base_order instead. Same for the buy-side test, after CodeRabbit flagged it. Details in the comment on the PR.

grunch (approved):
## Review: Approved

Re-reviewed head `1185d323` after the follow-up commit. All four of my prior comments are addressed, plus CodeRabbit's nitpick, and I verified each claim against the code rather than the reply.

**Comment coverage**

1. **Hoist the clear to `create_base_order`** — done (`src/app/release.rs:1007-1010`), both duplicated helper blocks dropped. I re-checked the safety of the move at the new head: `handle_buy_child_order` reads the rotation off the parent `order` (lines 360, 367) and `handle_sell_child_order` takes it from the message, so clearing the child cannot starve either path; `get_child_order:975` is the only production caller; and the clear now precedes `order_for_equal`/`order_for_greater`, closing the sign-a-stale-child-event gap the original placement left open.
2. **Description severity** — reworded to "persisted, not published" with the `SmallOrder`/`order_to_tags` evidence, and the parent-row residual is now called out explicitly. Covered.
3. **Distinct stale key** — done in the sell-child test and `create_base_order_resets_trade_state_for_sell_orders` (freshly generated key, `Some(9)` index).
4. **DB-layer pin** — done, and the adjustment (building the test child through `create_base_order` instead of seeding a hand-rolled clone) is the right call: after the hoist, seeding the clone would test a shape production never produces. Both `handle_child_order` tests now read the row back via `Order::by_id`, and the end-to-end release test seeds the parent and asserts on the persisted child.

CodeRabbit's follow-up nitpick (buy integration test bypassing the centralized reset) is also fixed — line 1752 builds through `create_base_order`.

**Independent verification** — relevant because `ci.yml` triggers on `push` only, so fmt/clippy/tests never ran in CI for this fork PR; only CodeRabbit reported on the head:

- `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --bin mostrod` 1199 passed / 0 failed / 2 ignored at `1185d323`.
- Mutation check: deleting the two clear lines from `create_base_order` makes **exactly the four claimed tests fail** (`create_base_order_resets_trade_state_for_sell_orders`, both `handle_child_order_creates_*`, `release_action_creates_child_order_for_range_order`) and nothing else. The "all four fail with the production lines removed" claim reproduces.

**Two non-blocking residues**

- The PR description's *What* / *Testing* / *Notes for the reviewer* sections still describe the first commit's shape: the two lines added to `handle_sell_child_order` (now gone), the test `handle_sell_child_order_clears_next_trade_fields` (since removed), and the note about the copied comment wording (moot). The severity framing was updated but these sections were not, and the body typically becomes the squash commit message — worth a quick refresh before merge.
- The `fiat_sent.rs` `is_range_order()`-only guard still has no tracking issue. Please open it as offered so the parent-row residual is tracked rather than assumed closed.

Neither blocks the merge.

PR #892, opened 2026-08-18
fix: verify the event signature before the spam gate

AndreaDiazCorreia:
The spam gate recorded an event id in its replay guard before the signature was checked. A nostr event id commits to `[0, pubkey, created_at, kind, tags, content]` and not to `sig`, so a copy of a victim's kind-14 event with only `sig` tampered keeps the victim's id: the copy got recorded, and the genuine event that followed was dropped as a replay — silently, since dropped events get no reply.

Not exploitable today: nostr-sdk verifies id and sig in `handle_event_msg` before delivering a subscription event, so such a copy never reaches the daemon. This removes the daemon's dependence on an external invariant it neither asserts nor tests.

- `accept_event` now runs PoW → kind → signature (hard, `return None`) → spam gate → decrypt. The hard check replaces the warn-only `event.verify()` that ran after the gate and rejected nothing.
- `accept_event` takes the gate explicitly instead of `is_v2` plus a global lookup: same argument count, and the ordering becomes testable without touching the process-wide `OnceLock`. `gate_for` holds the single "gate is v2-only" decision for both event loops and is resolved once per loop rather than once per event.
- Regression tests; `tampered_copy_does_not_censor_the_genuine_event` fails if the order is reverted.
- Ordering contract documented in `docs/TRANSPORT_V2_SPEC.md` and on `SpamGate::is_replay`.

**Behaviour change on the v1 path.** The signature check is transport-agnostic, and v1 is affected: an outer gift wrap whose id or signature does not verify previously logged a warning and continued into `unwrap_incoming`; it is now dropped. This is safe — no legitimately built gift wrap can hit that path, since the ephemeral key signs correctly — and it is strictly better than the warn-and-continue it replaces, because on v1 this is the daemon's *only* outer-event check: `unwrap_incoming` re-verifies the event on the v2 path alone (`unwrap_message_nip44`), while `nip59::unwrap_message` verifies the seal's signature and never the outer wrap. `v1_gift_wrap_with_invalid_signature_is_dropped` pins both halves: a tampered wrap is dropped, a well-formed one still goes through.

No wire-format change: validation order only.

`cargo test`: 1215 passed, 0 failed, 2 ignored.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of incoming events, including clearer rejection of invalid v1 gift wraps.
  - Prevented unauthenticated events from being recorded for replay tracking.
  - Ensured anti-spam protections select the appropriate gate for each event type.
- **Tests**
  - Added coverage for valid and invalid v1 gift-wrap handling.
  - Added verification for v2-specific anti-spam gate selection.
- **Documentation**
  - Clarified signature verification and replay-protection behavior for supported event formats.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -362,6 +362,13 @@ async fn accept_event(
         // must clear the stiffer `pow_first_contact` before
         // we decrypt. New orders/takes legitimately arrive
         // here — so does spam, hence the PoW toll.
+        //
+        // The lane is decided on a *verified* author on
+        // purpose: `is_known` keys off `event.pubkey`, and in
+        // v2 the trade keys of active orders are public by
+        // design. Deciding it before the signature check would
+        // let any flooder claim a known key and skip the
+        // first-contact toll entirely.
         if !gate.is_known(&event.pubkey.to_string()) && !event.check_pow(pow_first_contact) {
             tracing::info!(
                 "Dropping first-contact kind-14 event from unknown key {} below pow_first_contact ({} bits)",
@@ -616,6 +623,20 @@ mod tests {
         )
     }
 
+    // An AppContext backed by a fresh in-memory database with the migrations
+    // applied. Shared by every child test module through `use super::*`.
+    async fn create_migrated_ctx() -> AppContext {
+        let pool = std::sync::Arc::new(sqlx::SqlitePool::connect("sqlite::memory:").await.unwrap());
+        sqlx::migrate!("./migrations")
+            .run(pool.as_ref())
+            .await
+            .unwrap();
+        crate::app::context::test_utils::TestContextBuilder::new()
+            .with_pool(pool)
+            .with_settings(crate::app::context::test_utils::test_settings())
+            .build()
+    }
+
     // Helper function to create an UnwrappedMessage for testing. Identity and
     // sender (trade key) are distinct to mirror the canonical Mostro flow.
     fn create_test_unwrapped_message() -> UnwrappedMessage {
@@ -705,24 +726,9 @@ mod tests {
     /// event dropped as a replay.
     mod accept_event_ordering_tests {
         use super::*;
-        use crate::app::context::test_utils::{test_settings, TestContextBuilder};
         use crate::spam_gate::{SpamGate, REPLAY_WINDOW_SECS};
         use mostro_core::nip59::WrapOptions;
         use mostro_core::transport::wrap_message_nip44;
-        use sqlx::SqlitePool;
-        use std::sync::Arc;
-
-        async fn create_migrated_ctx() -> AppContext {
-            let pool = Arc::new(SqlitePool::connect("sqlite::memory:").await.unwrap());
-            sqlx::migrate!("./migrations")
-                .run(pool.as_ref())
-                .await
-                .unwrap();
-            TestContextBuilder::new()
-                .with_pool(pool)
-                .with_settings(test_settings())
-                .build()
-        }
 
         /// A protocol-v2 (kind 14) event addressed to `mostro`, in full-privacy
         /// mode (trade key doubles as identity, so no identity proof is needed).
@@ -756,13 +762,15 @@ mod tests {
             mostro: &Keys,
             gate: &SpamGate,
         ) -> Option<(Action, Message, UnwrappedMessage)> {
+            // Same constant the event loops derive `is_v2` from, so the test
+            // fails if it ever drifts from `Transport::Nip44Direct`'s kind.
             accept_event(
                 ctx,
                 event,
                 mostro,
                 0,
                 0,
-                NostrKind::PrivateDirectMessage,
+                NostrKind::from(crate::config::constants::DM_EVENT_KIND),
                 Some(gate),
             )
             .await
@@ -820,7 +828,7 @@ mod tests {
                 &mostro,
                 0,
                 0,
-                NostrKind::PrivateDirectMessage,
+                NostrKind::from(crate::config::constants::DM_EVENT_KIND),
                 None,
             )
             .await;
@@ -860,18 +868,6 @@ mod tests {
 
             let result = check_trade_index(&ctx, &event, &message).await;
             assert!(result.is_ok());
-        }
-
-        async fn create_migrated_ctx() -> AppContext {
-            let pool = Arc::new(SqlitePool::connect("sqlite::memory:").await.unwrap());
-            sqlx::migrate!("./migrations")
-                .run(pool.as_ref())
-                .await
-                .unwrap();
-            TestContextBuilder::new()
-                .with_pool(pool)
-                .with_settings(test_settings())
-                .build()
         }
 
         /// Insert a user row for `identity` with the given last_trade_index.
@@ -1179,21 +1175,6 @@ mod tests {
 
     mod dispatch_cashu_tests {
         use super::*;
-        use crate::app::context::test_utils::{test_settings, TestContextBuilder};
-        use sqlx::SqlitePool;
-        use std::sync::Arc;
-
-        async fn create_ctx() -> AppContext {
-            let pool = Arc::new(SqlitePool::connect("sqlite::memory:").await.unwrap());
-            sqlx::migrate!("./migrations")
-                .run(pool.as_ref())
-                .await
-                .unwrap();
-            TestContextBuilder::new()
-                .with_pool(pool)
-                .with_settings(test_settings())
-                .build()
-        }
 
         fn is_invalid_action(result: Result<()>) -> bool {
             matches!(
@@ -1214,7 +1195,7 @@ mod tests {
             let _ =
                 crate::config::MOSTRO_CONFIG.set(crate::app::context::test_utils::test_settings());
             let _ = crate::NOSTR_CLIENT.set(nostr_sdk::prelude::Client::default());
-            let ctx = create_ctx().await;
+            let ctx = create_migrated_ctx().await;
             let my_keys = create_test_keys();
             let event = create_test_unwrapped_message();
 
@@ -1249,7 +1230,7 @@ mod tests {
         /// returns `Ok` — proving it was NOT short-circuited to `InvalidAction`.
         #[tokio::test]
         async fn allows_restore_session_through_no_ln_router() {
-            let ctx = create_ctx().await;
+            let ctx = create_migrated_ctx().await;
             let my_keys = create_test_keys();
             let event = create_test_unwrapped_message();
             let msg = Message::new_restore(None);
--- a/src/app.rs
+++ b/src/app.rs
@@ -51,6 +51,7 @@ use crate::app::trade_pubkey::trade_pubkey_action;
 use crate::db::add_new_user;
 use crate::db::is_user_present;
 use crate::lightning::LndConnector;
+use crate::spam_gate::SpamGate;
 use crate::util::enqueue_cant_do_msg;
 use crate::Result;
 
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -321,41 +329,56 @@ async fn accept_event(
     if event.kind != accepted_kind {
         return None;
     }
+    // Authenticate the event BEFORE anything downstream records state keyed on
+    // it. A nostr event id commits to `[0, pubkey, created_at, kind, tags,
+    // content]` — **not** to `sig` — so a copy with a tampered signature keeps
+    // the victim's id. Verifying here (cheap Schnorr, pre-decrypt) is what
+    // makes the spam gate's dedup safe: only ids that are provably the
+    // author's own are ever recorded, so a forged copy cannot get the genuine
+    // event dropped as a replay. `unwrap_incoming` re-checks the signature for
```
grunch: _🛡️ Security Documentation_ | _🟠 Major_

**`unwrap_incoming` does *not* re-check the event signature for both transports — only for v2.**

`mostro-core-0.14.5/src/transport.rs::unwrap_message_nip44` verifies the outer event, so the claim holds on kind 14. The v1 branch does not: `nip59::unwrap_message` decrypts the wrap and then verifies **only the seal** (`seal.verify_signature()` — signature alone, not the id), never the outer gift-wrap event itself.

Two consequences, both worth correcting rather than leaving implied:

1. This comment describes the new check as belt-and-braces on top of an existing one. On v1 it is **the only** outer-event check the daemon performs. That is a strictly better position than the warn-and-continue it replaces, but a future reader trusting this comment could delete the check as redundant.
2. The PR body says "v1 gift wrap unaffected". It is affected: an outer gift wrap whose id or signature does not verify previously logged a warning and continued into `unwrap_incoming`; it is now dropped. No legitimately built gift wrap can hit that path (the ephemeral key signs correctly), so this is safe — but it is a behaviour change on the v1 path, and it is the one thing in this PR with no test behind it.

Suggested wording:

```rust
// `unwrap_incoming` re-checks the event signature on the v2 path
// (`unwrap_message_nip44`); the v1 gift wrap path verifies only the inner
// seal, so for v1 this is the daemon's *only* outer-event check.
```

And a line in the PR body noting the v1 drop path is new.

src/app.rs
```diff
@@ -463,14 +486,16 @@ pub async fn run(ctx: AppContext, ln_client: &mut LndConnector) -> Result<()> {
 
         while let Some(notification) = notifications.next().await {
             if let ClientNotification::Event { event, .. } = notification {
+                // The gate is v2-only; `None` fail-opens (v1, or not installed).
+                let gate = if is_v2 { SpamGate::global() } else { None };
```
grunch: _🧪 Test Coverage_ | _🟡 Minor_

**The `is_v2` → `Option<&SpamGate>` refactor moved the "gate is v2-only" policy out of the tested function into two untested, duplicated call sites** (here and line 543).

Making the ordering testable without the process-wide `OnceLock` is the right trade, but the decision that used to be covered by `accept_event`'s own body now lives in `run`/`run_cashu`, neither of which has a test. If one of the two lines is ever edited in isolation, the transports drift silently — which is precisely what the `accept_event` doc comment ("shared VERBATIM by `run` and `run_cashu` so the two event loops cannot drift") exists to prevent.

Also: `install_spam_gate()` runs before both loops in `main.rs`, so `SpamGate::global()` is loop-invariant here — this is a `OnceLock` load per inbound event for a value that cannot change.

Both are fixed by one small helper, hoisted above `loop`:

```rust
/// The gate applies to the v2 (kind-14) transport only; `None` fail-opens
/// (v1 gift wraps, or no gate installed).
fn gate_for(is_v2: bool) -> Option<&'static SpamGate> {
    if is_v2 { SpamGate::global() } else { None }
}
```

```rust
let gate = gate_for(is_v2);
loop {
    // ...
```

One definition, one place to test, and the lookup leaves the hot path.

src/app.rs
```diff
@@ -677,6 +718,124 @@ mod tests {
...
+            );
+            assert!(
+                accept(&ctx, &genuine, &mostro, &gate).await.is_none(),
+                "the replay guard still drops a re-sent identical event"
+            );
+        }
+
+        #[tokio::test]
+        async fn invalid_signature_is_dropped_without_a_gate() {
+            // v1 gift wraps and un-installed gates take the `None` path; the
```
grunch: _🧪 Test Coverage_ | _🟡 Minor_

**This test does not cover the v1 case its comment claims.** It builds a kind-14 event and passes `DM_EVENT_KIND` as `accepted_kind`; a real gift wrap would not even clear the kind check with those arguments. What is actually covered is "gate not installed", once.

Given that the hard `event.verify()` is the *only* outer-event check on the v1 path (see the comment on line 338), the gift-wrap case is the one that most deserves a test — it is the transport where nothing downstream will catch a bad outer signature:

```rust
#[tokio::test]
async fn v1_gift_wrap_with_invalid_signature_is_dropped() {
    let ctx = create_migrated_ctx().await;
    let mostro = create_test_keys();
    let genuine = /* wrap_message(...) -> Kind::GiftWrap */;
    let forged = with_tampered_signature(&genuine);

    assert!(accept_event(&ctx, &forged, &mostro, 0, 0, NostrKind::GiftWrap, None).await.is_none());
    assert!(accept_event(&ctx, &genuine, &mostro, 0, 0, NostrKind::GiftWrap, None).await.is_some());
}
```

The second assertion is the one that matters: it pins that the new hard drop does not reject well-formed gift wraps, which is the claim "v1 gift wrap unaffected" is resting on.

src/app.rs
```diff
@@ -321,41 +329,56 @@ async fn accept_event(
...
+    // Authenticate the event BEFORE anything downstream records state keyed on
+    // it. A nostr event id commits to `[0, pubkey, created_at, kind, tags,
+    // content]` — **not** to `sig` — so a copy with a tampered signature keeps
+    // the victim's id. Verifying here (cheap Schnorr, pre-decrypt) is what
+    // makes the spam gate's dedup safe: only ids that are provably the
+    // author's own are ever recorded, so a forged copy cannot get the genuine
+    // event dropped as a replay. `unwrap_incoming` re-checks the signature for
+    // both transports; this is the daemon's own gate, not a substitute.
+    if event.verify().is_err() {
+        tracing::warn!("Dropping event {} with an invalid signature", event.id);
```
grunch, line 346: _📊 Observability_ | _🔵 Trivial_ | _💤 Low value_

One `warn!` line per invalid-signature event, on a path that is remotely reachable and unmetered — base `pow` ships as `0`, and the first-contact toll is evaluated *after* this point. A flood of garbage-signature events writes one warn per event.

Defensible as-is: by the PR's own premise this branch is unreachable through a well-behaved `nostr-sdk` relay connection, so a hit means a compromised relay path or a bug, and that is worth shouting about. But if it fires it will fire at line rate. Worth a thought on `debug!` plus a counter, or leaving `warn!` with the amplification noted.

grunch (changes requested):
**Actionable comments posted: 4**

The core change is right and the ordering argument holds. `Event::verify()` in nostr `0.45.1` checks **both** `verify_id()` and `verify_signature()`, so moving it ahead of `SpamGate::is_replay` is exactly what makes the dedup safe, and `ReplayGuard`'s map can now only be grown by an attacker willing to sign every event. `cargo fmt --all -- --check` clean, `cargo test accept_event_ordering` → 3 passed locally on `0e2b3d9`.

What blocks: an inline comment about a security-relevant invariant is factually wrong for the v1 transport, and the PR body inherits the error ("v1 gift wrap unaffected"). v1 *is* affected, in a way that is neither documented nor tested. Details below.

<details>
<summary>📌 On the open Codex P2 ("keep first-contact PoW ahead of signature verification") — not repeating it, assessing it: <b>non-blocking, current order is the right one</b></summary>

The suggested order (`lane check` → `verify` → `is_replay`) preserves the replay-poisoning fix, so the disagreement is purely about who pays a Schnorr verify. But it makes the first-contact toll bypassable by exactly the population it targets: `is_known()` keys off the unauthenticated `event.pubkey`, and in v2 the trade keys of active orders are public by design, so a flooder sets `pubkey` to any live trade key, skips the toll, and *still* costs the daemon one verify before being dropped. Adaptive-attacker cost is 1 verify/event under both orders; only the naive flooder is cheaper under Codex's, and only when an operator actually sets `pow_first_contact > pow` (`settings.tpl.toml:66` ships `pow = 0`, and `effective_pow_first_contact()` falls back to `pow`, so the default config sees no change at all). The comment added at `src/app.rs:363-369` already makes this argument — it is correct and worth keeping.

</details>

<details>
<summary>✅ Verified against the tree</summary>

- `nostr-0.45.1/src/event/mod.rs:165` — `verify_with_ctx` = `verify_id()` + `verify_signature_with_ctx()`. The "id does not commit to `sig`" premise checks out (`EventId::compute` over `[pubkey, created_at, kind, tags, content]`).
- `nostr-relay-pool` `handle_relay_message` runs `verify_and_cache(&event)` *before* `database().check_id(&event.id)` — confirms the PR body's "not exploitable today", and is the same ordering this PR adopts locally.
- `mostro-core-0.14.5/src/transport.rs::unwrap_message_nip44` does call `event.verify()` — the v2 half of the comment at line 338 is accurate.
- `install_spam_gate()` (`src/main.rs:321`) runs before both `run` and `run_cashu`, so the per-event `SpamGate::global()` load is loop-invariant.

</details>

src/app.rs
```diff
@@ -677,6 +737,175 @@ mod tests {
...
+        }
+
+        /// The v2-only policy lives in one place now; both event loops read it
+        /// from here, so this is where it gets covered.
+        #[test]
+        fn gate_applies_to_v2_only() {
+            assert!(gate_for(false).is_none(), "v1 must fail open");
+            // `SpamGate::global()` is `None` unless `install_spam_gate` ran, so
+            // the v2 arm can only be pinned against the installed state.
+            assert_eq!(gate_for(true).is_some(), SpamGate::global().is_some());
```
grunch, line 905: _🧪 Test Coverage_ | _🟡 Minor_

**The v2 arm of this test is only as strong as test ordering, and has a small flakiness window.**

- If `install_global_then_second_install_is_rejected` (`spam_gate.rs:230`) has not run yet in this process, both sides are `None` and the assert passes without pinning anything — changing `gate_for(true)` to return `None` would go undetected.
- The two sides are read at different instants and tests run in parallel; an install landing between the two reads yields `false == true`. Tiny window, but real.

`install_global` returns `Err(AlreadyInstalled)` instead of panicking, so the test can force the state deterministically:

```rust
#[test]
fn gate_applies_to_v2_only() {
    assert!(gate_for(false).is_none(), "v1 must fail open");
    // Idempotent across the test binary: a second install is refused, not a panic.
    let _ = SpamGate::new(REPLAY_WINDOW_SECS).install_global();
    assert!(gate_for(true).is_some(), "v2 must resolve the installed gate");
}
```

src/app.rs
```diff
@@ -677,6 +737,175 @@ mod tests {
...
+            forged
+        }
+
+        async fn accept(
+            ctx: &AppContext,
+            event: &Event,
+            mostro: &Keys,
+            gate: &SpamGate,
+        ) -> Option<(Action, Message, UnwrappedMessage)> {
+            // Same constant the event loops derive `is_v2` from, so the test
```
grunch, line 784: _📝 Documentation_ | _🔵 Trivial_

`is_v2` no longer exists as a variable — the loops now derive `gate` via `gate_for(accepted_kind.as_u16() == DM_EVENT_KIND)`. Suggested:

```rust
// Same constant the event loops pass to `gate_for`, so the test
// fails if it ever drifts from `Transport::Nip44Direct`'s kind.
```

grunch (approved):
**Actionable comments posted: 2** (both 🟡 Minor / 🔵 Trivial — non-blocking)

Head `1e5ce45` resolves all four items from my previous review, and the PR body now states the v1 behaviour change explicitly. Re-verified against the tree:

- `nostr-0.45.1` `Event::verify_with_ctx` = `verify_id()` + `verify_signature_with_ctx()` — the "id does not commit to `sig`" premise and the hard drop are correct.
- `mostro-core-0.14.5` `nip59::unwrap_message` (`nip59.rs:237`) verifies only `seal.verify_signature()`; `transport::unwrap_message_nip44` (`transport.rs:257`) calls `event.verify()`. The corrected comment at `src/app.rs:340-344` and the new `TRANSPORT_V2_SPEC.md` paragraph now match the code.
- `install_spam_gate()` runs before `run` (`main.rs:289`) and before `run_cashu` (`main.rs:201`), so the `gate_for` loop-invariant claim holds on both boot paths.
- `v1_gift_wrap_with_invalid_signature_is_dropped` now covers the exact gap flagged earlier (tampered wrap dropped **and** well-formed wrap accepted through the full `unwrap_incoming` → stale → trade-index → inner-verify chain).
- `gate_for` is the single owner of the v2-only policy; both loops call it once, outside `loop {}`.

Local verification on `1e5ce45`: `cargo fmt --all -- --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test accept_event_ordering` → 5 passed; `cargo test spam_gate` → 8 passed. CI green (build, test, clippy, fmt, MSRV 1.94, markdownlint).

Approving. The two inline notes are polish on a test and a stale comment; neither needs a new round.

PR #893, opened 2026-08-18
fix: follow-ups to the payout dispatch queue (post-merge audit of #883)

Catrya:
Addresses the post-merge audit on #883 (all five findings) plus the two
non-blocking notes from the approval review.

| Audit finding | Commit |
|---|---|
| 1 (High) — a permit-queued task burns the buyer's retry budget for a payout that was never sent | 6cf06b4 |
| 2 — `classify_send_verdict` admits states its only caller can no longer produce | 7e7e351 |
| 3 — the dispatch path has no test coverage above the DB layer | da1ae64 |
| 4 (doc) — the permit doc overstates what RAII covers | 6cf06b4 |
| 5 (doc) — the claim token is wall-clock, not monotonic | 8482383 |
| approval note — semaphore doc implies it bounds connections | 6cf06b4 |
| approval note — the guard's degraded paths are silent | 77d4a54 |

**Finding 1** — while waiting for a send permit, the task now heartbeats its
claim: `touch_order_payout_claim` every `PAYOUT_QUEUE_HEARTBEAT` (derived as
2/3 of `MIN_GRACE_SECS`, promoted to a module constant so the relation is
structural), so a queued claim never ages past grace and reconciliation can
no longer re-arm a queued-but-never-sent payout. The acquire future is pinned
outside the `select!` loop so the task keeps its FIFO position in the
semaphore queue across heartbeats — the sketch in the review re-issued
`acquire()` per iteration, which would send the waiter to the back of the
queue every cadence. The post-permit touch stays as the final gate.

**Findings 2–3** — the drain now returns `StreamOutcome { Succeeded,
Failed(String), Ended }` (the impossible pairings are unrepresentable; the
six tests port 1:1), and the buyer-side dispatch decision got the same
extraction the bond side already had: a pure
`classify_dispatch(send_outcome, lookup)` returning
`StreamEnded / KeepMarker / ReArm`, with nine tests putting the
timeout-keeps-the-claim invariant under test instead of under a comment.

**Findings 4–5 and the approval notes** — docs narrowed to what the code
does (the semaphore bounds payment streams, not connections; the permit
spans the send and RPC-error reconcile; the token's 1s resolution and
clock-step caveats are named, with a monotonic per-order sequence recorded
as future work), and the duplicate guard's timeout/transport fall-throughs
now log instead of degrading silently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved payout processing reliability during payment-stream timeouts, failures, and unexpected termination.
  * Payout dispatches now maintain and revalidate claims while queued, dropping stale or unverifiable tasks.
  * Failed payment sends are classified more accurately to support appropriate retries.
  * Added logging when duplicate-payment checks time out or encounter transport errors.

* **Documentation**
  * Documented timestamp limitations that may affect payout claim reconciliation.

* **Reliability**
  * Standardized the grace period used during payout reconciliation.

### DIFF
--- a/src/app/bond/payout.rs
+++ b/src/app/bond/payout.rs
@@ -688,19 +688,18 @@ async fn pay_counterparty(
     // `Indeterminate` — `on_send_payment_failure` then keeps the invoice +
     // hash for reconciliation instead of re-prompting.
     let drain_fut = async move {
-        let mut succeeded = false;
-        let mut failure: Option<(PaymentFailureKind, String)> = None;
+        let mut outcome = StreamOutcome::Ended;
         while let Some(msg) = rx.recv().await {
             if let Ok(status) = PaymentStatus::try_from(msg.payment.status) {
                 match status {
                     PaymentStatus::Succeeded => {
-                        succeeded = true;
+                        outcome = StreamOutcome::Succeeded;
                         break;
                     }
                     PaymentStatus::Failed => {
-                        failure = Some((
-                            PaymentFailureKind::Terminal,
-                            format!("payment failed: reason {}", msg.payment.failure_reason),
+                        outcome = StreamOutcome::Failed(format!(
+                            "payment failed: reason {}",
+                            msg.payment.failure_reason
                         ));
                         break;
                     }
@@ -713,17 +712,32 @@ async fn pay_counterparty(
         // send future returns immediately instead of riding out the 75s
         // bound for a payment whose verdict we already hold.
         drop(rx);
-        (succeeded, failure)
+        outcome
     };
 
-    let (send_outcome, (succeeded, stream_failure)) = tokio::join!(send_fut, drain_fut);
+    let (send_outcome, stream) = tokio::join!(send_fut, drain_fut);
 
-    match classify_send_verdict(send_outcome, succeeded, stream_failure) {
+    match classify_send_verdict(send_outcome, stream) {
         SendVerdict::Settled => slash_after_success(pool, bond, counterparty_share).await,
         SendVerdict::Failure(kind, msg) => {
             on_send_payment_failure(pool, bond, max_retries, claim_window_seconds, kind, &msg).await
         }
     }
+}
+
+/// What the status drain actually observed — exactly the three states its
+/// loop can produce, so `classify_send_verdict`'s input domain has no
+/// unrepresentable-but-typeable values (no `(succeeded, Some(failure))`
+/// pairing, no indeterminate stream failure that nothing emits).
+#[derive(Debug, PartialEq)]
+enum StreamOutcome {
+    /// A `PaymentStatus::Succeeded` update was delivered.
+    Succeeded,
+    /// A `PaymentStatus::Failed` update was delivered, with its reason.
+    Failed(String),
+    /// The channel closed (send returned, errored, or was dropped by the
+    /// timeout) without a terminal update.
+    Ended,
 }
 
 /// Combined verdict of a bounded `send_payment` and its concurrent status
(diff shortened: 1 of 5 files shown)


grunch (approved):
**Review — approve.**

Verified locally on the head commit (`8482383`): `cargo fmt --check`, `cargo clippy --bin mostrod` and `cargo test --bin mostrod` all clean — 1219 passed, including the nine new `dispatch_*` tests and the six ported `classify_*` tests.

**Finding 1 — the heartbeat is correct, and the arithmetic is structural.** `PAYOUT_QUEUE_HEARTBEAT = MIN_GRACE_SECS * 2 / 3 = 20s` against `grace_secs = max(payment_retries_interval, MIN_GRACE_SECS) ≥ 30s`, so a queued claim is always re-stamped before `find_inflight_payouts`' cutoff can see it; promoting `MIN_GRACE_SECS` to a module constant makes the relation hold by construction rather than by comment. Pinning the `acquire` future outside the `select!` is the right call: tokio's semaphore is FIFO only while the `Acquire` future is alive, and re-issuing `acquire()` per iteration would indeed have rotated the waiter to the back of the queue every cadence.

I traced the token chain end to end: claim(T0) → heartbeat CAS(T0→T1) → … → post-permit touch(Tn-1→Tn), with the final rebind landing before the watcher closure is constructed, so the watcher's `clear_order_payout`/`fail_order_payout` and the outer `ReArm` path all carry Tn — no path is left on a stale token. Both heartbeat failure directions are the safe one (`Ok(None)`: a newer claim owns the order, drop; `Err`: keep the marker, the reconciler resolves). Race against the reconciler checked both ways: if the touch commits first, the reconciler's token-scoped release CAS no longer matches; if the release commits first, the touch sees a NULL hash and returns `None` → drop. No double-dispatch path.

**Findings 2–3 — faithful equivalences, now under test.** `StreamOutcome` collapses the old tuple exactly: `(true, _)` → `Succeeded`, `(false, Some((Terminal, msg)))` → `Failed(msg)`, `(false, None)` → `Ended`, and the `(succeeded, Some(_))` pairing the old type admitted is now unrepresentable. `classify_dispatch` preserves the old inline `Ok(Err)` arm (keep on InFlight/Succeeded/lookup-error; re-arm on Failed/Unknown/no-record/unusable-hash), and the timeout-keeps-the-claim invariant moved from a comment into `dispatch_timeout_keeps_the_marker`.

**Findings 4–5 and the approval notes — accurate.** The permit doc now matches reality (the watcher is a spawned sibling and finishes outside the RAII scope). The semaphore doc's "bounds payment streams, NOT connections" is true — `LndConnector::new()` runs before the claim, so queued tasks do hold idle connections, and the doc says so plainly. The duplicate guard's two degraded arms log and fall through to the same behavior as before; the match stays exhaustive.

**Two non-blocking nits, no re-review needed:**

1. `db.rs`, "Backward clock steps" bullet: a backward step *after* stamping only moves `now - grace` backwards, which makes a claim *less* reconcilable, not more. The grace-bypass case is a claim stamped *while* the clock sits behind (born old, immediately reconcilable once the clock corrects forward) — or a plain forward step. The exposure named is the right one; the sentence compresses the mechanism into the wrong direction.
2. `PAYOUT_QUEUE_HEARTBEAT`'s `* 2 / 3` degenerates to `0s` for `MIN_GRACE_SECS = 1` (integer division → `sleep(Duration::ZERO)` spin against the DB). Guarded today by the constant being 30 and the doc note next to it; a `.max(1)` would make the floor structural if the constant is ever expected to move.

All five audit findings and both approval notes are addressed faithfully, and the risky invariant — a queued-but-never-sent payout never ages past grace — is now enforced by construction and pinned by tests. LGTM.

PR #896, opened 2026-08-19
fix(yadio): canonicalise currency codes to uppercase

### LINKED ISSUE #859: [LOW] Yadio adapter does not canonicalise currency codes (§6.6) — the only adapter of six that skips it
## Summary

`src/price/providers/yadio.rs:50` forwards the currency code exactly as the API
ships it:

```rust
Some(v) if v.is_finite() && v > 0.0 => Some((code, Quote::PerBtc(v))),
```

It is the only one of the six adapters that does not canonicalise.

| Adapter | Line | Behaviour |
|---|---|---|
| `blockchain.rs` | 55 | `code.to_uppercase()` |
| `coingecko.rs` | 90 | `code.to_uppercase()` |
| `currency_api.rs` | 68 | `code.to_uppercase()` |
| `nostr.rs` | 146 | `code.to_uppercase()` |
| `eltoque.rs` | 142, 156 | hardcodes `"CUP"` / `"MLC"` |
| **`yadio.rs`** | **50** | **raw `code`** |

## Why this is a spec deviation, not a style preference

`docs/PRICE_PROVIDERS.md:196-202` states the requirement and its consequence:

> Two normalisation rules every adapter must follow (§6.6):
>
> - **Canonicalise currency codes to uppercase ISO-4217.** Providers
>   disagree on casing — `currency-api` ships lowercase (`"usd"`),
>   Yadio/Blockchain ship uppercase (`"USD"`). The aggregator keys on the
>   code, so an un-normalised adapter would silently fail to combine its
>   values with everyone else's.

And §6.6 itself, `docs/PRICE_PROVIDERS.md:372-375`:

> - **Code canonicalisation.** All currency codes are upper-cased before
>   they reach the combine step (adapters do this; the read path upper-cases
>   the requested code too). Without this, `currency-api`'s `"usd"` and
>   Yadio's `"USD"` would form two separate, single-source aggregates.

The responsibility is assigned explicitly to the adapter.

### The detail that makes the case

The spec names Yadio as a provider that "ships uppercase" — and it names
Blockchain in the same breath. Yet `blockchain.rs:53-55` canonicalises anyway,
and says why:

```rust
// Codes arrive uppercase already; canonicalise anyway so
// the adapter honours §6.6 even if the API drifts.
Some((code.to_uppercase(), Quote::PerBtc(v)))
```

Two adapters, the same assumption about their upstream, opposite choices. One
honoured the rule defensively; the other relied on the assumption holding.

## This pattern has already caused one bug

Not hypothetical. `restrict_nostr_to_fallback` (added during #841) compared raw
currency keys, while `aggregate_tick` upper-cases before grouping
(`aggregate.rs:141`). A raw `usd` from Yadio and a `USD` from Nostr therefore
failed to match in the dedup filter, both survived, and landed in the same
aggregate — the exact double-count that function exists to prevent. Fixed in
`2de2902` with a regression test verified by reverting the fix.

## Honest severity: there is no live bug today

Every downstream consumer normalises defensively — `store.rs`, `config.rs`,
`fiat.rs`, `aggregate.rs` and `manager.rs` all call `to_uppercase()` on the way
in (15 call sites between them). No price is wrong right now.

The argument is not "production is broken". It is: the contract says the
adapter does this, one adapter does not, and five downstream modules have to
remember to compensate — one already forgot, and it cost a review round.

Filing as maintainability / spec-compliance, low priority.

## Suggested fix

One line in `yadio.rs:50`:

```rust
Some((code.to_uppercase(), Quote::PerBtc(v)))
```

Plus an adapter test feeding a lowercase code through the fixture and asserting
it comes out uppercase, matching the existing tests in `blockchain.rs` /
`coingecko.rs`.

Removing some of the downstream defensive normalisation once all six adapters
comply is arguably a follow-up, but it is a separate decision and should not be
a precondition for the one-line fix — those calls are cheap and they are the
safety net.

---

Verified against `upstream/main` @ `7b1118a` on 2026-08-09. Searched open and
closed issues for duplicates (`yadio`, `uppercase currency`, normalisation
terms) — nothing overlapping.

Arowolokehinde:
## Summary

- Adds `.to_uppercase()` to the currency code at `yadio.rs:50`, making Yadio the sixth of six adapters to honour the §6.6 canonicalisation contract.
- Adds a `canonicalises_lowercase_currency_codes` test that feeds lowercase codes through `parse()` and asserts they emerge uppercase (and that the raw lowercase key is absent) — matching the pattern in `blockchain.rs` and `coingecko.rs`.

## Test plan

- [x] `cargo test price::providers::yadio` — all 5 tests pass
- [x] No behaviour change in production (Yadio ships uppercase codes today; this is a defensive correctness fix)

Closes #859

## Summary by CodeRabbit

* **Bug Fixes**
  * Currency codes from Yadio responses are now consistently normalized to uppercase.
  * Prevented lowercase currency keys from being retained during processing.

* **Tests**
  * Added coverage to verify currency code normalization.

### DIFF
--- a/src/price/providers/yadio.rs
+++ b/src/price/providers/yadio.rs
@@ -47,7 +47,9 @@ impl YadioProvider {
             .btc
             .into_iter()
             .filter_map(|(code, value)| match value {
-                Some(v) if v.is_finite() && v > 0.0 => Some((code.to_uppercase(), Quote::PerBtc(v))),
+                Some(v) if v.is_finite() && v > 0.0 => {
+                    Some((code.to_uppercase(), Quote::PerBtc(v)))
+                }
                 _ => None,
             })
             .collect())
@@ -117,7 +119,10 @@ mod tests {
         let quotes = YadioProvider::parse(body).unwrap();
         assert_eq!(quotes.get("USD"), Some(&Quote::PerBtc(75_000.0)));
         assert_eq!(quotes.get("EUR"), Some(&Quote::PerBtc(65_000.0)));
-        assert!(!quotes.contains_key("usd"), "raw lowercase key must not survive");
+        assert!(
+            !quotes.contains_key("usd"),
+            "raw lowercase key must not survive"
+        );
     }
 
     #[test]
--- a/src/price/providers/yadio.rs
+++ b/src/price/providers/yadio.rs
@@ -48,6 +48,8 @@ impl YadioProvider {
             .into_iter()
             .filter_map(|(code, value)| match value {
                 Some(v) if v.is_finite() && v > 0.0 => {
+                    // Yadio ships uppercase codes today; canonicalise anyway so
+                    // the adapter honours §6.6 even if the API drifts.
                     Some((code.to_uppercase(), Quote::PerBtc(v)))
                 }
                 _ => None,
@@ -115,8 +117,9 @@ mod tests {
 
     #[test]
     fn canonicalises_lowercase_currency_codes() {
-        let body = r#"{"BTC": {"usd": 75000.0, "eur": 65000.0}}"#;
+        let body = r#"{"BTC": {"usd": 75000.0, "Eur": 65000.0}}"#;
         let quotes = YadioProvider::parse(body).unwrap();
+        assert_eq!(quotes.len(), 2, "only the two canonical keys survive");
         assert_eq!(quotes.get("USD"), Some(&Quote::PerBtc(75_000.0)));
         assert_eq!(quotes.get("EUR"), Some(&Quote::PerBtc(65_000.0)));
         assert!(
--- a/src/price/providers/yadio.rs
+++ b/src/price/providers/yadio.rs
@@ -47,7 +47,7 @@ impl YadioProvider {
             .btc
             .into_iter()
             .filter_map(|(code, value)| match value {
-                Some(v) if v.is_finite() && v > 0.0 => Some((code, Quote::PerBtc(v))),
+                Some(v) if v.is_finite() && v > 0.0 => Some((code.to_uppercase(), Quote::PerBtc(v))),
                 _ => None,
             })
             .collect())
@@ -109,6 +109,15 @@ mod tests {
         let quotes = YadioProvider::parse(body).unwrap();
         assert_eq!(quotes.len(), 1, "only GBP is a usable rate");
         assert_eq!(quotes.get("GBP"), Some(&Quote::PerBtc(50_000.0)));
+    }
+
+    #[test]
+    fn canonicalises_lowercase_currency_codes() {
+        let body = r#"{"BTC": {"usd": 75000.0, "eur": 65000.0}}"#;
+        let quotes = YadioProvider::parse(body).unwrap();
+        assert_eq!(quotes.get("USD"), Some(&Quote::PerBtc(75_000.0)));
+        assert_eq!(quotes.get("EUR"), Some(&Quote::PerBtc(65_000.0)));
+        assert!(!quotes.contains_key("usd"), "raw lowercase key must not survive");
     }
 
     #[test]


src/price/providers/yadio.rs
```diff
@@ -47,7 +47,9 @@ impl YadioProvider {
             .btc
             .into_iter()
             .filter_map(|(code, value)| match value {
-                Some(v) if v.is_finite() && v > 0.0 => Some((code, Quote::PerBtc(v))),
+                Some(v) if v.is_finite() && v > 0.0 => {
+                    Some((code.to_uppercase(), Quote::PerBtc(v)))
```
grunch, line 53: _🧹 Consistency & Maintainability_ | _🔵 Nit_ | _⚡ Quick win_

**Add the "why" comment — every sibling that canonicalises carries one, and Yadio is the case that most needs it.**

The four adapters that already do this each explain themselves at the call site:

| File | Line | Comment |
|---|---|---|
| `blockchain.rs` | 53-54 | `// Codes arrive uppercase already; canonicalise anyway so the adapter honours §6.6 even if the API drifts.` |
| `coingecko.rs` | 89 | `// CoinGecko ships lowercase codes — canonicalise (§6.6).` |
| `currency_api.rs` | 67 | `// currency-api ships lowercase codes — canonicalise (§6.6).` |
| `yadio.rs` | 50-52 | *(none)* |

`blockchain.rs` is the exact analogue — upstream ships uppercase there too — and
#859 quoted its comment as the model for this fix precisely because, without it,
the next reader sees a `to_uppercase()` on data that is already uppercase and has
no way to tell it is deliberate rather than redundant. That is also what makes the
rustfmt-forced block form here read as arbitrary; the comment justifies the braces.

<details>
<summary>💡 Suggested change</summary>

```rust
                Some(v) if v.is_finite() && v > 0.0 => {
                    // Yadio ships uppercase codes today; canonicalise anyway so
                    // the adapter honours §6.6 even if the API drifts.
                    Some((code.to_uppercase(), Quote::PerBtc(v)))
                }
```

The module header would carry it well too — `coingecko.rs:5` and
`currency_api.rs:15` both state the casing contract in `//!` docs.
</details>

src/price/providers/yadio.rs
```diff
@@ -111,6 +113,18 @@ mod tests {
...
+    #[test]
+    fn canonicalises_lowercase_currency_codes() {
+        let body = r#"{"BTC": {"usd": 75000.0, "eur": 65000.0}}"#;
+        let quotes = YadioProvider::parse(body).unwrap();
+        assert_eq!(quotes.get("USD"), Some(&Quote::PerBtc(75_000.0)));
+        assert_eq!(quotes.get("EUR"), Some(&Quote::PerBtc(65_000.0)));
+        assert!(
+            !quotes.contains_key("usd"),
+            "raw lowercase key must not survive"
+        );
```
grunch, line 128: _🎯 Test Precision_ | _🔵 Nit_ | _⚡ Quick win_

**Pin the map size, and cover mixed case.**

CodeRabbit already asked for the missing `"eur"` assertion; the size assertion is
the stronger form of the same request and subsumes it — `drops_non_finite_and_non_positive`
(line 108) already uses that idiom in this very file:

```rust
        assert_eq!(quotes.len(), 2, "only the two canonical keys survive");
```

Separately, an all-lowercase input is the weakest possible probe for this fix.
Both `to_uppercase()` and a hypothetical wrong `to_ascii_uppercase()`/manual
first-char implementation pass it. A mixed-case code is what actually discriminates:

```rust
        let body = r#"{"BTC": {"usd": 75000.0, "Eur": 65000.0}}"#;
```

This is a nit — the fix is a stdlib call, so the risk of it being wrong is ~0. It
costs one character to make the test earn its name.

src/price/providers/yadio.rs
grunch, line 53: _🗄️ Data Integrity_ | _🟡 Minor_ | _📋 Context for the maintainer_

**Re: CodeRabbit's duplicate-key comment — here is what the change actually does to that case, which its analysis stops short of.**

Not a request for a change in this PR; posting because the finding above it is
half the picture and the maintainer should merge knowing the full delta.

CodeRabbit is right that `"USD"` + `"usd"` in one payload now collapse to one key
with the survivor picked by `HashMap` iteration order. What it does not say is
what happened **before** this PR, which I traced:

<details>
<summary>🔍 Pre-PR path for a case-variant duplicate</summary>

Both keys survived the adapter, then `aggregate_tick` (`src/price/aggregate.rs:141`)
uppercased them anyway and pushed **both** into the same bucket:

```rust
let currency = currency.to_uppercase();
match quote {
    Quote::PerBtc(v) => direct.entry(currency).or_default().push((*id, *v)),
```

Consequences, both real:

1. `combine()` sees two values from one provider → **Yadio is double-weighted** in
   the median.
2. `sources` counts *candidate values*, not distinct providers
   (`aggregate.rs:206-210`), so it reads `2` for a single-source currency —
   inflating the published figure and **suppressing the single-source `warn!`**
   in `observe_warnings` (`manager.rs:386`).

This is the same class of bug #859 documents as already having been hit once, in
`currencies_covered_by_non_nostr` (fixed in `2de2902`).
</details>

So the change trades a deterministic double-count for a nondeterministic pick. On
the axis that matters — one provider, one vote — it is a **net improvement**, and
the nondeterminism it introduces is unreachable today (Yadio ships uppercase) and
is shared verbatim by `blockchain.rs:55`, `coingecko.rs:90`, `currency_api.rs:68`
and `nostr.rs:146`.

Which is why I would not block on it: fixing collision precedence in this one
adapter would leave the other four inconsistent, and #859 explicitly scoped that
kind of cleanup out ("a separate decision and should not be a precondition for the
one-line fix"). If it is worth doing, it belongs in `aggregate_tick` — dedupe
`(provider, currency)` pairs once, where it fixes `sources` for every provider at
once — as a follow-up issue.

grunch (dismissed):
**Actionable comments posted: 3** (all non-blocking)

<details>
<summary>🧰 Verification performed</summary>

```
# on PR head 58c4115, clean worktree
cargo fmt --check            # clean
cargo clippy --all-targets --all-features   # clean, no warnings
cargo test price::           # 143 passed; 0 failed; 1 ignored
```

Also traced every downstream consumer of `ProviderQuotes` to confirm what this
change actually alters (see the inline comment on `parse`).
</details>

---

### Summary

The change is exactly what #859 prescribed: one line at the adapter boundary
plus an adapter-level regression test. It is correct, minimal, and in scope.
`fmt`/`clippy`/tests are green.

The findings below are consistency and test-precision items, not defects. I did
not repeat CodeRabbit's two comments — the third finding below adds the piece its
collision analysis was missing.

<details>
<summary>📝 Notes on the PR description (no code change needed)</summary>

- "matching the pattern in `blockchain.rs` and `coingecko.rs`" — `coingecko.rs`
  yes (`parses_captured_payload_and_uppercases_codes`, line 152). `blockchain.rs`
  has **no** canonicalisation test; it is the one adapter that canonicalises with
  a comment and no test — which is the closer analogue to Yadio, since both ship
  uppercase upstream.
- "sixth of six adapters" — `eltoque.rs` never calls `to_uppercase()`; it hardcodes
  `"CUP"`/`"MLC"` on the way out and does case-sensitive `tasas.get("USD")`
  lookups on the way in. #859's own table classifies it separately. Fine as a
  framing, but "five of five map-driven adapters" is the precise claim.
- Commit `58c4115` is a pure `cargo fmt` fixup of `dd06535`. Harmless — this repo
  squash-merges — just noting it.
</details>

src/price/providers/yadio.rs
Arowolokehinde: i just made the suggested changes 
Thank you

Arowolokehinde:
@grunch kindly review

src/price/providers/yadio.rs
```diff
@@ -47,7 +47,11 @@ impl YadioProvider {
             .btc
             .into_iter()
             .filter_map(|(code, value)| match value {
-                Some(v) if v.is_finite() && v > 0.0 => Some((code, Quote::PerBtc(v))),
+                Some(v) if v.is_finite() && v > 0.0 => {
+                    // Yadio ships uppercase codes today; canonicalise anyway so
+                    // the adapter honours §6.6 even if the API drifts.
+                    Some((code.to_uppercase(), Quote::PerBtc(v)))
```
grunch, line 53: _🧹 Consistency & Maintainability_ | _🔵 Nit_ | _🧷 Do not change in this PR_

**`to_ascii_uppercase()` is the precise primitive here — but changing it alone would be worse than leaving it.**

ISO-4217 codes are ASCII by definition, and `str::to_uppercase()` is the full
Unicode mapping: it can *expand* (`"ß"` → `"SS"`, `"ﬁ"` → `"FI"`) and it
allocates unconditionally even when the input is already uppercase — which,
per the comment two lines up, is every payload Yadio ships today. On a 120+
currency response that is 120+ throwaway `String`s per tick.

Neither cost is worth acting on here, because the four siblings
(`blockchain.rs:55`, `coingecko.rs:90`, `currency_api.rs:68`,
`nostr.rs:146`) all use `to_uppercase()`, and #859's whole argument is that
Yadio should look like its siblings. Diverging on this line would trade the
consistency the PR just bought for a micro-optimisation.

Flagging it so it is on the record: if it is ever revisited, all five change
together in one commit, not this one.

src/price/providers/yadio.rs
```diff
@@ -111,6 +115,19 @@ mod tests {
...
+    fn canonicalises_lowercase_currency_codes() {
+        let body = r#"{"BTC": {"usd": 75000.0, "Eur": 65000.0}}"#;
+        let quotes = YadioProvider::parse(body).unwrap();
+        assert_eq!(quotes.len(), 2, "only the two canonical keys survive");
+        assert_eq!(quotes.get("USD"), Some(&Quote::PerBtc(75_000.0)));
+        assert_eq!(quotes.get("EUR"), Some(&Quote::PerBtc(65_000.0)));
+        assert!(
+            !quotes.contains_key("usd"),
+            "raw lowercase key must not survive"
+        );
```
grunch, line 128: _🎯 Test Precision_ | _🔵 Nit_ | _⚡ Quick win_

**This assertion is now dead weight — the size assertion you added on line 122 already proves it.**

```rust
assert_eq!(quotes.len(), 2, "only the two canonical keys survive");
assert_eq!(quotes.get("USD"), Some(&Quote::PerBtc(75_000.0)));
assert_eq!(quotes.get("EUR"), Some(&Quote::PerBtc(65_000.0)));
```

Three assertions pin the map to exactly `{USD, EUR}`: length 2, with two
distinct known keys present. `"usd"` cannot also be in a 2-entry map that
already contains `USD` and `EUR`, so lines 125-128 can never fail while
122-124 pass — it is unreachable as a failure signal.

That is precisely why the size form was suggested over CodeRabbit's
"add the `eur` check": it subsumes both. Keeping the redundant assert is
harmless — dropping lines 125-128 leaves the same contract with less to read.

Your call: if you prefer to keep it as executable prose naming the bug the
test guards against, that is a legitimate reason to leave it, and I would not
hold the merge for it.

grunch (approved):
**Actionable comments posted: 2** (both non-blocking nits)

Re-review of head `d8e93db`, which addresses every finding from the dismissed review on `58c4115` and from CodeRabbit:

| Prior finding | Status |
|---|---|
| `grunch` — add the "why" comment at the call site (§6.6) | ✅ applied, verbatim-consistent with `blockchain.rs:53-54` |
| `grunch` — pin the map size, cover mixed case | ✅ `assert_eq!(quotes.len(), 2, …)` + `"Eur"` in the fixture body |
| CodeRabbit — assert the `"eur"` key is gone too | ✅ subsumed by the size assertion |
| CodeRabbit — duplicate-key (`"USD"` + `"usd"`) collision | ⚪️ informational; unchanged and unchangeable without a spec decision (see below) |

<details>
<summary>🧰 Verification performed</summary>

```
# on PR head d8e93db, clean worktree
cargo fmt --check                     # clean
cargo clippy --all-targets -D warnings # clean, no warnings
cargo test price::providers::yadio    # 5 passed; 0 failed
```

Consumer audit — no caller depends on the raw provider casing, so the change
is inert on today's payloads and correct on drifted ones:

| Consumer | Line | Behaviour |
|---|---|---|
| `price/config.rs` (`only`/`except` scoping) | 114-119 | upper-cases both sides |
| `price/fiat.rs` (allowlist) | 49 | upper-cases |
| `price/aggregate.rs` (combine key) | 141 | upper-cases |
| `price/store.rs` (write + read) | 75, 100, 113 | upper-cases |

</details>

<details>
<summary>📌 On the duplicate-key case — closing out the open thread</summary>

If Yadio ever shipped both `"USD"` and `"usd"` in one payload with different
values, `.collect()` into a `HashMap` is last-write-wins over the *source*
map's iteration order, which `RandomState` re-seeds per process — so the
surviving value would differ run to run. That is strictly better than the
pre-PR behaviour (two separate single-source aggregates, one of which never
matches any read because the read path upper-cases), but it is not
deterministic. Making it deterministic (min/max/reject) is a §6.6 spec
decision, not a code decision, and does not belong in this PR.

</details>

<details>
<summary>🔭 Out of scope — suggested follow-up issue</summary>

#859 and this PR frame the contract as *emitted* keys, and on that axis six
of six adapters now comply. The **ingest** side is still case-fragile in one
place: `eltoque.rs:137` and `:151` look their anchors up as
`tasas.get("USD")` / `tasas.get("MLC")` — exact-match reads against the
provider's own casing.

```rust
let cup_per_usd = match tasas.get("USD") {
    Some(Some(v)) if v.is_finite() && *v > 0.0 => *v,
    _ => return Ok(out),          // <- silent: Ok(empty), no error, no log
};
```

If El Toque drifted to `"usd"` — the exact drift this PR defends against —
the adapter would return an empty map, not an error. That lands in
`successes` but not `contributors` (`manager.rs:781-785`), so the tick stays
"green" while CUP/MLC quietly lose their only informal-market source. Worth
its own LOW issue; not a change for this PR.

</details>

---

Both inline comments below are optional. Approving; merge at will.

Issue #902, opened 2026-08-19
[MEDIUM] The #866 pre-trade CAS sweep missed the scheduler: full-row writes can resurrect or clobber pre-trade orders


AndreaDiazCorreia:
#866 converted every pre-trade order write to a compare-and-swap guarded by `status IN ('pending','waiting-taker-bond')`. The sweep covered `cancel.rs`, `take_buy.rs`, `take_sell.rs`, `add_invoice.rs` and `bond/flow.rs`, but not `src/scheduler.rs`.

## Where

`job_expire_pending_older_orders` (`src/scheduler.rs:1190`) runs unconditionally on a 60s tick. For each expired pre-trade order it calls `update_order_event` — a Nostr publish, a seconds-wide network window — and then persists with a full-row `order_updated.update(pool)` (`src/scheduler.rs:1257`, and `:1217` for the `WaitingMakerBond` branch). `Crud::update` is `UPDATE orders SET <all columns> WHERE id = ?`, with no status guard.

The timeout-cancel job at `src/scheduler.rs:708` has the same shape (`update_order_event` then full-row `update`).

## Problem

The scheduler runs in its own `tokio::spawn`, concurrently with the main event loop and with the bond invoice subscriber. Its read-publish-write sequence is exactly the pattern #866 identified as unsafe, and the race is bidirectional:

- The scheduler's stale snapshot lands last and reverts whatever committed during its Nostr publish.
- Or the scheduler commits `Expired` first and a concurrent stale full-row write resurrects the row to `pending`, leaving the DB disagreeing with the last published NIP-33 event.

With `[anti_abuse_bond]` enabled the bond-resume path (`resume_take_after_bond` -> `show_hold_invoice` -> `cas_complete_pretrade_take`) is a concurrent writer on the very same pre-trade window, so a committed take can be reverted with `hash`/`preimage` NULLed.

## Suggested fix

Route the scheduler's pre-trade persists through `cas_pretrade_order_status` (status + `event_id` only), matching the #866 pattern, and treat a miss as "someone else won — skip this order, retry next tick".

## Related

Same defect class as the `trade_pubkey.rs` rotation write (fixed in the PR for #811). Worth auditing the remaining `Crud::update` call sites on orders for other survivors of the sweep.

PR #903, opened 2026-08-19
fix: compare-and-swap the trade-pubkey rotation (#811)

### LINKED ISSUE #811: [MEDIUM] trade_pubkey confirmation sent before the rotation is persisted
### Severity: MEDIUM

`trade_pubkey` sends the confirmation to the user before the rotation is persisted, so a failed DB write leaves the user believing the key changed when it didn't.

### Where
`src/app/trade_pubkey.rs` (~lines 57–65) — the confirmation message is enqueued/sent ahead of (or without checking the result of) the DB update that stores the new trade pubkey.

### Problem
If the update fails or is discarded, the counterpart has been told the trade pubkey was updated while the stored value is stale. Subsequent messages signed against the "new" key won't match what the daemon has on record.

### Suggested fix
Persist the rotation first, check the result, and only send the confirmation after a successful commit.

AndreaDiazCorreia:
## What

`trade_pubkey_action` persisted the maker's trade-pubkey rotation with a full-row `Crud::update` — `UPDATE orders SET <all columns> WHERE id = ?`, with no status guard. This routes it through a compare-and-swap instead, and moves the confirmation message after the persist.

## Why

#866 converted every pre-trade order writer to a compare-and-swap guarded by `status IN ('pending','waiting-taker-bond')`, so that a writer working from a snapshot read earlier loses cleanly instead of overwriting whatever committed in the meantime. The sweep covered `cancel.rs`, `take_buy.rs`, `take_sell.rs`, `add_invoice.rs` and `bond/flow.rs`.

The trade-pubkey rotation was not part of it. #871 later rewrote this handler's ownership checks but left the persist as it was, so the invariant the sweep established did not actually hold across the whole pre-trade window.

The pre-trade window has concurrent writers outside the (sequential) event loop: the scheduler's expiry job, and — with anti-abuse bonds enabled — the bond invoice subscriber, whose resume path holds the order pre-trade across an LND `create_hold_invoice` round trip. A full-row write from this handler can therefore land on top of state another writer already committed, and it carries every column, not just the two the rotation means to move.

Separately, the confirmation was enqueued before the persist, so a write that never landed still told the maker their key had rotated. That is #811.

## Changes

- `src/db.rs`: `cas_rotate_maker_trade_pubkey` — writes only the maker-side trade pubkey (seller on a sell order, buyer on a buy order) and `creator_pubkey`, guarded by the existing `PRETRADE_STATUSES`.
- `src/db.rs`: `cas_complete_pretrade_take` narrowed to the **taker side only**. This is the widest behavioural change here — it affects `take_sell`, `take_buy`, the bond resume and `show_hold_invoice`, not just the rotation. A take never learns new maker-side values (every caller copies them from its own snapshot), so writing them back could only undo someone else's committed write: a rotation landing during the `create_hold_invoice` round trip used to be reverted here, and only halfway — `creator_pubkey` was left alone, and the resulting mismatch fails the `seller_pubkey == creator_pubkey` gate in `handle_child_order`.
- `src/app/trade_pubkey.rs`: persist through the CAS; reject with `NotAllowedByStatus` and log when the order left the pre-trade window; send the confirmation only after the write lands.

## Tests

- `cas_rotate_maker_trade_pubkey` wins from both pre-trade statuses and moves nothing but the two intended columns; wins on the buyer side of a buy order; misses on an order that already moved past the window, with the row left byte-for-byte intact.
- `cas_complete_pretrade_take` preserves a concurrent maker rotation, `creator_pubkey` included.
- `trade_pubkey_action` does not confirm a rejected rotation (#811) — both the pre-check rejection and the CAS miss — and a successful rotation leaves taker context and escrow material untouched.

Full suite green: 1241 passed. `cargo fmt --check` and `cargo clippy --all-features --all-targets -- -D warnings` clean.

## Scope

The pre-trade CAS invariant now holds for the Lightning paths. It does **not** yet hold for the Cashu take path: `show_cashu_escrow_request` (added by #830, after the #866 sweep) still persists with a full-row write built from a pre-claim snapshot and reverts a concurrent rotation. Filed as #919 rather than folded in here — routing it through `cas_complete_pretrade_take` is not a drop-in, since `claim_order_status` has already moved the row out of the guard's range; the issue spells out what the real fix involves.

## Related

- Closes #811
- Follows up on #866, which established the CAS invariant, and #871, which last touched this handler
- Follow-ups: #911 (the in-flight take still addresses the maker with the pre-rotation key — the row survives, the flow does not), #912 (`trade_index` is not rotated alongside the key), #919 (the Cashu take path, above)
- See also issue #902, the same class of missed writer in the scheduler

### DIFF
--- a/src/app/trade_pubkey.rs
+++ b/src/app/trade_pubkey.rs
@@ -74,6 +74,16 @@ pub async fn trade_pubkey_action(
         cas_rotate_maker_trade_pubkey(pool, order.id, kind, &event.sender.to_string()).await?;
     if !rotated {
         // The order left the pre-trade window while we were validating.
+        // Log it: this is the only trace the race leaves behind, and it is
+        // what tells us in production that a rotation and a take collided.
+        // `NotAllowedByStatus` matches the other CAS-miss sites
+        // (`take_sell.rs`, `show_hold_invoice`, `show_cashu_escrow_request`);
+        // the pre-check above keeps `InvalidOrderStatus`, the reason every
+        // pre-check in the repo reports for a status it can see up front.
+        tracing::info!(
+            "trade pubkey rotation: order {} left the pre-trade window before the CAS — refusing the rotation",
+            order.id
+        );
         return Err(MostroCantDo(CantDoReason::NotAllowedByStatus));
     }
 
@@ -367,6 +377,11 @@ mod tests {
 
     /// #811: the confirmation must never precede the persist. A rotation
     /// the handler rejects announces nothing to the caller.
+    ///
+    /// This is the *pre-check* rejection: the status is already out of the
+    /// pre-trade window when the handler reads the order, so it never
+    /// reaches the CAS. The CAS-miss branch is covered by
+    /// `trade_pubkey_action_does_not_confirm_a_cas_miss` below.
     #[tokio::test]
     async fn trade_pubkey_action_does_not_confirm_a_rejected_rotation() {
         let ctx = setup_ctx().await;
@@ -398,6 +413,60 @@ mod tests {
             confirmations, 0,
             "a rejected rotation must not be confirmed"
         );
+    }
+
+    /// The other half of #811, and the only handler-level coverage of the
+    /// `!rotated` branch: the order is still pre-trade when the handler
+    /// reads it, so validation passes and the CAS runs — but a concurrent
+    /// writer has moved the row out of the pre-trade window in between, so
+    /// the guarded `UPDATE` matches nothing.
+    ///
+    /// That interleaving has no injection point between `get_order` and the
+    /// CAS, so a `BEFORE UPDATE` trigger stands in for the racing writer:
+    /// `RAISE(IGNORE)` skips the row, which is exactly what the handler sees
+    /// from a status guard that no longer matches — `rows_affected() == 0`.
+    /// The handler must reject and stay silent; the row must be untouched.
+    #[tokio::test]
+    async fn trade_pubkey_action_does_not_confirm_a_cas_miss() {
+        let ctx = setup_ctx().await;
+        let maker = Keys::generate();
+        let sender = Keys::generate().public_key();
+
+        let order = maker_order(OrderKind::Sell, Status::Pending, &maker);
+        let order = order.create(ctx.pool()).await.unwrap();
+
+        sqlx::query(
+            "CREATE TRIGGER rotation_loses_the_cas BEFORE UPDATE ON orders \
+             WHEN NEW.creator_pubkey <> OLD.creator_pubkey \
+             BEGIN SELECT RAISE(IGNORE); END",
+        )
+        .execute(ctx.pool())
+        .await
+        .unwrap();
+
+        let event = trade_pubkey_event(sender, maker.public_key());
+        let result = trade_pubkey_action(&ctx, trade_pubkey_msg(order.id), &event).await;
+
+        assert!(
+            matches!(result, Err(MostroCantDo(CantDoReason::NotAllowedByStatus))),
+            "a CAS miss must be reported as NotAllowedByStatus: {result:?}"
+        );
+
+        // Nothing moved, and nothing was announced.
+        let after = order_by_id(ctx.pool(), order.id).await;
+        assert_eq!(after.creator_pubkey, maker.public_key().to_string());
+        assert_eq!(after.seller_pubkey, Some(maker.public_key().to_string()));
+        // The queue is process-global; filter by this test's unique key.
+        let confirmations = crate::config::MESSAGE_QUEUES
+            .queue_order_msg
+            .read()
+            .await
+            .iter()
+            .filter(|(m, pk)| {
+                *pk == sender && m.get_inner_message_kind().action == Action::TradePubkey
+            })
+            .count();
+        assert_eq!(confirmations, 0, "a CAS miss must not be confirmed");
     }
 
     /// The rotation is a targeted write: a pre-trade order carrying a
--- a/src/app/trade_pubkey.rs
+++ b/src/app/trade_pubkey.rs
@@ -1,7 +1,7 @@
 use crate::app::context::AppContext;
+use crate::db::cas_rotate_maker_trade_pubkey;
 use crate::util::{enqueue_order_msg, get_order};
 
-use mostro_core::db::Crud;
 use mostro_core::order::Kind as OrderKind;
 use mostro_core::prelude::*;
 
@@ -14,6 +14,9 @@ use mostro_core::prelude::*;
 /// maker-side trade pubkey and `creator_pubkey` (which tracks the maker's
 /// current trade key) are set to `event.sender`. Anything else is rejected
 /// with [`ServiceError::InvalidPubkey`] and the order is left untouched.
+///
+/// The rotation is persisted with a status-guarded compare-and-swap and the
+/// confirmation is sent only once that write lands.
 pub async fn trade_pubkey_action(
     ctx: &AppContext,
     msg: Message,
@@ -23,7 +26,7 @@ pub async fn trade_pubkey_action(
     // Get request id
     let request_id = msg.get_inner_message_kind().request_id;
     // Get order
-    let mut order = get_order(&msg, pool).await?;
+    let order = get_order(&msg, pool).await?;
 
     // Phase 1.5: accept both `Pending` and `WaitingTakerBond` as
     // pre-trade entry points. The trade pubkey is a maker-only piece of
@@ -56,15 +59,27 @@ pub async fn trade_pubkey_action(
     if maker_master_key != event.identity {
         return Err(MostroInternalErr(ServiceError::InvalidPubkey));
     }
-    match kind {
-        OrderKind::Sell => order.seller_pubkey = Some(event.sender.to_string()),
-        OrderKind::Buy => order.buyer_pubkey = Some(event.sender.to_string()),
+    // Persist through the pre-trade compare-and-swap (#866): it moves the
+    // maker-side trade pubkey and `creator_pubkey` — the maker is the order
+    // creator, and `creator_pubkey` must never move for anyone else — and
+    // nothing else, only while the order is still pre-trade.
+    //
+    // A full-row `Crud::update` here would write this handler's snapshot
+    // over whatever committed since the read at the top. The window is
+    // real: the post-bond resume keeps the order pre-trade across an LND
+    // `create_hold_invoice` round trip, so a rotation racing it could
+    // revert the committed take to `pending` with `hash`/`preimage`
+    // NULLed, orphaning a hold invoice the seller had already paid.
+    let rotated =
+        cas_rotate_maker_trade_pubkey(pool, order.id, kind, &event.sender.to_string()).await?;
+    if !rotated {
+        // The order left the pre-trade window while we were validating.
+        return Err(MostroCantDo(CantDoReason::NotAllowedByStatus));
     }
-    // The maker is the order creator: `creator_pubkey` tracks the maker's
-    // current trade key and must never move for anyone else.
-    order.creator_pubkey = event.sender.to_string();
 
-    // We a message to the seller
+    // Confirm only once the rotation is durable (#811): announcing it
+    // ahead of the write leaves the maker signing with a key the daemon
+    // never stored.
     enqueue_order_msg(
         request_id,
         Some(order.id),
@@ -75,11 +90,6 @@ pub async fn trade_pubkey_action(
     )
     .await;
 
-    order
-        .update(pool)
-        .await
-        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
-
     Ok(())
 }
 
@@ -87,6 +97,7 @@ pub async fn trade_pubkey_action(
 mod tests {
     use super::*;
     use crate::app::context::test_utils::{test_settings, TestContextBuilder};
+    use mostro_core::db::Crud;
     use nostr_sdk::prelude::{Keys, PublicKey, Timestamp};
     use sqlx::SqlitePool;
     use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


src/app/trade_pubkey.rs
```diff
@@ -56,15 +59,27 @@ pub async fn trade_pubkey_action(
...
+    // over whatever committed since the read at the top. The window is
+    // real: the post-bond resume keeps the order pre-trade across an LND
+    // `create_hold_invoice` round trip, so a rotation racing it could
+    // revert the committed take to `pending` with `hash`/`preimage`
+    // NULLed, orphaning a hold invoice the seller had already paid.
+    let rotated =
+        cas_rotate_maker_trade_pubkey(pool, order.id, kind, &event.sender.to_string()).await?;
+    if !rotated {
+        // The order left the pre-trade window while we were validating.
+        return Err(MostroCantDo(CantDoReason::NotAllowedByStatus));
```
grunch, line 87: Nit: same order state yields two different `CantDoReason`s depending on timing — `InvalidOrderStatus` from the pre-check at L39, `NotAllowedByStatus` here. The other CAS-miss sites in the repo (`take_sell.rs`, `show_hold_invoice`, `show_cashu_escrow_request`) use `NotAllowedByStatus`; consider unifying.

Also: every other CAS miss logs (`tracing::info!` with the order id) — this one returns silently. A trace here is what lets us see the race happening in production.

src/app/trade_pubkey.rs
```diff
@@ -353,4 +364,76 @@ mod tests {
         assert_eq!(after.buyer_pubkey, Some(maker.public_key().to_string()));
         assert_eq!(after.seller_pubkey, Some(taker.public_key().to_string()));
     }
+
+    /// #811: the confirmation must never precede the persist. A rotation
+    /// the handler rejects announces nothing to the caller.
+    #[tokio::test]
+    async fn trade_pubkey_action_does_not_confirm_a_rejected_rotation() {
```
grunch, line 386: This test uses `WaitingPayment`, so it dies at the pre-check on L39 (`InvalidOrderStatus`) and never reaches `cas_rotate_maker_trade_pubkey`. The new `if !rotated` branch (L75-78) has no handler-level coverage — only the DB function does.

I get that the race is hard to inject from the handler. At minimum, rename / adjust the doc-comment so it doesn't claim to cover the CAS-miss path; better, keep this one and add a second that reaches the CAS (e.g. a pre-trade order that a concurrent writer moves out of the window between `get_order` and the CAS, if the test context allows it).

src/db.rs
```diff
@@ -961,6 +977,48 @@ pub async fn cas_complete_pretrade_take(
...
+        OrderKind::Buy => "buyer_pubkey",
+    };
+    let query = format!(
+        "UPDATE orders SET {maker_column} = ?2, creator_pubkey = ?3 \
+         WHERE id = ?1 AND status IN ({PRETRADE_STATUSES})"
+    );
+    let result = sqlx::query(AssertSqlSafe(query.as_str()))
+        .bind(order_id)
+        .bind(maker_trade_pubkey)
+        .bind(maker_trade_pubkey)
```
grunch: Nit: same value bound twice. SQLite lets you reuse a numbered parameter — `SET {maker_column} = ?2, creator_pubkey = ?2` at L1008 and a single `.bind(maker_trade_pubkey)` here.

src/db.rs
```diff
@@ -961,6 +977,48 @@ pub async fn cas_complete_pretrade_take(
...
+/// seconds) can commit in between. A full-row write from the handler's
+/// stale snapshot would revert that take — status back to `pending`,
+/// `hash`/`preimage` NULLed, the seller's paid hold invoice orphaned.
+/// Guarding on the pre-trade statuses makes the rotation lose cleanly
+/// instead.
+///
+/// `maker_trade_pubkey` lands on the maker side fixed by the order kind:
+/// the seller of a sell order, the buyer of a buy order. Returns `false`
+/// when the CAS missed.
+pub async fn cas_rotate_maker_trade_pubkey(
```
grunch, line 862: Doc-comment reads as if the rotation/take race is fully resolved once the row survives. It is not: the take in flight still addresses its DMs with the pre-rotation key (#911). A one-line caveat here would stop the next reader from assuming otherwise.

grunch (changes requested):
Reviewed head `4905c76`. Verified locally: `cargo fmt --check`, clippy `-D warnings`, and the `cas_*` / `trade_pubkey*` tests (45) all green. CI green.

The fix itself is right — persist-then-confirm closes #811, the maker-side CAS carries `creator_pubkey` (avoiding the half-revert that breaks the `seller_pubkey == creator_pubkey` gate), and narrowing `cas_complete_pretrade_take` to the taker side is safe: I checked every caller (`take_buy.rs:154-197`, `take_sell.rs:240-242`, `show_hold_invoice` at `util.rs:1584-1585`, bond resume) and none carries *new* maker-side values — they all copy from the snapshot.

Requesting changes for one reason only:

### 1. Rebase — `main` has a new pre-trade full-row writer this PR does not cover
The branch is 4 commits behind `main` (merges clean, no conflicts). #830 (`0be0ab9`) added `show_cashu_escrow_request` (`src/util.rs:1679-1733` on main):

```rust
if !claim_order_status(pool, order.id, Status::Pending, Status::WaitingPayment).await? { ... }
order.seller_pubkey = Some(seller_pubkey.to_string());   // from the pre-claim snapshot
...
order_updated.update(pool)                                // full-row
```

A rotation landing between the order read and `claim_order_status` wins the CAS and confirms the maker — then this `update` reverts `seller_pubkey` **and** `creator_pubkey` to the snapshot. That is exactly the class of bug this PR's description says is now closed "across the whole pre-trade window". Not introduced here, but the claim no longer holds on `main`.

Ask: rebase, and route that write through `cas_complete_pretrade_take` (it fits now that the helper is taker-only; #830's own comment already warns about "narrowing the write"). If you'd rather keep it out of this PR, open an issue and link it from the description instead — but one of the two.

### 2. Follow-ups filed (not blocking)
- #911 — the in-flight take still sends `PayInvoice` / `WaitingSellerToPay` to the maker's **pre-rotation** key: the row survives, the flow does not. The doc-comment on `cas_rotate_maker_trade_pubkey` reads as if the race were fully resolved; worth a one-line caveat pointing at #911.
- #912 — `trade_index` is not rotated with the key (pre-existing).

Please add `Follow-ups: #911, #912` to the PR description.

### 3. Description
- The **Changes** section omits the third commit (`cas_complete_pretrade_take` taker-only), which is the widest behavioral change in the PR — it affects `take_sell`, `take_buy` and the bond resume, not just the rotation.
- "See also #902" does not resolve to a PR.

Inline nits below are non-blocking. Will approve once rebased and the Cashu writer is either fixed or tracked.

AndreaDiazCorreia:
@grunch thanks for the review — rebased and addressed. New head `f6e3f16`; `cargo fmt --check`, `cargo clippy --all-features --all-targets -- -D warnings` and the full suite (1241 passed) are green locally.

## 1. Rebase and the Cashu writer

Rebased onto `main` — clean, no conflicts.

The `show_cashu_escrow_request` write is filed as **#919** and linked from the description, rather than folded in here.

One thing worth recording from evaluating it: routing that write through `cas_complete_pretrade_take` is not a drop-in. `claim_order_status` has already advanced the row to `WaitingPayment` by the time the persist runs, so the helper's pre-trade guard would match zero rows. The real fix is to drop `claim_order_status` and let a single `cas_complete_pretrade_take` carry both the taker-side write and the status transition — the way `show_hold_invoice` and `take_sell` already do. That moves the guard to *after* `update_order_event`, so the loser publishes a NIP-33 event before finding out it lost and has to repair the relay view with `republish_winning_state_after_cas_miss`. It also has to keep the two ordering constraints the current comments at L1712-1727 protect (no clobbering a TA-1 lock CAS, neither `enqueue_order_msg` above the persist — a taker-only CAS satisfies the first by construction), and the three existing tests would need to follow. All of that is spelled out in the issue.

## 2. Description

- **Changes** now documents `cas_complete_pretrade_take` becoming taker-only, called out as the widest behavioural change in the PR (`take_sell`, `take_buy`, the bond resume, `show_hold_invoice`).
- New **Scope** section stating where the invariant actually holds: the Lightning paths, not yet the Cashu take path.
- `Follow-ups: #911, #912, #919`.
- #902 reworded — it is an issue, not a PR.

## 3. Inline nits

**`src/db.rs` — duplicate bind.** Done, `?2` reused for both columns with a single `.bind`.

**`src/db.rs` — doc-comment overclaims.** Done. The caveat now says the row surviving is not the flow surviving: the in-flight take keeps addressing the maker from its own snapshot, and points at #911.

**`src/app/trade_pubkey.rs` — no log on the CAS miss.** Done, `tracing::info!` with the order id.

**`src/app/trade_pubkey.rs` — the CAS-miss branch is untested.** Did both halves of your ask. The existing test's doc-comment now says explicitly that it covers the *pre-check* rejection, and `trade_pubkey_action_does_not_confirm_a_cas_miss` covers the `!rotated` branch at handler level. There is no injection point between `get_order` and the `UPDATE`, so a `BEFORE UPDATE ... RAISE(IGNORE)` trigger stands in for the racing writer: the row is skipped and the handler sees the same `rows_affected() == 0` a no-longer-matching status guard would produce. Same technique as `simulate_concurrent_claim` (`dev_fee.rs`) and `vanish_after_pending` (`bond/flow.rs`).

## 4. Not done — `CantDoReason` unification

Left as is, and I want to flag it rather than quietly skip it.

The pre-check / CAS-miss split is already the repo convention, not an inconsistency local to this handler:

- `take_sell.rs:84` and `take_buy.rs:46` — pre-check, `InvalidOrderStatus`
- `take_sell.rs:44` — CAS miss, `NotAllowedByStatus`

Exactly the same pair this handler has. Unifying only here would make `trade_pubkey` the outlier; unifying repo-wide changes a client-facing error code and reads like its own PR. I documented the reasoning in a comment at the CAS-miss site so the next reader does not re-litigate it.

Happy to change it either way if you'd rather — just say which direction.

src/db.rs
```diff
@@ -784,15 +788,27 @@ pub async fn cas_complete_pretrade_take(
...
+        OrderKind::Buy => "seller_pubkey = ?5, master_seller_pubkey = ?7, trade_index_seller = ?9",
+    };
     let query = format!(
-        "UPDATE orders SET status = ?2, event_id = ?3, \
-         buyer_pubkey = ?4, seller_pubkey = ?5, \
-         master_buyer_pubkey = ?6, master_seller_pubkey = ?7, \
-         trade_index_buyer = ?8, trade_index_seller = ?9, \
+        "UPDATE orders SET status = ?2, event_id = ?3, {taker_columns}, \
          fiat_amount = ?10, amount = ?11, fee = ?12, dev_fee = ?13, created_at = ?14, \
          buyer_invoice = ?15, preimage = ?16, hash = ?17, taken_at = ?18 \
```
grunch, line 808: **Minor — the maker side is not quite fully excluded.**

`buyer_invoice = ?15` is still written unconditionally. On a **buy** order the buyer *is* the maker, so this is a maker-side column written from the caller's snapshot — precisely the category the new doc comment at L765-767 says is "deliberately excluded", and precisely the shape the sibling `cas_promote_taker_context` already avoids: it puts `buyer_invoice = ?5` inside the `Sell` arm and binds a throwaway `None` for `Buy` (L711-726).

It is latent rather than live today — I checked the writers, and nothing can change `buyer_invoice` on a pre-trade buy order (`add_invoice_action` rejects anything that is not `WaitingBuyerInvoice` or `SettledHoldInvoice`, and `take_buy` passes `payment_request: None` so `show_hold_invoice` leaves the field at its snapshot value). So this is about the invariant, not a bug you are shipping.

But the invariant is the whole point of the change, and it is the one a future reader will rely on. Either move it into the `Sell` arm of `taker_columns` the way the sibling does, or add half a sentence to the doc comment saying why `buyer_invoice` stays unconditional. Silently having one exception to "the maker side is deliberately excluded" is the version that goes wrong later.

src/db.rs
```diff
@@ -2614,6 +2680,102 @@ mod tests {
...
+        // The take landed in full...
+        assert_eq!(after.status, "waiting-payment");
+        assert_eq!(after.buyer_pubkey, Some("bb".repeat(32)));
+        assert_eq!(after.master_buyer_pubkey, Some("bb".repeat(32)));
+        assert_eq!(after.trade_index_buyer, Some(3));
+        assert_eq!(after.hash, Some("ee".repeat(32)));
+        // ...and the rotation survived it, `creator_pubkey` included, so
+        // the `seller_pubkey == creator_pubkey` gate still holds.
+        assert_eq!(after.seller_pubkey, Some(rotated.clone()));
+        assert_eq!(after.creator_pubkey, rotated);
```
grunch, line 2728: **Minor — the narrowing is only one-third covered.**

The change dropped *three* columns per kind (`seller_pubkey`, `master_seller_pubkey`, `trade_index_seller` on the `Sell` branch), but this test only proves the trade pubkey and `creator_pubkey` survive. `stale` comes from `insert_pretrade_order`, which leaves `master_seller_pubkey` and `trade_index_seller` at `None`, and `committing` is a `stale.clone()` — so on those two columns the old wide `UPDATE` and the new narrow one write exactly the same thing, and no assertion can tell them apart.

I confirmed this is not theoretical. Restoring two-thirds of the old behaviour on the `Sell` branch:

```rust
OrderKind::Sell => "buyer_pubkey = ?4, master_buyer_pubkey = ?6, trade_index_buyer = ?8, \
                    master_seller_pubkey = ?7, trade_index_seller = ?9",
```

leaves the **entire suite green — 1241 passed, 0 failed**. A regression that reintroduces the maker-side clobber for the master key and the trade index would ship unnoticed.

Cheapest fix: give the fixture distinct maker-side values and assert they survive. On `stale`, set `master_seller_pubkey = Some("aa".repeat(32))` and `trade_index_seller = Some(7)`; on `committing`, set them to something else (`Some("cc".repeat(32))` / `Some(99)`) to simulate the stale snapshot; then assert the row still holds the original pair after the take. Same treatment on the buy-side mirror at L2776 for `master_buyer_pubkey` / `trade_index_buyer`.

grunch (approved):
Re-reviewed at head `f6e3f16`. Verified locally: `cargo fmt --check` clean, `cargo clippy --all-features --all-targets -- -D warnings` clean, full suite **1241 passed / 0 failed**. All four points from my previous pass are addressed correctly:

1. `NotAllowedByStatus` on the CAS miss, matching the other CAS-miss sites — and it now logs. ✅
2. The CAS-miss branch has real handler-level coverage (`trade_pubkey_action_does_not_confirm_a_cas_miss`); the `RAISE(IGNORE)` trigger is a legitimate stand-in — a `BEFORE UPDATE` trigger that raises `IGNORE` abandons the row without rolling back, so `rows_affected()` really is `0`, which is exactly what a no-longer-matching status guard produces. The doc comment says as much rather than pretending it injected a real race. ✅
3. `?2` bound once. ✅
4. The #911 caveat is on `cas_rotate_maker_trade_pubkey`. ✅

I also re-derived the two claims the PR rests on rather than taking them from the description:

- **The taker-side narrowing is safe.** All four callers of `cas_complete_pretrade_take` reach it through a path that derives both pubkeys from its own DB snapshot — `take_sell` (`order.get_seller_pubkey()`), `take_buy` (`order.get_buyer_pubkey()`), `add_invoice` (both from `order`), and `resume_take_after_bond` (both from the promoted row). None of them ever *learns* a maker-side value, so dropping those columns cannot lose information. Confirmed too that `waiting-buyer-invoice` is only ever reached from sell-order takes (`set_waiting_invoice_status` is called from `take_sell` and the bond resume's `Sell` leg only), so the `allow_waiting_buyer_invoice` caller is always a sell order and its taker side is the buyer.
- **The bind slots are legal.** `?18` is referenced on both branches, so SQLite's `bind_parameter_count()` is 18 either way and the unreferenced `?5/?7/?9` (or `?4/?6/?8`) slots cost nothing. Matches the `cas_promote_taker_context` precedent.

One thing worth stating explicitly so nobody "fixes" it later: this handler correctly does **not** call `republish_winning_state_after_cas_miss`, unlike `take_sell.rs`, `show_hold_invoice`, `cancel.rs` and the bond resume. Those all publish a kind-38383 event *before* their CAS, so a miss leaves a stale event on relays that needs repairing. `trade_pubkey_action` publishes nothing, so there is no divergence to repair. The asymmetry is right.

Approving. Two minor items and a note below — none blocking.

---

**Out of band (no code needed here):** `TradePubkey` is not covered by `check_trade_index` (`src/app.rs:137` matches only `NewOrder | TakeBuy | TakeSell`), and `spam_gate`'s replay dedup is explicitly short-window. So a replayed `TradePubkey` gift wrap can still rotate the maker key *backwards* while the order is pre-trade — now that the write itself is guarded and confirmed, that is the one remaining unguarded axis on this handler. It overlaps #912 but is not the same thing (#912 is about rotating the index *forward* alongside the key). Worth a line on #912 or its own issue.

Also, while you are here: `dispatch_cashu`'s doc comment (`src/app.rs:590-593`) lists `TradePubkey` under "read-only / session; never touch escrow, LND, or order lifecycle". It writes the order row — it always did, and after this PR it does so through a CAS. That is a one-line comment fix whenever the file is next touched; not this PR's job.

PR #908, opened 2026-08-20
fix: run failure bookkeeping on every pre-claim payout failure

Catrya:
Follow-up to the buyer-payout work in #881, #883 and #893, covering the one
residual case surfaced:
Only three of `do_payment`'s pre-claim exits ran `check_failure_retries_or_log`.
The rest , no buyer invoice, fee eating the whole amount, an unreadable buyer
pubkey, `LndConnector::new` failing on its cert/macaroon read, an undecodable
payout invoice, a DB error on the claim write, returned `Err` with no
bookkeeping. Four of them exit through `?`, which is why they were easy to miss.

An order that takes one of those exits keeps `failed_payment = false` and has no
payout marker, so `find_failed_payment` (needs the flag) and
`find_inflight_payouts` (needs the marker) both skip it. The seller is already
settled by then, so the trade sits in settled-hold-invoice forever: the buyer is
never paid, never notified, never re-prompted for an invoice. Nothing is logged
either — all three callers discard `do_payment`'s error. Recovery needs a manual
DB write.

An unreachable LND is *not* a trigger: `connect()` is lazy, so it fails after the
claim, where reconciliation already owns the outcome.

### The fix

Do the bookkeeping once, in `do_payment`, over the result of `dispatch_payout`,
instead of at each exit. Every error path is covered, including the `?` ones and
any added later — which is the point: the bug exists because the per-site
convention was forgettable, and it was forgotten twice.

 `dispatch_payout` is only reached once the
order has been validated, so each of its error exits means the buyer could not be
paid. Documented on the function, together with the two paths that deliberately
return `Ok` (a lost claim CAS, and the post-claim work that runs in the spawned
task and does its own bookkeeping).

Three commits: a no-behavior-change split, the fix itself (~15 lines), and tests.

### Verification

- Full suite: 1225 passed, 0 failed. `clippy --all-targets` clean.
- The four `do_payment` tests now assert the order is left recoverable:
  `failed_payment` set, `payment_attempts` at 1, buyer notified, no payout marker
  left for a payment that was never sent.
- Reverting the wrapper makes all four fail (46 pass), so they pin the fix rather
  than the current behavior.
- The LNURL test pins `payment_attempts == 1`: moving the bookkeeping up cannot
  start counting twice and silently halve the buyer's retry budget.

Verified on regtest (Polar, LND alice): with the daemon running, the cert file
it reads was moved aside, its long-lived client keeps working, only
`do_payment`'s fresh connector fails. On release, the hold invoice settled and
the order was marked 70ms later (`has 1 failed payments retries`), with
`failed_payment = 1`, `payment_attempts = 1` and no payout marker; the buyer got
`payment-failed`. Restoring the cert, the retry job picked the order up on its
next tick and paid it unattended, `success`, `purchase-completed` and `rate`
within 40s, no intervention. On the old code that order would have stayed in
settled-hold-invoice with nothing logged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved payout failure handling so retry status and payment attempts are recorded even when buyer details or payout data are invalid.
  * Reduced the risk of settled orders losing failure tracking when payout processing exits early.

* **Tests**
  * Added coverage for unusable buyer details during payout processing, confirming failure tracking is preserved and no unnecessary notification is sent.

### DIFF
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -1475,6 +1475,38 @@ mod tests {
             .collect()
     }
 
+    /// Assert a pre-claim payout failure left the order recoverable: the
+    /// failure bookkeeping ran, so the row carries the retry flag and its
+    /// first attempt, the buyer was told, and no payout marker was left
+    /// behind for a payment that was never sent. Without the flag the order
+    /// is invisible to `find_failed_payment`, and without a marker it is
+    /// invisible to `find_inflight_payouts` — it would sit in
+    /// settled-hold-invoice forever.
+    async fn assert_payout_failure_recorded(pool: &SqlitePool, order_id: uuid::Uuid) {
+        let db_order = Order::by_id(pool, order_id).await.unwrap().unwrap();
+        assert!(
+            db_order.failed_payment,
+            "a pre-claim payout failure must set failed_payment"
+        );
+        assert_eq!(
+            db_order.payment_attempts, 1,
+            "the failure must count exactly once"
+        );
+        let marker: Option<String> =
+            sqlx::query_scalar("SELECT payout_payment_hash FROM orders WHERE id = ?")
+                .bind(order_id)
+                .fetch_one(pool)
+                .await
+                .unwrap();
+        assert!(
+            marker.is_none(),
+            "no payout marker may be left for a payment that was never sent"
+        );
+        assert!(queued_actions_for(order_id)
+            .await
+            .contains(&Action::PaymentFailed));
+    }
+
     struct StubEscrow;
 
     #[async_trait]
@@ -2302,7 +2334,10 @@ mod tests {
         let ctx = build_ctx(&pool);
         let seller = Keys::generate().public_key();
         let buyer = Keys::generate().public_key();
-        let order = fiat_sent_sell_order(seller, buyer);
+        let mut order = fiat_sent_sell_order(seller, buyer);
+        order.status = Status::SettledHoldInvoice.to_string();
+        let order = order.create(&pool).await.unwrap();
+        let order_id = order.id;
 
         let result = do_payment(&ctx, order, None).await;
 
@@ -2310,6 +2345,7 @@ mod tests {
             result,
             Err(MostroInternalErr(ServiceError::InvoiceInvalidError))
         ));
+        assert_payout_failure_recorded(&pool, order_id).await;
     }
 
     #[tokio::test]
@@ -2322,6 +2358,9 @@ mod tests {
         order.buyer_invoice = Some("lnbc1notchecked".to_string());
         order.amount = 100;
         order.fee = 100;
+        order.status = Status::SettledHoldInvoice.to_string();
+        let order = order.create(&pool).await.unwrap();
+        let order_id = order.id;
 
         let result = do_payment(&ctx, order, None).await;
 
@@ -2329,8 +2368,13 @@ mod tests {
             result,
             Err(MostroInternalErr(ServiceError::InvoiceInvalidError))
         ));
+        assert_payout_failure_recorded(&pool, order_id).await;
     }
 
+    /// Regression test for the stranded-order class: this exit is reached
+    /// through `?`, and before the bookkeeping moved into `do_payment` it
+    /// left a settled order with neither the retry flag nor a payout marker
+    /// — invisible to both recovery jobs, with the buyer never notified.
     #[tokio::test]
     async fn do_payment_fails_fast_when_lnd_is_unreachable() {
         // Arrange: with the global config set to test defaults, the LND cert
@@ -2343,12 +2387,16 @@ mod tests {
         let buyer = Keys::generate().public_key();
         let mut order = fiat_sent_sell_order(seller, buyer);
         order.buyer_invoice = Some("lnbc1notchecked".to_string());
+        order.status = Status::SettledHoldInvoice.to_string();
+        let order = order.create(&pool).await.unwrap();
+        let order_id = order.id;
 
         // Act
         let result = do_payment(&ctx, order, None).await;
 
         // Assert
         assert!(result.is_err());
+        assert_payout_failure_recorded(&pool, order_id).await;
     }
 
     /// An expired regtest invoice: rejected by `validate_payout_invoice` on
@@ -2420,6 +2468,9 @@ mod tests {
 
         let mut order = fiat_sent_sell_order(seller, buyer);
         order.buyer_invoice = Some(lnurl);
+        order.status = Status::SettledHoldInvoice.to_string();
+        let order = order.create(&pool).await.unwrap();
+        let order_id = order.id;
 
         let result = do_payment(&ctx, order, None).await;
 
@@ -2430,6 +2481,10 @@ mod tests {
             ),
             "the resolved invoice must be rejected before LND is contacted: {result:?}"
         );
+        // This branch used to run the bookkeeping inline; `do_payment` now
+        // owns it. `payment_attempts == 1` pins that it runs exactly once —
+        // counting twice would silently halve the buyer's retry budget.
+        assert_payout_failure_recorded(&pool, order_id).await;
 
         server.abort();
     }
--- a/src/app/release.rs
+++ b/src/app/release.rs
@@ -584,25 +584,55 @@ async fn handle_child_order(
 /// Lightning Addresses **and** bech32 LNURLs are resolved via
 /// [`resolv_ln_address`] under the LNURL host policy. A non-empty `pr` must
 /// decode as BOLT11 and pass [`validate_payout_invoice`] (chain match, final
-/// CLTV bound, not already expired) before LND submission; resolve, decode
-/// and validation failures — all pre-claim — go through
-/// [`check_failure_retries_or_log`] and return `Err`. `send_payment` RPC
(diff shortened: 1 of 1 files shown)


AndreaDiazCorreia (commented):
Reviewing this, it looks to me like it's solving the same thing as #864, which is also about `do_payment` early exits leaving `failed_payment = false` and is linked to #806.

Yours is broader (it covers the `?` exits too), so if that's right maybe we close #864 as superseded and add `Closes #806` here?

AndreaDiazCorreia (approved):
tACK

src/app/release.rs
```diff
@@ -98,6 +98,25 @@ pub async fn check_failure_retries(
...
+    //
+    // Only the payment-retry fields, to avoid overwriting fields modified by
+    // concurrent processes (dev_fee_paid, dev_fee_payment_hash, status, etc.)
+    sqlx::query("UPDATE orders SET failed_payment = ?, payment_attempts = ? WHERE id = ?")
+        .bind(order.failed_payment)
+        .bind(order.payment_attempts)
+        .bind(order.id)
+        .execute(pool)
+        .await
+        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
```
grunch, line 118: _🪵 Observability_ | _🟡 Minor_

**`check_failure_retries_or_log` now misreports a successful write as a failed one.**

With the write hoisted above `get_buyer_pubkey()`, the function can return `Err` *after* the bookkeeping has been persisted — that is exactly the case the new `do_payment_records_the_failure_when_the_buyer_pubkey_is_unusable` test pins. But the caller still collapses both outcomes:

```rust
Err(e) => {
    warn!("Order id {}: check_failure_retries failed: {:?}", order.id, e);
}
```

An operator reading `check_failure_retries failed: InvalidPubkey` will conclude the order was **not** marked and go looking for a manual DB write — the precise recovery step this PR exists to remove. The `has N failed payments retries` line, which is what regtest verification greps for, is also skipped on that path.

Since the persistence is no longer coupled to the rest, consider making that visible — e.g. have `check_failure_retries` log (or return) "retry state persisted" before the notification path, so the `Err` arm reads as "persisted, notification skipped" rather than "bookkeeping failed". The docstring on `check_failure_retries_or_log` ("surface bookkeeping failures") drifted in the same way.

src/app/release.rs
```diff
@@ -98,6 +98,25 @@ pub async fn check_failure_retries(
...
+    // order selectable by `find_failed_payment`, so it must not depend on the
+    // notification path below succeeding: resolving the buyer pubkey, and the
+    // three payload checks in the retries-exhausted branch, all return early.
+    // An order reaching here with an unusable buyer key would otherwise end up
+    // with neither the flag nor a payout marker — invisible to every recovery
+    // job, exactly the state this bookkeeping exists to prevent. Writing first
+    // also means the buyer is never told about a failure that was not stored.
+    //
+    // Only the payment-retry fields, to avoid overwriting fields modified by
+    // concurrent processes (dev_fee_paid, dev_fee_payment_hash, status, etc.)
```
grunch, line 111: _🧪 Test coverage_ | _🟡 Minor_

**The hoist also changes three paths the PR description doesn't mention, and no test pins them.**

Besides `get_buyer_pubkey`, the write now also happens before the three retries-exhausted validations (`InvalidAmount`, `InvalidOrderKind`, `InvalidOrderStatus`) — the comment above says so, but the consequence is bigger than "the flag survives".

Take the `amount - fee <= 0` branch. Before this PR, on the attempt that pushes `payment_attempts` to `retries_number`, `check_failure_retries` returned `Err(InvalidAmount)` *before* the `UPDATE`, so the final increment was never persisted. The row stayed at `retries_number - 1`, which keeps satisfying `job_retry_failed_payments`' `payment_attempts < retries_number` guard — an order that can never be paid was re-dispatched every `payment_retries_interval`, forever. The hoist terminates that loop. That's a real second fix riding along in this commit and it deserves a line in the PR body.

The three `check_failure_retries_rejects_*` tests exercise exactly these paths but still assert only the `Err` variant, so they pass identically with the write hoisted or not — nothing pins the new ordering. Suggest adding to each (or to one, table-style):

```rust
let db_order = Order::by_id(&pool, order.id).await.unwrap().unwrap();
assert!(db_order.failed_payment);
assert_eq!(db_order.payment_attempts, retries_number);
```

(they currently build the order in memory only, so they'd need `order.create(&pool)` first, like the new `do_payment` tests do).

src/app/release.rs
```diff
@@ -584,17 +593,59 @@ async fn handle_child_order(
...
+/// implements.
+///
+/// **Every `Err` returned from here is treated as a payment failure** by
+/// [`do_payment`], which runs [`check_failure_retries_or_log`] on it. That
+/// holds exactly today — this function is only reached once the order itself
+/// has been validated, so every error exit means "the buyer could not be
+/// paid" — and centralizing it is what keeps the bookkeeping off the
+/// individual `?` sites, where it was repeatedly forgotten. Keep it that way:
+/// an error that is *not* a payment failure would burn a retry and send the
+/// buyer a spurious `PaymentFailed`.
```
grunch, line 638: _📄 Contract / docs_ | _🟡 Minor_

**"Every error exit means the buyer could not be paid" is true; "every error exit is a *payment* failure" isn't — and the doc rules out precisely what two exits now do.**

This paragraph sets the rule ("an error that is *not* a payment failure would burn a retry and send the buyer a spurious `PaymentFailed`") and then asserts the current exits all satisfy it. Two of them don't, in the sense that matters to the buyer:

- `LndConnector::new().await?` — a cert/macaroon read failure. Mostro-side infrastructure. This is the case verified on regtest, so it's clearly known and considered acceptable.
- `claim_order_payout(...).await?` — a DB error. Also Mostro-side (and note the bookkeeping that follows writes to the same DB, so it will usually fail too and only log).

Both consume one of the buyer's finite `payment_attempts` and deliver a user-facing `PaymentFailed` for an outage the buyer had no part in. During a sustained LND or DB outage the budget drains at one attempt per `payment_retries_interval` until `payment_attempts >= retries_number`; at that point `job_retry_failed_payments` stops picking the order up entirely and recovery depends on the buyer submitting a fresh invoice — a Mostro outage converted into buyer homework, for an invoice that was never the problem.

I'm not asking to split the error classification in this PR (that's a bigger change and the current behaviour is still strictly better than stranding). But the doc should state the tradeoff it's making instead of implying it doesn't exist — something like "including LND client setup and the claim write, which are Mostro-side failures: they burn a retry deliberately, because the alternative (no bookkeeping) is the stranding this fix removes."

Same family, smaller: the `order.buyer_invoice.is_none()` exit now sends `PaymentFailed` for an order where no payment was ever attempted and nothing failed — `AddInvoice` is the semantically correct prompt there. Practically unreachable today (`FiatSent`/`Dispute` both imply an invoice), so a nit, not a request.

src/app/release.rs
```diff
@@ -584,17 +593,59 @@ async fn handle_child_order(
...
+    // the flag) nor `find_inflight_payouts` (which needs the marker), and it
+    // sits in settled-hold-invoice forever with the buyer never notified and
+    // no retry ever attempted.
+    match dispatch_payout(ctx, &order, request_id).await {
+        Ok(()) => Ok(()),
+        Err(e) => {
+            check_failure_retries_or_log(ctx, &order, request_id).await;
+            Err(e)
+        }
+    }
```
grunch, line 623: _🧹 Nitpick_ | _🔵 Low_

`do_payment` no longer consumes the order — both uses are behind `&`. Taking `&Order` would drop the now-pointless clone at `src/scheduler.rs:236` (`do_payment(&ctx, payment_failed.clone(), None)`) on the hot retry loop, and the `admin_settle` / `release_action` call sites just lose a move.

While there, the match collapses:

```rust
if let Err(e) = dispatch_payout(ctx, order, request_id).await {
    check_failure_retries_or_log(ctx, order, request_id).await;
    return Err(e);
}
Ok(())
```

grunch (approved):
## Review

Strict pass over the three commits. The core change is right: hoisting the bookkeeping from the individual `?` sites into a single wrapper around `dispatch_payout` closes the stranded-order class for good, and the `dispatch_payout` contract is documented where the next person will read it. Verified locally on `d76df6d`: `cargo fmt --check` clean, `cargo clippy --all-targets -- -D warnings` clean, `cargo test --bin mostrod app::release::` → 51 passed / 0 failed.

I also checked the two invariants the design leans on:

- **No error exit between the claim CAS and `tokio::spawn`.** Confirmed — after `claim_order_payout` there is only `ctx.keys().clone()`, two clones and the spawn, so `do_payment` can never run the bookkeeping on a payout that already left a marker behind. The doc comment's "two paths deliberately return `Ok(())`" is accurate.
- **No double counting.** The background-task paths (watcher `PaymentStatus::Failed`, `DispatchVerdict::ReArm`) and `reconcile_inflight_payout` are all post-claim and gated on `fail_order_payout`'s CAS, so none of them can overlap with the new wrapper. The LNURL test pinning `payment_attempts == 1` is the right guard to have added.

Nothing blocking. Four non-blocking comments below: one observability regression introduced by the new write ordering, one contract/doc gap around infrastructure-caused errors, one behaviour change that is a genuine (unstated) bug fix but has no test, and a small API nit.

**Verdict: APPROVE.**

PR #909, opened 2026-08-20
fix(ci): scope publishing to version tag pushes

AndreaDiazCorreia:
Both release workflows run on tag pushes and on `workflow_dispatch`, but neither distinguished the two. A manual run on any branch reached `cargo publish`, created a GitHub Release named after the branch, and overwrote the `latest` Docker tag.

Actions expressions have no regex, so "an exact `vX.Y.Z` tag push" cannot be written in an `if:` on its own. Both workflows resolve it in a job instead, and the tag filters are anchored (`v[0-9]+.[0-9]+.[0-9]+`) so a pre-release no longer starts a run at all.

**rust.yml** — a `version-tag` job publishes `is_release`, true only for a push of an exact `vX.Y.Z` tag, and `publish` and `release` gate on it. A `workflow_dispatch` can target a tag as well as a branch, hence the event check. The build jobs stay reachable from `workflow_dispatch`, so a manual run is still a way to check the cross-compilation before tagging. `release` keeps running independently of `publish`; the only change to its `needs` is `version-tag`. Both gates are now spelled the same way, `success() && ...`: the status function was already implied on `release`, but only one of the two said so.

**docker-build-startOs.yml** — the dispatch input defaulted to `latest`, so any manual run replaced the published image. A `resolve-tag` job now resolves the tag once for the whole run: a tag push publishes its exact `vX.Y.Z`, and everything else, a dispatch aimed at a tag included, defaults to `dev-<short-sha>` and rejects `latest` or `vX.Y.Z` as an explicit override. #610 added `check_stable` for this, but it only gated the second `latest` tag, not the input default.

`latest` is moved in a separate `promote-latest` job once both images are pushed. Deciding at build time froze the answer for the length of a multi-arch build, so a release tagged during that window could be overwritten by the older run finishing later. The promotion is independent of the tag that triggered it: it moves `latest` to the highest version tag whose images are already published. GitHub keeps one pending job per concurrency group, so a burst of releases can have its queued promotion replaced by a later one; because a promotion is only queued once its own images exist, whichever job survives the queue repairs `latest` for the ones that were dropped. It also never regresses on a backport run, and a failed build of the highest tag no longer freezes it.

That job answers "is this version published?" with `imagetools inspect`, so it has to be careful about what a non-zero exit means:

- **Only a missing manifest counts as "not published".** Reading every failure as absence let a rate limit, a 5xx or an expired token skip the newest release and move `latest` *back* to an older one, with the job green. That is a wrong publish, whereas failing to promote is recoverable, so `not found` / `manifest unknown` advances to the next candidate and anything else fails the job.
- **The scan is bounded to the five newest version tags.** `latest` belongs to one of them; walking all 85 costs two registry calls each and is also the path that ends with an old version being promoted.
- **A failing `git ls-remote` fails the step.** `set -euo pipefail` does not cover a process substitution, so an unreachable remote produced an empty tag list and the step ended on the "nothing to promote" path with a green check. The list is assigned before it is filtered, so `set -e` sees it.
- **An empty result is an error, not a warning.** This job is the only thing that moves `latest`, and on a release it runs after its own images were pushed, so finding nothing to promote is a broken state rather than a no-op.

Same file, while in there:

- Workflow context (`github.ref`, `github.event_name`, the dispatch input) read from environment variables instead of being interpolated into the `run:` scripts, plus `set -euo pipefail`. The tag reaches `$GITHUB_OUTPUT` only after an OCI charset check, so a newline cannot append a second `tag=` line.
- A `contents: read` permissions block. It was the only workflow handling secrets without one.
- A per-ref concurrency group for the builds, so two runs of the same ref cannot push at once. Per-ref rather than global, because a single group would drop an intermediate version tag when several tag pushes arrive together.
- The two build jobs, which differed only in the image and the Dockerfile, are a matrix.
- A `dry_run` dispatch input. The workflow does not run on pull requests and `promote-latest` is gated on a push, so the first real execution of any of this would otherwise be a release, with `latest` at stake. A dry run builds both images without pushing and prints the promotion target without writing it. Its `promote-latest` uses a separate concurrency group, since taking a slot in the real queue would cancel a pending promotion.
- The `type=gha` cache is scoped per image. With a single shared scope the two matrix jobs evicted each other's cache on every run.

`actionlint` 1.7.12 with `shellcheck` 0.11.0 is clean on the Docker workflow, and the four pre-existing SC2193 warnings there are gone. It still reports one SC2012 in `rust.yml`, in the checksums step this PR does not touch.

The `promote-latest` script was also exercised outside CI, by extracting the `run:` block and running it against stubbed `git` and `docker`: all images published promotes the newest; a 404 on the newest falls through to the next; a rate limit fails the job without writing any tag; a failing `git ls-remote` exits non-zero; a dry run resolves the target and writes nothing; and `v0.19.0-rc.1`, `v0.19.0foo` and `foo/v9.9.9` are all rejected as candidates.

## Summary by CodeRabbit

- **Release Improvements**
  - Manual and non-release Docker builds now use guarded development tags instead of release tags.
  - Docker publishing is queued per branch or tag, preventing unrelated builds from blocking one another.
  - Stable releases are recognized only for exact `vX.Y.Z` version tags, with `latest` assigned only to the highest version.
  - Docker image tags are validated against OCI naming rules.

- **Bug Fixes**
  - Crate, release, and Docker publishing no longer run for branches or invalid tag formats.

### DIFF
--- a/.github/workflows/docker-build-startOs.yml
+++ b/.github/workflows/docker-build-startOs.yml
@@ -20,17 +20,21 @@ jobs:
 
       - name: Set image tag for metadata
         id: set_tag
+        env:
+          INPUT_TAG: ${{ github.event.inputs.image_tag }}
         run: |
-          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+          set -euo pipefail
+          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
             echo "tag=${GITHUB_REF#refs/tags/}" >> "$GITHUB_OUTPUT"
           else
-            echo "tag=${{ github.event.inputs.image_tag || 'latest' }}" >> "$GITHUB_OUTPUT"
+            echo "tag=${INPUT_TAG:-latest}" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Check if stable release
         id: check_stable
         run: |
-          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" =~ ^refs/tags/v && "${{ github.ref }}" != *"-"* ]]; then
+          set -euo pipefail
+          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" =~ ^refs/tags/v && "$GITHUB_REF" != *"-"* ]]; then
             echo "is_stable=true" >> "$GITHUB_OUTPUT"
           else
             echo "is_stable=false" >> "$GITHUB_OUTPUT"
@@ -77,17 +81,21 @@ jobs:
 
       - name: Set image tag for metadata
         id: set_tag
+        env:
+          INPUT_TAG: ${{ github.event.inputs.image_tag }}
         run: |
-          if [[ "${{ github.ref }}" == refs/tags/* ]]; then
+          set -euo pipefail
+          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
             echo "tag=${GITHUB_REF#refs/tags/}" >> "$GITHUB_OUTPUT"
           else
-            echo "tag=${{ github.event.inputs.image_tag || 'latest' }}" >> "$GITHUB_OUTPUT"
+            echo "tag=${INPUT_TAG:-latest}" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Check if stable release
         id: check_stable
         run: |
-          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" =~ ^refs/tags/v && "${{ github.ref }}" != *"-"* ]]; then
+          set -euo pipefail
+          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" =~ ^refs/tags/v && "$GITHUB_REF" != *"-"* ]]; then
             echo "is_stable=true" >> "$GITHUB_OUTPUT"
           else
             echo "is_stable=false" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/docker-build-startOs.yml
+++ b/.github/workflows/docker-build-startOs.yml
@@ -65,7 +65,10 @@ jobs:
         id: check_stable
         run: |
           set -euo pipefail
-          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" =~ ^refs/tags/v && "$GITHUB_REF" != *"-"* ]]; then
+          # Anchored: the v*.*.* trigger is a glob, so it also admits refs like
+          # v1.2.3foo. Only an exact vX.Y.Z names a stable release.
+          release_re='^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$'
+          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" =~ $release_re ]]; then
             echo "is_stable=true" >> "$GITHUB_OUTPUT"
           else
             echo "is_stable=false" >> "$GITHUB_OUTPUT"
@@ -143,7 +146,10 @@ jobs:
         id: check_stable
         run: |
           set -euo pipefail
-          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" =~ ^refs/tags/v && "$GITHUB_REF" != *"-"* ]]; then
+          # Anchored: the v*.*.* trigger is a glob, so it also admits refs like
+          # v1.2.3foo. Only an exact vX.Y.Z names a stable release.
+          release_re='^refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$'
+          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" =~ $release_re ]]; then
             echo "is_stable=true" >> "$GITHUB_OUTPUT"
           else
             echo "is_stable=false" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/docker-build-startOs.yml
+++ b/.github/workflows/docker-build-startOs.yml
@@ -182,8 +182,8 @@ jobs:
           # than from a checkout of the whole history.
           tags="$(git ls-remote --tags --refs "$GITHUB_SERVER_URL/$GITHUB_REPOSITORY" 'v*')"
           mapfile -t candidates < <(
-            sed 's#.*refs/tags/##' <<<"$tags" \
-              | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' \
+            grep -oE 'refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$' <<<"$tags" \
+              | cut -d/ -f3 \
               | sort -rV \
               | head -n "$max_candidates"
           )
--- a/.github/workflows/docker-build-startOs.yml
+++ b/.github/workflows/docker-build-startOs.yml
@@ -37,11 +37,12 @@ jobs:
           INPUT_TAG: ${{ github.event.inputs.image_tag }}
         run: |
           set -euo pipefail
-          if [[ "$GITHUB_REF" == refs/tags/* ]]; then
+          if [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF" == refs/tags/* ]]; then
             tag="${GITHUB_REF#refs/tags/}"
           else
-            # A manual run gets a build-specific tag. latest and vX.Y.Z name
-            # published images, so only a tag push is allowed to write them.
+            # Anything that is not a tag push takes the guarded path, including a
+            # dispatch aimed at a tag: latest and vX.Y.Z name published images,
+            # so only a tag push is allowed to write them.
             tag="${INPUT_TAG:-dev-${GITHUB_SHA::7}}"
             release_re='^v[0-9]+\.[0-9]+\.[0-9]+'
             if [[ "$tag" == "latest" || "$tag" =~ $release_re ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build-startOs.yml
```diff
@@ -125,3 +173,52 @@ jobs:
...
+  # build, so a release tagged during that window could be overwritten by the
+  # older run finishing later. The group is shared by every ref, which makes
+  # this the one place where release runs serialize, and the tag list is
+  # refetched inside it so the decision is current at the moment of the move.
+  promote-latest:
+    runs-on: ubuntu-latest
+    needs: [build-and-push-plain, build-and-push-startos]
+    if: github.event_name == 'push'
+    concurrency:
+      group: docker-latest-promotion
```
grunch: _🗄️ Data Integrity & Integration_ | _🟠 Major_

**A shared concurrency group can silently drop a `latest` promotion.**

GitHub keeps at most one running and one *pending* job per concurrency group; when a third job arrives, the previously pending one is **cancelled** (the workflow-level comment at lines 20-23 of this file relies on exactly this behaviour to justify the per-ref group for the builds). `promote-latest` now shares one group across every release ref, so the same drop applies here, and the promotion logic only moves `latest` when `$tag == $newest`. Combined, this is a reachable "latest never advances" state:

1. `v1.2.3`, `v1.2.4`, `v1.2.5` are pushed close together (patch burst / backfill). Builds run in parallel per-ref and finish in arbitrary order.
2. `v1.2.3`'s `promote-latest` starts → sees `newest=v1.2.5` → *"leaving latest untouched"* (correct).
3. `v1.2.5`'s `promote-latest` is queued (pending). Then `v1.2.4`'s finishes building and is queued → **`v1.2.5`'s pending job is cancelled.**
4. `v1.2.4`'s `promote-latest` runs → `newest=v1.2.5 != v1.2.4` → untouched.

Result: every image exists in the registry, but `latest` still points at `v1.2.2`, and the `v1.2.5` run shows as *cancelled*. Nothing left in the queue will ever move it; it needs a manual re-run. This is the property the PR description (and the last two review rounds) say is now guaranteed.

The fix is to make the promotion idempotent and independent of *which* tag triggered it: promote the highest version tag whose images are already published. Since a `promote-latest` job is only queued after its own builds finished, whichever promote job survives the queue always sees the images of the ones that were cancelled, so `latest` converges. It also never regresses (a `v1.2.3` backport run finds `v1.2.4` images first and leaves it there), and it covers the "highest tag's build failed" case (the previous published version wins instead of `latest` freezing).

```suggestion
      - name: Move latest to the highest published release
        run: |
          set -euo pipefail
          git fetch --tags --force --prune
          # Independent of which tag triggered this run: a pending promote job can
          # be replaced in the shared queue, so whichever run survives repairs
          # latest for all of them. A promote job is only queued once its own
          # images are pushed, so the survivor always sees the cancelled ones.
          for candidate in $(git tag --list | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -rV); do
            published=true
            for image in mostrop2p/mostro mostrop2p/mostro-startos; do
              docker buildx imagetools inspect "$image:$candidate" >/dev/null 2>&1 || { published=false; break; }
            done
            [[ "$published" == true ]] || continue
            for image in mostrop2p/mostro mostrop2p/mostro-startos; do
              docker buildx imagetools create -t "$image:latest" "$image:$candidate"
            done
            echo "latest -> $candidate"
            exit 0
          done
          echo "::warning::no published version tag found, leaving latest untouched"
```

(The `$GITHUB_REF` release check at the top of the step becomes redundant with this; keep it if you want the early log line, but the job-level `if: github.event_name == 'push'` plus the `needs` already guarantee it.)

.github/workflows/docker-build-startOs.yml
```diff
@@ -20,21 +34,39 @@ jobs:
 
       - name: Set image tag for metadata
```
grunch, line 45: _🧹 Nitpick_ | _Refactor_

**Identical 35-line `set_tag` block in both jobs.**

Both build jobs carry byte-for-byte the same validation script; the two copies must be kept in sync by hand, and each job independently fails instead of the run failing once up front. Extracting it into a small `resolve-tag` job with an `outputs.tag` that both build jobs `need` (and `promote-latest` can reuse) removes the duplication and rejects a bad ref/input before any QEMU/buildx setup runs.

Same goes for the two build jobs themselves: they differ only in `images:` and `file:`, which is a textbook `strategy.matrix` with `include:`.

.github/workflows/docker-build-startOs.yml
```diff
@@ -125,3 +173,52 @@ jobs:
...
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Move latest to this release
+        run: |
+          set -euo pipefail
+          git fetch --tags --force --prune
```
grunch: _🧹 Nitpick_

`actions/checkout@v6` with `fetch-depth: 0` + `fetch-tags: true` already runs inside the job, i.e. after the concurrency slot was acquired, so this `git fetch --tags --force --prune` re-fetches what was fetched seconds earlier. Conversely, the job doesn't need the working tree at all — `git ls-remote --tags --refs origin 'v*'` (or `gh api repos/.../git/matching-refs/tags/v`) would give the same list without a full-history clone. Either drop the extra fetch or drop the checkout; keeping both is the one combination that buys nothing.

grunch (changes requested):
**Actionable comments posted: 3** (1 Major, 2 nitpicks)

Reviewed head `1face8a` against `main` (`9caa5f9`). Two workflow files, no Rust changes; CI on the exact head is green (`fmt`, `clippy`, `MSRV`, `test`, `build`). I simulated the `set_tag` / `version-tag` predicates locally for push/dispatch × valid/rc/suffixed/`latest`/`vX.Y.Z`/newline/leading-dash inputs and they all resolve as the comments in the file describe — that part of the PR is solid. `actionlint` isn't available on my machine, so I'm relying on the author's and @ermeme's runs for that.

### 🟠 Major — blocking
- **`promote-latest` can drop a promotion** (`docker-build-startOs.yml:188`). The shared `docker-latest-promotion` group inherits GitHub's one-pending-job-per-group semantics — the very behaviour the per-ref build group is documented to avoid — and the step only moves `latest` when the triggering tag is the newest. Three overlapping release pushes can leave `latest` stale forever with one run shown as *cancelled*. Concrete sequence and an idempotent fix (promote the highest *published* version, regardless of trigger) are in the inline comment. Since this is exactly the guarantee the last three review rounds converged on, I'd rather get it right before merge.

### 🟡 Minor (outside the diff, so not inline)
- **`on.push.tags: 'v*.*.*'`** (`docker-build-startOs.yml:4-6`, same in `rust.yml:5`). GitHub filter patterns support `+` and `[]`, and they are anchored, so `v[0-9]+.[0-9]+.[0-9]+` would stop `v1.2.3-rc.1` / `v1.2.3foo` from starting a run at all. Right now every pre-release tag produces a red ❌ run in the Actions tab (the build jobs `exit 1` by design). Keep the in-script regex as defence in depth, but not triggering is cleaner than triggering-to-fail.

### 🔵 Nitpicks
- Duplicated `set_tag` block / near-identical build jobs → extract a `resolve-tag` job + matrix (inline).
- `promote-latest`: full-history checkout *and* a manual `git fetch --tags` — one of the two is redundant (inline).
- PR description: "`release.needs` is unchanged" — it now includes `version-tag`. Trivial, but the description is what ends up in the merge commit.

### Verified
- Manual dispatch (branch or tag) can no longer reach `cargo publish`, `action-gh-release`, or any release/`latest` Docker tag. ✅
- Malformed `v*.*.*` tag pushes fail before `metadata-action`/`build-push-action`. ✅
- Tag/input values only reach shell through `env:`; output written after the OCI charset check, so no `$GITHUB_OUTPUT` injection. ✅
- `permissions: contents: read` is sufficient (`type=gha` cache uses the runtime token, not `GITHUB_TOKEN`). ✅
- `if:` conditions on `publish`/`release` still imply `success()` (no status function used), so a failed `build` can't be bypassed. ✅

Catrya (changes requested):
Good direction, and the parts I could exercise locally hold up: I ran the
`resolve-tag` script against 12 cases and a dispatch aimed at a tag now
resolves to `dev-<sha>` (the bypass flagged earlier is closed), `latest` and
`vX.Y.Z` inputs are rejected, and an input with a newline is caught by the OCI
charset check before it reaches `$GITHUB_OUTPUT`. The "highest published
version" promotion does close the dropped-queue hole: a promotion is only
queued once its own images exist, so the survivor always runs later than the
one it replaced and repairs `latest` for it. The matrix is a faithful dedup of
the two jobs, and no pre-release tag has ever existed in this repo (85 tags,
none suffixed), so the anchored filter costs nothing.

Two things in `promote-latest` need fixing before this can ship, because both
fail in the worst possible way for a release path.

### 1. `latest` can be moved *backwards*

```bash
docker buildx imagetools inspect "$image:$candidate" >/dev/null 2>&1 || { published=false; break; }

Any non-zero exit is read as "not published" — a Docker Hub 500, a rate limit,
an expired token, a timeout. If that hits while inspecting v0.19.0, the loop
falls through to the next candidate and runs
imagetools create -t latest v0.18.5. A transient network error silently
demotes latest to the previous release, with the job green, which is the
opposite of what the job comment promises ("never regresses"). Failing to
promote is recoverable; promoting backwards is a wrong publish.

Please separate "manifest unknown / 404" from every other failure and
exit 1 on the latter. While in there, bound the scan to the first few
candidates — worst case today is 85 tags × 2 inspects = 170 registry calls,
and that walk is exactly the path to the regression above.

2. Tag enumeration fails silently

set -euo pipefail does not cover a process substitution. Reproduced locally
with an unreachable remote:

fatal: could not read Username for 'https://github.com'
candidates count: 0
step exit=0

The step prints ::warning::no published version tag found and exits 0. Since
this job is now the only thing that moves latest, a transient git ls-remote
failure leaves latest stale on a fully green release.

Assign first so set -e applies (tags="$(git ls-remote ...)"), and make that
final warning an ::error:: + exit 1: the job only runs on a version tag
push, so "no published version tag" is never a normal state.

3. None of this path is exercised by CI (not blocking, but please address)

The Docker workflow doesn't run on PRs — the head's check runs are only
build/test/clippy/fmt/MSRV — and promote-latest is gated on
github.event_name == 'push', so it can't even be smoke-tested with a manual
dispatch. The first real execution of resolve-tag, the matrix and
promote-latest will be a release, with latest at stake. A dry_run input
that resolves and prints the target without calling imagetools create would
cover it, and would also confirm docker buildx imagetools works in that job
without setup-buildx-action.

Nits: cache-from/to: type=gha has no scope, so the two matrix images evict
each other's cache (pre-existing, but you're on those lines —
scope=${{ matrix.name }} fixes it); and publish keeps success() && while
release doesn't — same semantics, worth unifying.

Catrya (approved):
ACK

grunch (approved):
ACK

PR #914, opened 2026-08-22
fix: advertise the plain invoice_expiration_window

### LINKED ISSUE #895: Info event advertises hold_invoice_expiration_window under the invoice_expiration_window tag
## Summary

The kind-38385 info event publishes the wrong value for `invoice_expiration_window`: it is filled from `ln_settings.hold_invoice_expiration_window` instead of `ln_settings.invoice_expiration_window`. Clients that use the advertised value to build a payout invoice get their invoice rejected with a `cant-do`.

mostrod **0.18.4** (`2f2b813`).

## The bug

`src/nip33.rs:622-633`:

```rust
Tag::custom(
    "hold_invoice_expiration_window",
    vec![ln_settings.hold_invoice_expiration_window.to_string()],   // correct
),
Tag::custom(
    "hold_invoice_cltv_delta",
    vec![ln_settings.hold_invoice_cltv_delta.to_string()],
),
Tag::custom(
    "invoice_expiration_window",
    vec![ln_settings.hold_invoice_expiration_window.to_string()],   // <-- wrong field
),
```

The third tag should read `ln_settings.invoice_expiration_window`. It looks like a copy-paste from the first.

## Observed

Daemon configured with two clearly different values:

```toml
[lightning]
invoice_expiration_window = 600
hold_invoice_expiration_window = 120
```

The published info event:

```
['hold_invoice_expiration_window', '120']
['invoice_expiration_window', '120']      <-- should be 600
```

Both report `120`. The enforced value is `600`.

## Impact

`invoice_expiration_window` is the minimum remaining lifetime a buyer payout invoice must have. It is enforced in `src/lightning/invoice.rs:214-223`:

```rust
let expiration_window = ln_settings.invoice_expiration_window as i64;
let latest_date = Utc::now() + TimeDelta::try_seconds(expiration_window)…;
if expires_at < latest_date {
    return Err(MostroInternalErr(ServiceError::InvoiceInvalidError));
}
```

So the one number a client needs in order to build an acceptable invoice is precisely the number the daemon misreports. A client that trusts the info event and builds a 120 s-plus-margin invoice against a daemon enforcing 600 s gets a `cant-do`, and the order sits in `waiting-buyer-invoice` until it is reaped.

This is exactly what happened in the [`ortsom`](https://github.com/MostroP2P/ortsom) harness: the failing scenario could not use the advertised value and had to fall back to a locally declared expectation, with a comment explaining that this tag cannot be trusted on 0.18.4 (MostroP2P/ortsom#9). Any other client doing invoice construction from the info event will hit the same wall.

The mismatch is also silently self-consistent when the two settings happen to be equal, which is why it can go unnoticed — mostrod's own template ships `invoice_expiration_window = 3600` and `hold_invoice_expiration_window = 300`, so real deployments do diverge.

## Suggested fix

One-line change in `src/nip33.rs:632`:

```diff
 Tag::custom(
     "invoice_expiration_window",
-    vec![ln_settings.hold_invoice_expiration_window.to_string()],
+    vec![ln_settings.invoice_expiration_window.to_string()],
 ),
```

Worth adding a regression test alongside the existing tag tests in that module, asserting that the two windows are emitted from **different** settings fields — building the tags from a `LightningSettings` whose two windows differ and asserting each tag carries its own value. A test using equal values would pass either way, which is likely how this survived.

Happy to send the PR if useful.

## Environment

- mostrod 0.18.4, commit `2f2b813`
- LND 0.20.0-beta, regtest (Polar)
- Local relay, protocol v2 (NIP-44)

21Mill:
Closes #895.

## The bug

The kind-38385 info event filled the `invoice_expiration_window` tag from
`ln_settings.hold_invoice_expiration_window` — a copy-paste from the tag two
entries above, so both tags published the same number.

The two are independent settings: the hold invoice window is the lifetime of
the invoice the seller pays, while `invoice_expiration_window` is the minimum
remaining lifetime a buyer payout invoice must have, enforced in
`src/lightning/invoice.rs`. So the one number a client needs in order to build
an acceptable payout invoice was precisely the one the daemon misreported. A
client that trusted the info event got a `cant-do` back, and the order sat in
`waiting-buyer-invoice` until it was reaped.

## The change

- Read the field each tag names — the fix itself.
- Extract the three `ln_settings`-derived tags into `ln_policy_tags`, mirroring
  the existing `bond_policy_tags`. The helper is spliced in at the original
  position, so the wire order of the info event's tags is unchanged.
- Document what separates the two windows on the settings fields themselves
  (`src/config/types.rs`), where the doc comments only paraphrased the names.

## Why the extraction

Without it the regression test would be vacuous. `test_settings()` builds
`LightningSettings::default()`, which leaves *both* windows at `0`, and
`MOSTRO_CONFIG` is a process-wide OnceLock no test can mutate — so a test
driven through `info_to_tags()` passes just as happily with the fields swapped.
Testing the pure helper allows pinning each window against a distinct value.
This is the same reason `bond_policy_tags` was split out, and its doc comment
says so.

## Verification

The regression test was confirmed to fail against the original bug before the
fix was applied:

```text
---- nip33::tests::ln_policy_tags_reports_each_window_from_its_own_field stdout ----
assertion `left == right` failed: invoice_expiration_window must report the buyer payout window
  left: Some("120")
 right: Some("600")
```

With the fix, the full suite is green:

```text
test result: ok. 1234 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 16.44s
```

`cargo clippy --all-targets --all-features -- -D warnings` is clean and
`cargo fmt` has been applied. No schema or config changes: the tag name and the
settings keys are untouched, only the value the tag is filled from. Clients
pick up the corrected value on the next info event.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected Lightning policy information so buyer invoice expiration windows are reported separately from seller-paid hold invoice lifetimes.
  - Ensured hold-invoice, hold-CLTV, and invoice-expiration settings are each represented accurately.

- **Documentation**
  - Clarified the distinction between buyer payout invoice expiration and seller-paid hold invoice lifetime in Lightning settings documentation.

- **Tests**
  - Added coverage to verify the two expiration values remain independent.

### DIFF
--- a/src/config/types.rs
+++ b/src/config/types.rs
@@ -446,11 +446,13 @@ pub struct LightningSettings {
     pub lnd_macaroon_file: String,
     /// LND gRPC host
     pub lnd_grpc_host: String,
-    /// Invoice expiration window in seconds
+    /// Minimum remaining lifetime, in seconds, a buyer payout invoice must
+    /// have to be accepted. Enforced in `lightning::invoice`.
     pub invoice_expiration_window: u32,
     /// Hold invoice CLTV delta
     pub hold_invoice_cltv_delta: u32,
-    /// Hold invoice expiration window in seconds
+    /// Lifetime, in seconds, of the hold invoice the seller pays. Distinct
+    /// from `invoice_expiration_window`, which governs the buyer's payout.
     pub hold_invoice_expiration_window: u32,
     /// Number of payment attempts
     pub payment_attempts: u32,
--- a/src/nip33.rs
+++ b/src/nip33.rs
@@ -1,6 +1,6 @@
 use crate::config::constants::NOSTR_EXCHANGE_RATES_EVENT_KIND;
 use crate::config::settings::Settings;
-use crate::config::types::{BondApplyTo, MostroSettings};
+use crate::config::types::{BondApplyTo, LightningSettings, MostroSettings};
 use crate::lightning::LnStatus;
 use crate::util::{get_expiration_timestamp_for_kind, get_keys, monotonic_dispute_event_timestamp};
 use crate::LN_STATUS;
@@ -631,18 +631,13 @@ pub fn info_to_tags(ln_status: &LnStatus) -> Tags {
         // `transport` setting so clients pick the right wire format before
         // sending anything. See docs/TRANSPORT_V2_SPEC.md.
         Tag::custom("protocol_version", vec![protocol_version.to_string()]),
-        Tag::custom(
-            "hold_invoice_expiration_window",
-            vec![ln_settings.hold_invoice_expiration_window.to_string()],
-        ),
-        Tag::custom(
-            "hold_invoice_cltv_delta",
-            vec![ln_settings.hold_invoice_cltv_delta.to_string()],
-        ),
-        Tag::custom(
-            "invoice_expiration_window",
-            vec![ln_settings.hold_invoice_expiration_window.to_string()],
-        ),
+    ];
+
+    // Spliced in at its original position: the extraction leaves the wire
+    // order of the info event unchanged.
+    tags_vec.extend(ln_policy_tags(ln_settings));
+
+    tags_vec.extend([
         Tag::custom("lnd_version", vec![ln_status.version.to_string()]),
         Tag::custom("lnd_node_pubkey", vec![ln_status.node_pubkey.to_string()]),
         Tag::custom("lnd_commit_hash", vec![ln_status.commit_hash.to_string()]),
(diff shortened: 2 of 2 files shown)


Catrya (changes requested):
The fix itself is right. Blocking on one doc comment this PR gets wrong.

**`hold_invoice_expiration_window`: "Lifetime, in seconds, of the hold invoice
the seller pays" is false.** After this PR nothing reads that setting except
the info tag itself: `create_hold_invoice` (`src/lightning/mod.rs:150-156`)
passes only `cltv_expiry` and leaves `expiry` at `0`, so LND applies its
default of 86400s. The invoice lives 24h while we advertise 300s, the same
class of bug this PR is fixing on the sibling tag, now stated explicitly in a
doc comment instead of merely implied. Please document what's true and open a
follow-up for the real fix:

    /// Advertised in the info event only, it is *not* applied to the hold
    /// invoice, which is created without an `expiry` and so uses LND's
    /// default (86400s). Distinct from `invoice_expiration_window`, which
    /// governs the payout invoice. See #<follow-up>.

Two non-blocking notes while you're in there:

- `invoice_expiration_window` also gates bond payout invoices
  (`src/app/bond/payout.rs:1391`), and does *not* apply to LNURL /
  Lightning-address destinations (`src/lightning/invoice.rs:56-59`), so
  "a buyer payout invoice" is off in both directions.
- No test asserts `info_to_tags` still emits the block — drop the
  `extend(ln_policy_tags(...))` line and the suite stays green.

21Mill:
Closing as superseded by #952, which landed the same fix on 2026-09-04.

That PR corrects the tag the same way this one did and goes one better: `invoice_window_tags` is the equivalent of the helper extracted here, its test pins 3600/300 against distinct values for the same reason (the shared test settings leave both windows at `0`, so the natural test passes with the bug present), and `info_to_tags_emits_both_invoice_window_tags` closes the gap @Catrya flagged in review — nothing here asserted `info_to_tags` still emitted the block, so deleting the call left the suite green.

Rebasing this onto `main` conflicts squarely with #952 in `info_to_tags`, and there is nothing left to salvage from the diff.

Two things this PR raised that #952 does not cover, so they do not get lost:

- @Catrya was right that the doc comment added here was false, and the underlying problem outlives it. It turned out to be wider than a wrong comment: `hold_invoice_expiration_window` has no reader at all, and the deadline its template entry describes is enforced by `expiration_seconds` under a different name and a different value. Filed as #955.
- #895 is still open even though its bug is fixed — #952 did not link it. Worth closing.

Thanks for the review, @Catrya — the blocking point was correct and it is what led to #955.

PR #922, opened 2026-08-24
test(dispute): cover CantDo(DisputeCreationError) inconsistent flag path

### LINKED ISSUE #907: Test gap: no coverage for the new CantDo(DisputeCreationError) path in dispute_action
## Summary

#848 changes `dispute_action` so that a `setup_dispute` failure is propagated to the client instead of being silently swallowed. That is the right call, but the new error arm has **no test**: `grep -rn "DisputeCreationError" src/` returns nothing.

This is a behavior change on a user-facing protocol response, so it should be pinned by a test before it can regress.

## Context

Before #848, `src/app/dispute.rs` did:

```rust
if order.setup_dispute(is_buyer_dispute).is_ok() {
    order.clone().update(pool).await.map_err(...)?;
}
// execution continued, and the dispute row was created anyway
```

On failure this skipped `order.update()` but kept going, creating a `disputes` row while leaving the order flags unset. After #848:

```rust
order.setup_dispute(is_buyer_dispute).map_err(MostroCantDo)?;
order.clone().update(pool).await.map_err(...)?;
```

The caller now receives `CantDo(DisputeCreationError)` and no dispute row is created.

`setup_dispute` (mostro-core 0.14.5, `order.rs:522`) returns `Err(CantDoReason::DisputeCreationError)` only when the initiating side's dispute flag was **already** set. `dispute_action` guards at the top with `find_dispute_by_order_id(...).is_ok()` → `DisputeAlreadyExists`, so the ordinary double-dispute flow never reaches `setup_dispute` twice. The new arm is reachable only from an inconsistent database state: an order with `buyer_dispute`/`seller_dispute` set but **no** matching row in `disputes`.

That state should not arise, and failing loudly on a violated precondition is the correct behavior — this was discussed by @ToRyVand in the #848 review thread. But "should not arise" is exactly the kind of assumption that deserves a test, both to lock in the semantics and to document the trade-off in executable form.

## Current coverage in `src/app/dispute.rs`

The `dispute_action_*` tests already cover: missing order id (`NotFound`), unknown order id (`NotFound`), order that already has a dispute (`DisputeAlreadyExists`), non-disputable status, missing seller pubkey, missing buyer pubkey, sender that is not a party, and both the buyer-initiated and seller-initiated happy paths.

The `setup_dispute` error arm is the one gap.

## What to do

Add one test to the `mod tests` block in `src/app/dispute.rs`, next to `dispute_action_rejects_order_that_already_has_a_dispute`.

### How to build the state

Follow the existing helpers in that module (`create_test_pool`, and the setup used by `dispute_action_buyer_initiated_flow_persists_dispute_and_notifies`):

1. Create a test pool and insert an order in a disputable status (`fiat-sent` or `active`) with both `buyer_pubkey` and `seller_pubkey` set.
2. Set `buyer_dispute = 1` on that order row — directly via `sqlx::query("UPDATE orders SET buyer_dispute = 1 WHERE id = ?1")` — and insert **no** row into `disputes`. This is the inconsistent state described above; going through the normal flow would trip the `DisputeAlreadyExists` guard instead and never reach `setup_dispute`.
3. Build the `Message` / `UnwrappedMessage` with the **buyer** as sender, so `get_counterpart_info` yields `is_buyer_dispute == true` and `setup_dispute` hits the already-set flag.

### What to assert

```rust
let err = dispute_action(&ctx, msg, &event, &my_keys)
    .await
    .expect_err("inconsistent dispute state must be rejected");
assert!(matches!(
    err,
    MostroError::MostroCantDo(CantDoReason::DisputeCreationError)
));
```

Plus the side-effect assertion that distinguishes new behavior from old — this is the part that actually catches a regression:

```rust
// The old code created the dispute row anyway; the new code must not.
assert!(find_dispute_by_order_id(&pool, order_id).await.is_err());
```

Optionally also assert the order row is untouched (status unchanged, `seller_dispute` still 0), confirming nothing was persisted before the early return.

### Related nit worth folding in

`setup_dispute` sets `self.status = Status::Dispute.to_string()` **before** its error return (mostro-core 0.14.5, `order.rs:537`), so the local `order` is left dirty on `Err`. The current code returns before any persist, so this is harmless today. While adding the test, add the one-line comment that pins the invariant:

```rust
// setup_dispute leaves order.status dirty on Err; we return before any persist.
order.setup_dispute(is_buyer_dispute).map_err(MostroCantDo)?;
```

## Acceptance criteria

- [ ] A test in `src/app/dispute.rs` asserts `dispute_action` returns `CantDo(DisputeCreationError)` for an order whose dispute flag is set with no matching `disputes` row.
- [ ] The same test asserts no dispute row is created — the behavior that changed in #848.
- [ ] `cargo test --all`, `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` are clean.

Severity: **Low** (test gap, no bug). Can be resolved directly in #848 if it has not merged yet; otherwise as a follow-up.

VedantMadane:
## Summary
#848 propagates `setup_dispute` failures as `CantDo(DisputeCreationError)`, but nothing exercised that arm.

Add a unit test that constructs the inconsistent DB state the issue describes (initiator dispute flag set, no `disputes` row) and asserts:
- `dispute_action` returns `MostroCantDo(DisputeCreationError)`
- no dispute row is created
- order status stays `Active`

## Test plan
- [ ] `cargo test dispute_action_returns_dispute_creation_error_when_flag_set_without_dispute_row`
- [ ] Existing `dispute` module tests still pass

Fixes #907

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for inconsistent dispute data.
  * Verifies dispute actions fail safely without creating duplicate records or changing the order status.

### DIFF
--- a/src/app/dispute.rs
+++ b/src/app/dispute.rs
@@ -544,6 +544,49 @@ mod tests {
         ));
     }
 
+    /// Inconsistent DB state: the initiating side's dispute flag is already set,
+    /// but there is no `disputes` row (so the early `DisputeAlreadyExists` guard
+    /// does not fire). `setup_dispute` must fail with `DisputeCreationError`, and
+    /// no dispute row may be created — the path introduced by #848.
+    #[tokio::test]
+    async fn dispute_action_returns_dispute_creation_error_when_flag_set_without_dispute_row() {
+        let pool = create_test_pool().await;
+        let ctx = build_ctx(&pool);
+        let buyer = Keys::generate().public_key();
+        let seller = Keys::generate().public_key();
+
+        let mut order = create_order(Some(buyer), Some(seller), Status::Active);
+        order.buyer_dispute = true;
+        let order = order.create(&pool).await.unwrap();
+
+        let event = create_event(buyer);
+        let result = dispute_action(
+            &ctx,
+            dispute_msg_for(Some(order.id)),
+            &event,
+            &Keys::generate(),
+        )
+        .await;
+
+        assert!(
+            matches!(
+                result,
+                Err(MostroCantDo(CantDoReason::DisputeCreationError))
+            ),
+            "expected DisputeCreationError, got {result:?}"
+        );
+        assert!(
+            find_dispute_by_order_id(&pool, order.id).await.is_err(),
+            "no dispute row must be created when setup_dispute fails"
+        );
+        let stored = Order::by_id(&pool, order.id).await.unwrap().unwrap();
+        assert_eq!(
+            stored.status,
+            Status::Active.to_string(),
+            "order must not move to Dispute when setup_dispute fails"
+        );
+    }
+
     /// Full buyer-initiated flow on an `Active` order. All DB side effects
     /// (order flags/status, dispute row) and both queue notifications happen
     /// before the final Nostr publish, which fails offline (default client


Catrya (changes requested):
duplicate #916

Catrya:
Hi @VedantMadane thanks for your work, good job! but this is already addressed in PR #916.
Will close this PR as duplicated. New contributions are welcome.

PR #923, opened 2026-08-24
fix(dispute): make the dispute writes atomic (#921)

### LINKED ISSUE #921: dispute_action persists the dispute flag before creating the dispute row
## Summary

`dispute_action` persists the order's dispute flag **before** it inserts the `disputes` row, in two separate statements with no transaction. If the insert fails, the order keeps a persisted `buyer_dispute`/`seller_dispute` with no matching dispute row — and the initiating party is then permanently unable to dispute that order.

This is the state #907 describes as one that "should not arise" and that #916's test builds by hand with `sqlx::query`. It arises from this function.

## The window

`src/app/dispute.rs`, `dispute_action`:

```rust
order
    .clone()
    .update(pool)                    // buyer_dispute = 1 is now persisted
    .await
    .map_err(|cause| MostroInternalErr(ServiceError::DbAccessError(cause.to_string())))?;

// Save dispute to database
let dispute = dispute
    .create(pool)                    // the row — may fail
    .await
    .map_err(|cause| MostroInternalErr(ServiceError::DbAccessError(cause.to_string())))?;
```

A `DbAccessError` on `create` returns to the client, leaving the order row flagged and the `disputes` table without its row. Nothing later reconciles it.

## Why this is the wrong ordering

The codebase already states the invariant, in `job_escrow_deadline` (`src/scheduler.rs`), which performs the same two writes in the opposite order:

> Row first, then the status flip: a dispute row pointing at a `fiat-sent` order is recoverable (the next tick sees it and stops), a `dispute` order with no row would be invisible.

That path also guards with `if disputed.buyer_dispute { ... }` before calling `setup_dispute`, precisely because it knows the already-set flag returns `Err`. `dispute_action` has the two writes the other way around and no such guard.

## Impact

Given an order left in this state:

- The initiating party can never dispute it. Every retry reaches `setup_dispute` with the flag already set and returns `CantDo(DisputeCreationError)` — the arm #916 now covers.
- The counterparty is unaffected: their flag is still unset, so `setup_dispute` succeeds for them.
- With no `disputes` row, solvers cannot see the order at all — "invisible", in the scheduler's own words.

Reaching it requires a DB failure on the insert, so this is not a common path. But the failure is silent, one-sided and permanent, which is what makes it worth fixing rather than documenting.

## Suggested fix

Create the dispute row first, then persist the order flag — matching `job_escrow_deadline` and its stated reasoning.

A transaction around both writes would be the stronger fix, and `pool.begin()` is already used elsewhere (`src/app/rate_user.rs`, `src/db.rs`), but it is not cheap here: `Crud::create` and `Crud::update` (mostro-core 0.14.5, `src/db/mod.rs`) take `&Pool<Sqlite>` rather than a generic executor, so a transaction means either a mostro-core change or replacing both calls with raw `sqlx`. The reorder needs neither and is consistent with the precedent already in the tree.

## Note on #907 and #916

#907 states the arm "is reachable only from an inconsistent database state" and that the state "should not arise". #916 implemented exactly that, correctly — its test and its call-site comment both check out, and the test pins the resulting behaviour whichever way this issue is resolved. The premise is the only thing that needs revising: the state does arise, from `dispute_action` itself.

## Acceptance criteria

- [ ] `dispute_action` cannot leave an order with a dispute flag persisted and no `disputes` row when the row insert fails.
- [ ] A test covers the failure ordering.
- [ ] `cargo test`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo fmt --check` clean.

Severity: **Low-Medium** — requires a DB failure to trigger, but the resulting lockout is silent and permanent.

Happy to pick this up. @21Mill found the ground here via #916 — if you would rather take it, it is yours.

ToRyVand:
Closes #921. Taken with @21Mill's go-ahead ([comment](https://github.com/MostroP2P/mostro/issues/921#issuecomment-5390349116)).

## The problem

`dispute_action` wrote the order's dispute flag and its `Dispute` status, and the `disputes` row, as two separate statements with no transaction. Either one failing alone leaves a state nothing repairs:

- **Flag written, row not** (the original order). The order sits in `Dispute` with no row. `get_valid_order` admits only `Active`/`FiatSent` and runs before the sender is identified, so every retry fails with `CantDo(NotAllowedByStatus)`, for the counterparty as much as for the initiator, and solvers can't see it. Corrected from the issue's first description [here](https://github.com/MostroP2P/mostro/issues/921#issuecomment-5397361779).
- **Row written, order not** (what the previous revision of this PR, a reorder, opened; @ermeme caught it). The order stays `Active` with an `initiated` dispute nobody was notified about. Retries hit `DisputeAlreadyExists`, and the escrow-deadline pass only resumes a half-transition from `FiatSent`: on `Active` it cancels the escrow.

Reordering only swaps which half can be stranded.

## The fix: one transaction

Both writes now run inside one `pool.begin()` transaction, so they land together or not at all. `setup_dispute` stays the in-memory validation it already was: a rejection writes nothing, and `Dispute::new` still captures the pre-dispute status for `order_previous_status`.

**Raw sqlx, and why.** `Crud::create`/`Crud::update` (mostro-core 0.14.6) take `&Pool<Sqlite>`, not a generic executor, so they cannot join a transaction. The comment next to the statements says so, along with @Matobi98's point that sqlx-crud, which mostro-core#154 replaced, *was* generic over the executor: this works around something that rewrite narrowed, not a limit that always existed. The order `UPDATE` writes only the three fields `setup_dispute` changes (`status`, `buyer_dispute`, `seller_dispute`) instead of the whole row from a snapshot.

**Why not recover a half-written dispute instead.** @Matobi98 laid out why that loses:

1. The half-state is lossy. `Dispute` has no initiator field, and who filed lives only in the order flags that failed to persist, so a resume could only credit whoever retries.
2. The natural hook, the early `DisputeAlreadyExists` return, runs before `get_valid_order` and before the sender is identified, so a resume there would let a third party drive someone else's order into `Dispute`.

**One more thing the transaction closes.** `disputes.order_id` is `UNIQUE`, so when two parties dispute at once the second insert fails and its whole transaction rolls back, with no half state left behind.

**The update is guarded by the status it validated.** `AND status = ?`, bound to `dispute.order_previous_status` — the status `get_valid_order` admitted — with a `rows_affected() != 1` check that refuses with `CantDo(NotAllowedByStatus)` and rolls the inserted row back with it. Matching on `id` alone would let a release, or a `fiat_sent` that committed in between, be dragged back to `Dispute` from a stale snapshot. The full-row `Crud::update` this replaces had that gap too; @coderabbitai flagged it here after my own `/code-review` did, and it is cheap now that the transaction exists.

## Tests

| Test | Fails against | What it pins |
|---|---|---|
| `dispute_action_leaves_the_order_untouched_when_the_dispute_row_fails` | the original update-first ordering (verified Aug 24: `assertion failed: !stored_order.buyer_dispute`) | the insert-failure side: no flag, no `Dispute` status without the row |
| `dispute_action_rolls_back_when_the_order_moved_on` | the guard removed — verified: without `AND status = ?` the handler stops refusing and the test fails | a transition that commits inside the window is refused, and the inserted row rolls back with it |
| `dispute_action_rolls_back_the_row_when_the_order_update_fails` | the reorder without a transaction, this PR's previous head: `active: the dispute row must roll back with the failed order update` | @ermeme's inverse window, from both `Active` and `FiatSent`; a trigger aborting any `orders` update injects the failure deterministically |

## Verification

Rebased onto `main` @ `e5ea779` (0.18.7).

- `cargo fmt --check` — clean
- `cargo test --bin mostrod` — **1341 passed**, 0 failed, 2 ignored

Local only: `ci.yml` does not build pull requests (#929).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dispute processing reliability by ensuring order details remain unchanged if dispute creation fails.
  * Ensured dispute records and related order changes succeed or fail together.
  * Preserved consistent order statuses and flags when errors occur during dispute processing.
  * Dispute creation now fails safely when an order changes status during processing, preventing unintended updates.
* **Documentation**
  * Updated cooperative-cancellation guidance to reference `CooperativelyCanceled` instead of `SellerRefunded`.


ToRyVand:
Status note, not a nudge: this has been open 15 days with no human review yet. `mergeable` is true and it merges clean; the `blocked` state is only the required review.

Recording what it fixes, since the diff is small enough to hide it: `dispute_action` writes the order's dispute flag **before** inserting the `disputes` row, two statements with no transaction. If the insert fails, the order is flagged as disputed while no dispute exists — and the initiating party is then **permanently unable to dispute that order**, because the flag makes the code treat it as already disputed. `job_escrow_deadline` already does the same two writes row-first and says why in a comment.

The fix is the reorder, 1 file, +54/−7. A transaction would be the stronger fix but `Crud` takes `&Pool`, so it is not available at this layer without a wider change.

Two premises in my own issue (#921) turned out wrong while building this and were corrected publicly on the issue — worth reading before the code if anyone picks this up.

Matobi98:
Hi again @ToRyVand! Had a look at this one. Both of you are right, and I think that points at a third answer rather than a compromise.

**@ermeme's window is real.** `scheduler.rs:895-940`: the `Active` arm cancels the hold invoice and the order without ever calling `find_dispute_by_order_id`. Only the `FiatSent` arm knows how to resume a half-transition. So "create succeeds, update fails" on an `Active` order strands a live dispute nobody was notified about, and the escrow gets cancelled out from under it.

**Your mostro-core argument checks out.** I read 0.14.6: `Crud::create` and `Crud::update` take `&Pool<Sqlite>`, not a generic executor, so a transaction really does mean a mostro-core change or raw sqlx. Not a hand-wave.

**And the reorder is the right direction**, which is worth saying because the thread reads like it trades one broken state for an equally broken one. It doesn't. On the old ordering the order lands in `Dispute`, which `get_valid_order` rejects (`dispute.rs:88`) before the sender is even identified — no retry from either party can reach the function at all. On yours it stays `Active`/`FiatSent`, which `get_valid_order` accepts. The remaining blocker is our own early return at `dispute.rs:149`.

So my first instinct was to suggest recovery instead of a transaction: teach that early return to finish a half-transition the way the `FiatSent` arm already does, and give the scheduler's `Active` arm the same check. Two things killed it.

**1. The half-state is lossy.** `Dispute` (mostro-core 0.14.6) has no initiator field — id, order_id, status, order_previous_status, solver_pubkey, two timestamps. Who filed lives only in the order's `buyer_dispute` / `seller_dispute` flags, which are exactly what failed to persist. A resume cannot restore the initiator; it can only credit whoever retries. Buyer files, write fails, seller retries, dispute recorded as seller-initiated. Not exploitable on demand — you would have to induce the DB failure — but silently wrong, and it races on who retries first.

**2. The natural place for the recovery is the wrong place.** `dispute.rs:149` runs before `get_valid_order` (153) and before `get_counterpart_info` identifies the sender (163). Harmless today, since the branch only returns an error. A resume there would let a third party who is neither buyer nor seller drive someone else's order into `Dispute`.

Both argue for @ermeme's first option over his second: **make the two writes atomic here, with raw sqlx inside a `pool.begin()`**, leaving `setup_dispute` as the in-memory validation it already is. That removes the half-state instead of building machinery to live with it — the right trade on a path whose failure mode is a frozen escrow or a dispute attributed to the wrong party.

On the raw sqlx: worth a comment next to the statements saying why it's there. The pool-only signature isn't a permanent limit — sqlx-crud, which mostro-core#154 replaced, was generic over the executor, so transactions used to compose. The rewrite narrowed it, understandably (most of those bounds existed to stay database-agnostic, which this crate isn't), but it means this is a workaround for something that was lost rather than never available.

ToRyVand:
@Matobi98 thanks. Both of your arguments against recovery hold, and I had missed the first one: with no initiator field on `Dispute`, a resume can only guess. Done the way you proposed:

- The dispute insert and the order update run in one `pool.begin()` transaction, as raw sqlx, and `setup_dispute` stays in-memory validation.
- The comment next to the statements says why the raw sqlx is there, including your mostro-core#154 point: it works around what that rewrite narrowed, not a limit that was always there.
- The order `UPDATE` writes only the three fields `setup_dispute` changes, instead of the whole row from a snapshot.

@ermeme the inverse window has its test: `dispute_action_rolls_back_the_row_when_the_order_update_fails` injects the failure with a trigger that aborts any `orders` update, from both `Active` and `FiatSent`. I checked that it catches the bug rather than just running: against this PR's previous head (the reorder without a transaction) it fails with `active: the dispute row must roll back with the failed order update`, and now it passes.

A side effect worth knowing: `disputes.order_id` is `UNIQUE`, so when two parties dispute at once, the second insert fails and its whole transaction rolls back. No half state there either.

There's one pre-existing gap I didn't touch: the `UPDATE` has no status guard, so a status change between `get_valid_order` and the write would be overwritten. The full-row `Crud::update` had the same gap. I can make it `AND status IN ('active', 'fiat-sent')` with a `rows_affected` check here if you'd like it in this PR, or file it separately.

Rebased onto `main` @ `e5ea779` (0.18.7). `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1340 passed**, 0 failed. As usual, nothing here has CI (#929).

ToRyVand:
@coderabbitai good catch, and it lines up with what my own `/code-review` flagged and what I offered @Matobi98 two comments up — so rather than leave it hanging on an answer, it's in.

The order update is now `WHERE id = ? AND status = ?`, bound to `dispute.order_previous_status`, which *is* the status `get_valid_order` admitted, with a `rows_affected() != 1` check that refuses with `CantDo(NotAllowedByStatus)`. Since both writes are already one transaction, the refusal rolls the inserted dispute row back with it — no extra bookkeeping.

The regression test injects exactly your scenario deterministically: a trigger that moves the order to `settled-hold-invoice` **as soon as the dispute row lands**, inside the same transaction. Without the guard the dispute is written on an order that has moved on; with it, the handler refuses and the row rolls back. Verified by removing the guard and watching the test fail, not just by watching it pass.

@Matobi98 this is the gap I said I'd either do here or file separately — it turned out to be six lines on top of the transaction you asked for, so I did it here. Shout if you'd rather it had been its own PR.

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1341 passed**, 0 failed.

Matobi98:
ACK `fbea573`.

Doing the guard here was the right call — it's six lines on the same transaction, and splitting it out would have shipped the transaction with the gap still open.

One nit on the new comment, not blocking: it names "a release or a `fiat_sent` that committed in between" as what the guard stops. Those can't interleave with `dispute_action`: user messages go through the event loop one at a time, each handler awaited before the next event is read (`app.rs:532-556`). What does run alongside is the scheduler — `enforce_escrow_deadline_pass` (`scheduler.rs:909-970`) moves an `Active` order to `Canceled` from its own task. Naming that job instead would keep the comment from suggesting user messages run concurrently.

Verified:

- `cargo test --bin mostrod`: 1341 passed. With `AND status = ?` removed from the update, `dispute_action_rolls_back_when_the_order_moved_on` fails.
- Manually against a local regtest daemon on this commit: took an order to `Active` with two `mostro-cli` identities and opened a dispute. The order row ended as `dispute` with `seller_dispute = 1`, and exactly one `disputes` row, `initiated`, with `order_previous_status = active`. A second dispute from the other party was refused ("Dispute already exists") and left no second row.

ToRyVand:
Fixed in `06562f8` — and you were right to check the concurrency rather than take the comment's word for it. `handle_message_action` is awaited inline before the next notification is read, so a release or a `fiat_sent` cannot interleave, and the comment said otherwise.

**One addition: there are two of our own tasks, not one.** Alongside `enforce_escrow_deadline_pass`, `hold_invoice_canceled` (`flow.rs`) also moves an `Active` order to `Canceled` — not on every cancel, only when the order is still escrow-backed and past the escrow-action deadline — and it runs on the LND invoice-subscription task, not the event loop. So the guard has two writers to stop and the comment now names both.

Thanks for the regtest run, too — the `order_previous_status = active` row and the refused second dispute are exactly the two things I could not check without a node.

arkanoider (approved):
tACK

arkanoider:
> **Your mostro-core argument checks out.** I read 0.14.6: `Crud::create` and `Crud::update` take `&Pool<Sqlite>`, not a generic executor, so a transaction really does mean a mostro-core change or raw sqlx. Not a hand-wave.

This comment could be a possible future development on `mostro-core`

ToRyVand:
@arkanoider thanks for the `tACK` — and for the rebase and the `persist_dispute` extraction. I checked the rebase with `git range-diff` rather than assume it: both of my commits carry over intact, the only delta being a context line that upstream #967 had already changed.

One question before this can land. The branch is even with `main` at `b34bdf5`, and the API reports `mergeable: true` / `rebaseable: true`, but `mergeable_state` is `blocked` with all three check-runs `skipped`. My permission level on the repo is `pull`, so the branch-protection endpoint 404s for me and I cannot tell which of the two it is: a required context that is still outstanding, or the "a pull request never gets built" situation from #929.

Is there anything left on my side, or is it only waiting on the merge?

Issue #649, opened 2026-03-10
feat: Include instance name in "y" tag for kind 38383 events


Catrya:
Order events (kind 38383) currently use the y tag with only the platform identifier:  `["y", "mostro"]`

While Mostro nodes can be differentiated by their public key, having the name explicitly in the event allows them to be more easily distinguished in order aggregators.

**Proposed Solution**

Append the existing name field from `settings.toml` to the y tag:
eg
```
# NIP-01 Kind 0 Metadata (optional)
# Human-readable name for this Mostro instance
name = "FreeSatoshi"
```

Results in:
event 38383
```
["y", "mostro", "FreeSatoshi"]
```

**Benefits**

• Immediate visual identification without pubkey lookups
• Easier filtering by Mostro Nodes for clients and aggregators
• Backward compatible — clients reading only y[1] continue to work
• Zero configuration changes — reuses existing name field

PR #948, opened 2026-09-04
docs: document the y tag multi-value compatibility contract (#660)

### LINKED ISSUE #660: docs: document y-tag multi-value compatibility contract after #653
## Context
PR #653 expands `y` tags from a single token form like `["y", "mostro"]` to potentially include platform metadata values (e.g. `["y", "mostro", "<instance>"]`).

## Problem
External indexers/clients may parse `y` as a single-value tag and could break if this contract is not documented clearly.

## Requested work
- [ ] Add compatibility section in docs/changelog describing the `y` tag contract
- [ ] Explicitly state:
  - `y[0]` remains `mostro`
  - additional values are optional metadata and must be treated as additive/non-breaking
- [ ] Add migration guidance for consumers that currently assume `len(y)==1`

## Acceptance criteria
- Public-facing docs explain the new tag shape with before/after examples
- Changelog/release notes include compatibility note
- No ambiguity about required vs optional `y` values

## Related
- Implemented in #653

Denver2828:
## Summary

- Adds `docs/Y_TAG_COMPATIBILITY.md` describing the `y` tag contract introduced by #653: the first value is always `mostro`, the second value is the optional `[mostro] name` from settings, additive and non-breaking.
- Fixes stale examples that still showed the pre-#653 shape or a wrong value.
- Documentation only. No protocol or code changes.

## Changes

- New `docs/Y_TAG_COMPATIBILITY.md`: tag shape with explicit indexing (element 0 is the tag name, values start at index 1), before/after examples, the four emitters (order 38383, info 38385, dispute 38386, dev-fee audit 8383; rating 38384 does not carry `y`), the contract, migration guidance for consumers that read one value, assume a fixed value count, or filter with `#y` (NIP-01 indexes only the first value, so `{"#y": ["mostro"]}` keeps working), and history (#649, #653, v0.16.5).
- `docs/README.md`: link to the new page.
- `INSTALL.md`: info event example showed `["y", "mostrop2p"]`.
- `docs/DEV_FEE.md`: audit event example now shows the two-value form.
- `docs/SEPARATE_EVENT_KINDS_SPEC.md`: one sentence before the examples pointing to the contract; examples left as historical.

## Notes for review

- The changelog entry lands through this PR title, since `CHANGELOG.md` is regenerated by git-cliff on release.
- The Contract section states that any future value would be appended after the instance name. That is a maintainer guarantee; please confirm or adjust the wording.
- The `protocol` book already documents the optional second value. A compatibility subsection can be mirrored there in a follow-up if wanted.

## Testing

- [x] `npx markdownlint-cli2@0.23.2`: 42 files, 0 issues (same command as `markdown.yml`).
- [x] Emitters and kinds checked against `src/nip33.rs`, `src/util.rs`, and `mostro-core` 0.14.6 `Rating::to_tags`.

Closes #660

## Summary by CodeRabbit

* **Documentation**
  * Documented the optional Mostro instance name in Nostr event `y` tags.
  * Added compatibility guidance for clients, indexers, relay filters, and platform displays.
  * Updated event examples and developer-fee audit specifications to show the instance name format.
  * Added a quick link to the new `y` tag compatibility documentation.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -237,7 +237,7 @@ In few minutes you should see a nostr event with Mostro settings:
       ["hold_invoice_expiration_window", "900"],
       ["hold_invoice_cltv_delta", "298"],
       ["invoice_expiration_window", "900"],
-      ["y", "mostrop2p"],
+      ["y", "mostro", "<instance name>"],
       ["z", "info"]
     ],
     "content": "",
--- a/docs/DEV_FEE.md
+++ b/docs/DEV_FEE.md
@@ -1145,7 +1145,7 @@ Why kind 8383 (Regular Event)?
     "status": "success"
   },
   "tags": [
-    ["y", "mostro"],
+    ["y", "mostro", "<instance name>"],
     ["z", "dev-fee-payment"],
     ["order", "550e8400-e29b-..."],
     ["amount", "100"],
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,6 +13,7 @@ Quick links to architecture and feature guides.
 - Maintenance Mode & LN Node Migration: MAINTENANCE_MODE_LN_MIGRATION.md (design spec; operator procedure lives in LIGHTNING_OPS.md)
 - RPC Interface Reference: RPC.md
 - NIP-01 Kind 0 Metadata: NIP01_KIND0_METADATA.md
+- `y` Tag Compatibility Contract: Y_TAG_COMPATIBILITY.md (platform identifier plus optional instance name; guidance for indexers and clients)
 
 Tips
 - Run tests and lints before pushing: `cargo test`, `cargo fmt`, `cargo clippy --all-targets --all-features`.
--- a/docs/SEPARATE_EVENT_KINDS_SPEC.md
+++ b/docs/SEPARATE_EVENT_KINDS_SPEC.md
@@ -272,6 +272,8 @@ client.send_event(&new_event).await?;
 
 ## Event Examples
 
+The `y` tag in the examples below shows the single-value form. Since #653 it may carry a second value with the configured instance name, `["y", "mostro", "<instance name>"]`. See `Y_TAG_COMPATIBILITY.md` for the contract.
+
 ### Order Event (Kind 38383 - Unchanged)
 
 ```json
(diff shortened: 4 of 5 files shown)


Catrya (changes requested):
Thanks @Denver2828 
Three things before this goes in.

**1. `SEPARATE_EVENT_KINDS_SPEC.md` wasn't fixed, and one of its examples is now actively contradicted.** In the issue thread you scoped this PR as fixing the stale examples in `INSTALL.md`, `docs/SEPARATE_EVENT_KINDS_SPEC.md` and `docs/DEV_FEE.md`, and grunch approved that scope. Two of the three landed; this one got a pointer sentence instead, and the sentence only qualifies the *value count* ("shows the single-value form"), which implicitly vouches for the rest of each example. The rating example (kind 38384, around line 314) carries `["y", "mostro"]` — a shape mostrod has never emitted, since `Rating::to_tags` produces only the rating fields and `z`. So the file now points readers at a contract page that says the opposite in plain words. Please drop that line from the 38384 example, or say in the added sentence that rating events carry no `y` at all. "Left as historical" is fine for a stale value; it isn't for a tag the event never had.

(That example also uses `min_rating`/`max_rating`/`days` where the code emits `min_rate`/`max_rate` and no `days`. Out of scope — noting it so nobody reads the block as current.)

**2. The forward-looking guarantee can't be honored as worded** — this is the item you flagged for a maintainer. "Mostro will not remove the first value or insert values before it": yes, I'll guarantee that. "Any future value is appended after the instance name": no, because the instance name is *optional*, which leaves the position undefined. On a nameless instance a future third value lands at index 2 — exactly where your own migration snippet reads it:

```
instance_name = tag[2] if len(tag) >= 3 else None
```

so a consumer following this page would display a future metadata value as the instance name. Suggested rewording: index 1 is the only positionally defined value, the instance name is index 2 *when present*, and any future metadata will arrive in its own tag rather than as a third `y` value. That one we can keep.

**3. Nit in `INSTALL.md`.** The old `["y", "mostrop2p"]` was plainly wrong and had to go, but `["y", "mostro", "<instance name>"]` isn't what a reader following that page sees either: `settings.tpl.toml` ships `name` commented out and `MostroSettings::default()` has `name: None`, so a fresh install emits the single-value form. Under a sentence that says "you should see", I'd put `["y", "mostro"]` and let the new page carry the optional second value. `DEV_FEE.md` is a spec example rather than expected output, so the placeholder reads fine there.

Beyond what #660 asked, take it or leave it: exchange rate events (kind 30078, `docs/NOSTR_EXCHANGE_RATES.md`) also carry no `y` — no `z` either. Since the page lists four emitters and then names rating as *the* exception, a consumer concludes every other public Mostro event has one. One clause next to the rating sentence closes it. Same for the protocol book's `order_event.md`, which shows `["y", "lnp2pbot", "[Platform instance name]"]`: value 1 is the *platform* identifier and other platforms on this protocol use their own, so `#y: ["mostro"]` excludes them by design — worth a line for the aggregator audience this page is written for.

PR #957, opened 2026-09-06
chore: relicense project under GPLv3

grunch:
Relicenses this project from MIT to the **GNU General Public License v3.0** (SPDX `GPL-3.0-only`), as part of an org-wide relicensing of every MostroP2P repository.

### Changes

- `Cargo.toml`
- `LICENSE`
- `README.md`

- `LICENSE` now holds the verbatim GPLv3 text from https://www.gnu.org/licenses/gpl-3.0.txt (674 lines).
- The README license section (and badge, where present) points at GPLv3. Where the README had no license section, one was added.
- Package metadata (`Cargo.toml` / `package.json`) declares `GPL-3.0-only` so the license is machine-readable; the field was added where it was previously absent.

### Notes for reviewers

- Vendored third-party files keep their own licenses and were deliberately left untouched.
- MIT is GPLv3-compatible, so the previously MIT-licensed history can be redistributed under GPLv3. Merging this is a maintainer/copyright-holder decision, not a mechanical one — please confirm before merge.

https://claude.ai/code/session_01AeaT3mK6fazbUn9kcTrKxL

## Summary by CodeRabbit

* **Documentation**
  * Updated the project license from MIT to GNU General Public License v3.0-only.
  * Replaced the license text with the full GPLv3 terms.
  * Updated the README license badge and License section to reflect the new license.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "mostro"
 version = "0.18.7"
 edition = "2021"
-license = "MIT"
+license = "GPL-3.0-only"
 authors = ["Francisco Calderón <negrunch@grunch.dev>"]
 description = "Lightning Network peer-to-peer nostr platform"
 homepage = "https://mostro.network"
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,674 @@
-MIT License
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-Copyright (c) 2022 Francisco Calderón
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+                            Preamble
 
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
+  The licenses for most software and other practical works are designed
+to take away your freedom to share and change the works.  By contrast,
+the GNU General Public License is intended to guarantee your freedom to
+share and change all versions of a program--to make sure it remains free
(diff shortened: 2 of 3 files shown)


ToRyVand (commented):
Reviewed this as an audit rather than a read, since a relicense is hard to walk back. The mechanical parts are right: `Cargo.toml` declares valid SPDX, the README badge and the License section both change, and there are no per-file license headers in `src/` to sweep (verified — no file carries `MIT License` or an `SPDX-License-Identifier`).

One thing needs fixing before merge, and one thing I checked that clears the way.

## 1. The new `LICENSE` carries no copyright notice at all

The file it replaces began:

```
MIT License

Copyright (c) 2022 Francisco Calderón
```

The new file is 675 lines of verbatim GPLv3 and nothing else. I checked rather than skimmed: `MIT` occurs in it only inside the words `LIMITED` and `PERMITTED`, and no `Copyright (c) <year>` line matches anywhere in the file.

That drops two things that are not discretionary:

- **MIT's notice-retention condition.** *"The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."* Relicensing forward is fine — MIT permits it, as the PR body says — but the permission that makes it fine is the one carrying that condition. The history being redistributed is MIT-licensed code, so its notice has to survive somewhere: a `LICENSE-MIT` alongside, or a short preamble in `LICENSE` naming the original notice and the portions it covers.
- **GPLv3's own instructions.** The appendix the pasted text ends with (*"How to Apply These Terms to Your New Programs"*) asks for a copyright line, and §5(a) requires each modified work to carry "appropriate copyright notice". As it stands the repository's license file names no copyright holder, which is a strange state for a copyleft licence in particular: the party with standing to enforce it is not identified anywhere.

Both are one or two lines to fix.

## 2. Dependency compatibility — checked, and it passes

This is the check that could have blocked the relicense, so here is the result rather than an opinion. `cargo metadata --all-features` over the full tree, **503 packages**:

| licence | count | GPLv3? |
|---|---|---|
| `MIT OR Apache-2.0` and variants | 259 + 28 + 13 + … | yes |
| `MIT` | 109 | yes |
| **`Apache-2.0`** (alone) | **13** | **yes — v3 only** |
| `Unicode-3.0` | 18 | yes |
| `CC0-1.0` | 16 | yes |
| `BSD-2/3-Clause`, `ISC`, `Zlib`, `0BSD`, `BSL-1.0`, `MITNFA`, `Unlicense` | ~20 | yes |
| `MPL-2.0` — `option-ext` | 1 | yes (§3.3 secondary licence) |
| `WTFPL` — `terminfo` | 1 | yes (FSF-listed) |
| `CDLA-Permissive-2.0` — `webpki-roots` | 2 | permissive, no copyleft — see below |

**Nothing in the tree is GPLv3-incompatible.** No AGPL, no `GPL-2.0-only`, no CDDL, no SSPL/BUSL.

Two notes on that table:

- The 13 **`Apache-2.0`-only** crates are why this had to be GPLv**3**. Apache-2.0 is incompatible with GPLv2 and compatible with GPLv3, so the version choice is load-bearing, not cosmetic — worth a line in the PR body, because the next person to wonder "why not GPLv2" has the answer in the dependency tree.
- `CDLA-Permissive-2.0` (`webpki-roots`) is permissive with no copyleft, so there is no conflict, but it is not on the FSF's vetted list and it is a trust-root bundle rather than code. Probably nothing; flagging it because it is the only entry I could not tie to a published compatibility statement.

## 3. Two questions, not findings

**`mostro-core` is still MIT** — verified on the repository today (`spdx_id: MIT`, last push Sep 21). GPLv3 `mostro` consuming MIT `mostro-core` is fine and nothing here depends on the order. But the body says this is org-wide, and `mostro-core` is the repo in the set with *external* consumers: relicensing it pulls every third-party client that links it into GPLv3, which is a much bigger decision than relicensing the daemon. Is `mostro-core` in scope, and if so is that consequence intended?

**`GPL-3.0-only` rather than `-or-later`** — deliberate? `-only` means a future GPL version needs consent gathered again, and it also forecloses combining with `GPL-2.0-only` code should any ever appear in the tree. Most relicensing projects choose `-or-later` for exactly that reason. The body states the SPDX id explicitly so I assume it is intentional; worth one sentence saying why, since it is the part nobody can change later without another round of this.

**One note for the record:** `mostro` is published on crates.io, so the versions already published under MIT stay MIT for anyone depending on them. That is expected and needs no action, but it means the relicense applies going forward only — which is worth knowing if anyone asks whether 0.18.x is now GPL.

Issue #855, opened 2026-08-07
hold_invoice_paid guard is read-then-act, not an atomic claim


grunch:
Follow-up from the review of #853.

`hold_invoice_paid` (`src/flow.rs`) now guards on the order's status and `invoice_held_at`, but it reads them and then acts without atomicity. A concurrent writer on the main loop — `cancel_order_by_maker` in `src/app/cancel.rs` is the concrete one — can interleave between the read and the write, so the guard can pass against state that is already stale by the time the flow commits.

Adding `invoice_held_at == 0` to the guard (done in #853) closes re-entry from LND redelivering `Accepted`. It does **not** close this race — that is a separate problem and needs a conditional write.

The codebase already has the right pattern: `on_bond_invoice_accepted` in `src/app/bond/flow.rs` claims the transition with an atomic `UPDATE ... WHERE` and branches on `rows_affected`, bailing cleanly when the claim misses. `hold_invoice_paid` should do the same rather than read-then-act.

Low severity — it needs an exact interleaving — but the fix is a known-good pattern already in the tree.

grunch:
Important constraint for whoever picks this up, found while reviewing a suggestion to fix it inside #853.

**An atomic claim cannot simply be moved ahead of the message sends.** `enqueue_order_msg` (`src/util.rs:1355`) pushes onto `MESSAGE_QUEUES.queue_order_msg`, an in-memory `RwLock<Vec<...>>`. Nothing about it is durable.

That makes the current ordering — send messages, then write `invoice_held_at` — accidentally crash-safe in the useful direction. If the process dies in between, the queued messages are lost but `invoice_held_at` is still 0, so the replay on the next restart re-runs the flow and re-sends them.

Invert it to claim-first, and a crash after the claim loses those messages permanently *and* suppresses the replay that would have regenerated them. The buyer never receives `AddInvoice` and the order stalls with the escrow locked. That trades a duplicate-message bug for a lost-message bug, which is the worse of the two.

So this issue is really two pieces of work, in order:

1. A durable outbox (or writing the claim and the messages in one transaction), so work survives a crash after claiming.
2. The atomic `UPDATE ... WHERE` claim itself, branching on `rows_affected` as `on_bond_invoice_accepted` does.

Doing 2 without 1 is a regression.

[...]

Issue #810, opened 2026-07-20
[MEDIUM] admin_cancel lacks the atomic CAS status guard that admin_settle has


grunch:
### Severity: MEDIUM

`admin_cancel` lacks the atomic compare-and-set status guard that `admin_settle` has.

### Where
`src/app/admin_cancel.rs` vs `src/app/admin_settle.rs`.

### Problem
`admin_settle` guards its transition with a conditional update (only acts if the order is still in the expected status), which makes concurrent invocations idempotent. `admin_cancel` does not have the equivalent guard, so two concurrent admin-cancel requests (or an admin-cancel racing another transition) can both proceed. Combined with the irreversible `cancel_hold_invoice` side effect (see the "refund before validation" issue), this is worth hardening.

### Suggested fix
Mirror `admin_settle`'s CAS pattern in `admin_cancel`: transition status atomically and only run the LN side effect when the CAS actually updated the row.

_Found during the deep review accompanying #803._

ToRyVand:
I'd like to take this one, and ideally #809 and #855 along with it — they're three instances of the same defect, so the context carries straight across.

Verified all three premises against current `main` first:

- **#810** — `admin_settle.rs:114` guards with `if result.rows_affected() == 0`; `admin_cancel.rs` has no `rows_affected` check at all. Confirmed exactly as written.
- **#809** — in `admin_settle.rs`, `settle_seller_hold_invoice` runs at line **95** and the CAS guard at line **114**: the irreversible LN side effect fires 19 lines *before* the guard meant to serialize it.
- **#855** — `hold_invoice_paid` (`src/flow.rs:54`) is `if !is_pre_payment || order.invoice_held_at != 0` — a read, then an unconditional act.

The tree already contains the pattern all three need, in two places: `on_bond_invoice_accepted` (`src/app/bond/flow.rs:771`, claiming with `rows_affected() == 1` at `:278`) and `admin_settle`'s own guard. So this is applying an established in-repo idiom to the three call sites that still read-then-act, not inventing anything. Each one gets a regression test that fails against today's ordering.

One scoping question worth a maintainer call before I start: **#809 also names `src/app/release.rs`**, which is the hot path and considerably more delicate than the two admin handlers. I'm happy to harden it too, but if you'd rather leave `release` alone until the pattern has been reviewed on the admin paths, say so and I'll keep it out.

[...]

PR #961, opened 2026-09-11
fix(admin-settle): never abort past the settle (#809)

### LINKED ISSUE #809: [MEDIUM] Hold invoice settled before the CAS status guard (double-settle race)
### Severity: MEDIUM

The hold invoice is settled **before** the compare-and-set status guard runs, opening a double-settle race.

### Where
`src/app/release.rs` and `src/app/admin_settle.rs` — the LN `settle` action executes prior to (or independently of) the atomic status transition that is meant to guard against concurrent handlers acting on the same order.

### Problem
Two near-simultaneous release/settle events for the same order can both pass the pre-check, both trigger the settle side effect, and only then contend on the status write. The window is small and usually benign, but it can produce duplicated settlement attempts and confusing state/log output.

### Suggested fix
- Perform the atomic status CAS (`... WHERE status = <expected>`) **first**, and only proceed to `settle` if the CAS affected a row.
- Treat "0 rows updated" as "already handled" and no-op.

ToRyVand:
## Summary

Refs #809. **Rescoped**: the status claim this PR used to carry is withdrawn — the reasoning is in [this comment](https://github.com/MostroP2P/mostro/pull/961#issuecomment-5770339372) and its [correction](https://github.com/MostroP2P/mostro/pull/961#issuecomment-5770509266). What is left is the defect @Matobi98 found in the [review of 2026-09-15](https://github.com/MostroP2P/mostro/pull/961#issuecomment-5683418581) and @Catrya identified as standing on its own (point 6): **past the settle, nothing may abort.**

Once `settle_seller_hold_invoice` returns, the escrow has moved and the `Dispute` guard above rejects a retry, so an early return has no path back. It strands every `Locked` bond on the order and skips the buyer's payout, with no reconciler that picks either up.

## What changes

Four steps after the settle could still return early. Each now logs and carries on:

| Step | What an early return cost |
|---|---|
| the dispute row update | bonds stranded, payout skipped |
| the dispute event build | same |
| the seller notification | same, over a message nobody could receive |
| the buyer notification | same |

The dispute close is delegated to `close_dispute_after_resolution` (renamed from `close_dispute_after_user_resolution`) rather than hand-rolled. That helper already publishes the dispute event **only when the row actually moved** — the local version announced `Settled` over a dispute the database still recorded as open. Generalizing it also fixed a silently dropped error for its two existing callers: `find_dispute_by_order_id` uses `fetch_one`, so a missing row and a transient DB failure arrived indistinguishably and both were discarded.

The notification fan-out now uses the same `for (role, …)` loop `admin_cancel` already uses, instead of two near-identical blocks.

## What deliberately does not change

Three exits from `main` stay, and each says why in place:

- **the `?` on the settle itself** — nothing has moved when it fires, so the order stays `Dispute` for a retry;
- **the `?` on the status write** — without it `claim_order_payout` cannot claim (it requires `settled-hold-invoice`), so carrying on would not pay the buyer anyway;
- **`return Ok(())` on a CAS miss** — another path owns the transition and completes the payout; continuing would double it.

So the rule is narrower than "nothing aborts": **nothing aborts after the CAS is won.**

## Correcting the previous description

The old body said two concurrent settles "both settled the escrow, the loser having already moved money". @Catrya is right that this is not what the old code did (point 4): a hold invoice settles once, so the loser either got an error from LND or missed the CAS after a no-op — either way it never reached `do_payment` or the bond resolution. #809's own framing, "duplicated settlement attempts and confusing state/log output", is the accurate one.

## Not covered

- **No test reaches this code.** `settle_seller_hold_invoice` takes a concrete `LndConnector`, and `dead_lnd()` fails every call, so every handler test stops at or before the settle. Pinning the rule needs a settle seam that can succeed — worth its own change, and flagged here rather than left implicit.
- **An RPC timeout on a settle LND did apply** still returns `Err` with the escrow gone, and the retry then meets "invoice already settled", which this path does not tolerate the way `admin_cancel` tolerates a gone hold invoice.
- **On a CAS miss the solver's `BondResolution` is dropped**: the path that wins (`release_action`) releases the taker bonds instead of applying the ruling. Pre-existing.

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1348 passed**, 0 failed.

The `release.rs` half of #809 is tracked in #972, which also records why a claim before the Lightning effect is not the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dispute settlement reliability by rejecting ambiguous dispute initiator information.
  * Settlement processing now continues when individual notification keys are missing or invalid.
  * Dispute resolution updates and notifications are handled more reliably across releases, refunds, and settlements.
  * Dispute closure issues no longer prevent successful order resolution or payout processing.

### DIFF
--- a/src/app/admin_settle.rs
+++ b/src/app/admin_settle.rs
@@ -1,8 +1,8 @@
 use crate::app::bond::{self, BondSlashReason};
 use crate::app::context::AppContext;
 use crate::db::{
-    ensure_dispute_finalize_permission, find_dispute_by_order_id, is_assigned_solver,
-    is_dispute_taken_by_admin,
+    claim_order_status, ensure_dispute_finalize_permission, find_dispute_by_order_id,
+    is_assigned_solver, is_dispute_taken_by_admin,
 };
 use crate::lightning::LndConnector;
 use crate::nip33::{create_dispute_event_tags, new_dispute_event};
@@ -113,33 +113,51 @@ pub async fn admin_settle_action(
         }
     };
 
+    // #809: claim `Dispute → SettledHoldInvoice` before the settle, so a
+    // losing concurrent handler never reaches LND; a miss means another path
+    // owns the transition, so return `Ok`. The helper's `cashu_escrow_locked_at
+    // IS NULL` predicate never excludes an order here: `dispatch_cashu` refuses
+    // `AdminSettle`, and the admin RPC server and the escrow-deadline job (the
+    // other writer of `Dispute`) are Lightning-only.
+    if !claim_order_status(pool, order.id, Status::Dispute, Status::SettledHoldInvoice).await? {
+        tracing::warn!(
+            order_id = %order.id,
+            "admin_settle: order transitioned out of dispute concurrently; escrow untouched"
+        );
+        return Ok(());
+    }
+
     // Settle seller hold invoice
-    settle_seller_hold_invoice(event, ln_client, Action::AdminSettled, true, &order)
-        .await
-        .map_err(|e| MostroInternalErr(ServiceError::LnNodeError(e.to_string())))?;
-    // Update order event
+    if let Err(e) =
+        settle_seller_hold_invoice(event, ln_client, Action::AdminSettled, true, &order).await
+    {
+        // The claim is standing and the `Dispute` guard blocks a retry, so
+        // release it, only if the status is still ours. Best effort: the
+        // settle's error is what the caller needs.
+        if let Err(release) =
+            claim_order_status(pool, order.id, Status::SettledHoldInvoice, Status::Dispute).await
+        {
+            error!(
+                order_id = %order.id,
+                "admin_settle: could not release the settle claim after a failed settle; the order \
+                 stays settled-hold-invoice with its escrow unsettled: {release}"
+            );
+        }
+        return Err(MostroInternalErr(ServiceError::LnNodeError(e.to_string())));
+    }
+
+    // The claim already persisted the status; patch only the republished
+    // `event_id` rather than writing back a pre-claim snapshot.
     let order_updated = update_order_event(my_keys, Status::SettledHoldInvoice, &order)
         .await
         .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
 
-    // Persist the status change to DB before calling do_payment (same reason as release_action)
-    let result =
-        sqlx::query("UPDATE orders SET status = ?, event_id = ? WHERE id = ? AND status = ?")
-            .bind(&order_updated.status)
-            .bind(&order_updated.event_id)
-            .bind(order_updated.id)
-            .bind(Status::Dispute.to_string())
-            .execute(pool)
-            .await
-            .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
-
-    if result.rows_affected() == 0 {
-        tracing::warn!(
-            "Order {} not transitioned to settled-hold-invoice: status changed concurrently",
-            order_updated.id
-        );
-        return Ok(());
-    }
+    sqlx::query("UPDATE orders SET event_id = ? WHERE id = ?")
+        .bind(&order_updated.event_id)
+        .bind(order_updated.id)
+        .execute(pool)
+        .await
+        .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
 
     // we check if there is a dispute
     let dispute = find_dispute_by_order_id(pool, order.id).await;
@@ -617,6 +635,107 @@ mod handler_tests {
             result,
             Err(MostroInternalErr(ServiceError::LnNodeError(_)))
         ));
+    }
+
+    /// The two transitions `admin_settle_action` drives through
+    /// `claim_order_status`: the claim, and its release on a failed settle.
+    async fn claim_settle(pool: &SqlitePool, order_id: uuid::Uuid) -> bool {
+        claim_order_status(pool, order_id, Status::Dispute, Status::SettledHoldInvoice)
+            .await
+            .unwrap()
+    }
+
+    async fn release_settle(pool: &SqlitePool, order_id: uuid::Uuid) -> bool {
+        claim_order_status(pool, order_id, Status::SettledHoldInvoice, Status::Dispute)
+            .await
+            .unwrap()
+    }
+
+    /// #809: exactly one caller can win `Dispute → SettledHoldInvoice`. The
+    /// race itself is not deterministically testable; this exclusivity is.
+    #[tokio::test]
+    async fn settle_claim_admits_exactly_one_winner() {
+        let pool = setup_pool().await;
+        let ctx = build_ctx(pool.clone());
+        let seller = Keys::generate().public_key();
+        let buyer = Keys::generate().public_key();
+
+        let order = dispute_order(seller, buyer)
+            .create(ctx.pool())
+            .await
+            .unwrap();
+
+        assert!(
+            claim_settle(ctx.pool(), order.id).await,
+            "the first claim against a disputed order must win it"
+        );
+        assert!(
+            !claim_settle(ctx.pool(), order.id).await,
+            "the second claim must miss — the row has left dispute"
+        );
+
+        let stored = Order::by_id(ctx.pool(), order.id).await.unwrap().unwrap();
+        assert_eq!(stored.status, Status::SettledHoldInvoice.to_string());
+    }
+
+    /// The release is conditional on the status this handler wrote, so a
+    /// release racing a later transition must not drag the order backwards.
+    #[tokio::test]
+    async fn releasing_a_settle_claim_leaves_other_statuses_alone() {
+        let pool = setup_pool().await;
+        let ctx = build_ctx(pool.clone());
+        let seller = Keys::generate().public_key();
+        let buyer = Keys::generate().public_key();
+
+        let mut order = dispute_order(seller, buyer);
+        order.status = Status::CanceledByAdmin.to_string();
+        let order = order.create(ctx.pool()).await.unwrap();
+
+        assert!(
+            !release_settle(ctx.pool(), order.id).await,
+            "a release against a status this handler does not own must miss"
+        );
+        let stored = Order::by_id(ctx.pool(), order.id).await.unwrap().unwrap();
+        assert_eq!(stored.status, Status::CanceledByAdmin.to_string());
+    }
+
+    /// #809: a failed settle must surface and hand the order back to
+    /// `Dispute`; the `Dispute` guard would otherwise block any retry.
+    #[tokio::test]
+    async fn failed_settle_releases_the_claim() {
+        let pool = setup_pool().await;
+        let ctx = build_ctx(pool.clone());
+        let mut ln = dead_lnd().await;
+        let admin = Keys::generate();
+        let seller = Keys::generate().public_key();
+        let buyer = Keys::generate().public_key();
+
+        let mut order = dispute_order(seller, buyer);
+        order.seller_dispute = true;
+        order.preimage = None; // settle short-circuits with InvalidInvoice
+        let order = order.create(ctx.pool()).await.unwrap();
+        assign_solver(ctx.pool(), order.id, &admin.public_key()).await;
+
+        let result = admin_settle_action(
+            &ctx,
+            settle_msg(order.id),
+            &admin_event(admin.public_key()),
+            &admin,
+            &mut ln,
+        )
+        .await;
+
+        assert!(
+            matches!(result, Err(MostroInternalErr(ServiceError::LnNodeError(_)))),
+            "the failed settle must surface, got {result:?}"
+        );
+
+        let stored = Order::by_id(ctx.pool(), order.id).await.unwrap().unwrap();
+        assert_eq!(
+            stored.status,
+            Status::Dispute.to_string(),
+            "a failed settle must release the claim so the solver can retry"
+        );
     }
 }
 


ToRyVand:
Thanks — this is the sharpest form of the residual the body already flags, so let me answer it rather than leave it sitting.

**What I'm not doing here, and why.** A distinct intermediate status means a new `Status` variant, and `Status` lives in `mostro-core` — the same crate boundary that keeps `Crud` on `&Pool` and forced raw sqlx in #923. That is a dependency release, a migration for the stored string, and every `check_status` call site, on a PR whose diff is one file.

**On the crash window specifically: this PR does not open it, it moves it.** Before the change the CAS ran *after* `settle_seller_hold_invoice`, so a cancellation, panic or process exit between the settle and the write left the order in `Dispute` with the escrow already settled. A retry then passes the `Dispute` guard, settles again, and LND answers "already settled" — which this path does not classify as benign (only the bond slash path does), so it returns an error and the order is stuck, with nothing in the row saying the money moved. After the change the same interruption leaves `SettledHoldInvoice` with the escrow unsettled: also stuck, also needing an operator, but the row now says a settle was claimed rather than implying nothing happened.

**On the ambiguous LND result**, the body says it outright: if the settle succeeds and the error path still fires, the claim is released and the order returns to `Dispute` with the escrow settled. Same as before this PR, for the same reason.

**What would actually fix both orderings** is the second half of your suggestion, not the first: reconcile by payment hash, so the daemon can ask LND what happened instead of inferring it from a status. That repairs the pre-existing window too, and it is a job for the payout reconciliation path rather than for this handler. I'd rather file it as an issue with the two failure modes written down than bolt a half version onto this PR — say the word and I'll open it, or I'm happy to be overruled and widen the PR.

Matobi98:
Hi @ToRyVand! Went through this one against `main`, and the ordering looks right to me. Three things before it lands.

**The race is real, and it is gRPC against Nostr.** `main.rs:276-279` starts the admin RPC server in its own task with its own mutex over a clone of the LND client, while the event loop drives the other one (`main.rs:310`). So an RPC settle and a Nostr settle or cancel on the same dispute can interleave. Two RPC calls can't, since `service.rs:154` holds that mutex across the whole handler. Might be worth a line in the description, since it is what makes the fix necessary rather than defensive.

**1. Nothing pins the fix.** I mutated the handler and ran `cargo test --bin mostrod` each time:

| Mutation | Result |
|---|---|
| claim moved back after `settle_seller_hold_invoice` | 1341 passed |
| the `return Ok(())` removed on a claim miss, so the loser settles anyway | 1341 passed |
| the release removed | `failed_settle_releases_the_claim` fails |

Agreed that the interleaving itself isn't deterministically testable, but the property it protects — a lost claim never reaches LND — is, through the claim's other predicate. An order in `Dispute` with `cashu_escrow_locked_at = Some(_)` passes the handler's status read and misses `claim_order_status`. Give it a preimage and the existing `dead_lnd()`, and any settle that gets through comes back as `LnNodeError`:

```rust
order.preimage = Some("00".repeat(32)); // a reached settle would hit the dead node
order.cashu_escrow_locked_at = Some(1); // passes the Dispute read, misses the claim
// ...
assert!(!matches!(result, Err(MostroInternalErr(ServiceError::LnNodeError(_)))));
assert_eq!(stored.status, Status::Dispute.to_string());
```

I tried it locally: it passes on this branch and fails on **both** mutations above, with `LnNodeError(... client error (Connect))`. No production change. The one catch is that it leans on the Cashu predicate as the trigger, so if that predicate ever leaves the helper the test would stop exercising the miss — worth a comment on the test saying so.

**2. A lost claim reports success.** On a miss this returns `Ok(())`. Through `call_admin_settle` that surfaces to the operator as a successful settle when nothing was settled, and a solver on the Nostr path gets no reply. I'd suggest `CantDo(NotAllowedByStatus)`, as #963 does for the same miss on the cancel side. It also matches this handler: a few lines above, an order that isn't in `Dispute` already gets `CantDo`; losing the claim is the same situation, detected a moment later.

**3. The tail after the settle still aborts.** Once the escrow has settled, the `?`s on `update_order_event` (151), the `event_id` patch (155), the dispute row update (170) and `new_dispute_event` (183) return early, skipping `apply_bond_resolution` (242) and `do_payment` (276). The description lists this as a residual, framed around the buyer's payout. In #963 you made the cancel tail best effort for a reason that applies here as well: past the claim a retry is rejected by the `Dispute` guard, so an early return also strands every `Locked` bond on the order. Doing the same here would log those failures and carry on to the bond resolution and the payout; your trigger-based test in #963 would pin it for the dispute row. I only read this one — the tests here can't reach a successful settle, since `dead_lnd()` fails every call — so I haven't run it.

Verified locally on `e92f722`: `cargo test --bin mostrod` 1341 passed. The three residuals in the description hold as written — I checked the rating one against `rate_user.rs:85`.

ToRyVand:
All three are in, on `bc9d18d`, with two deviations worth stating.

**1. The test.** Same shape you proposed, different trigger: a `BEFORE UPDATE … RAISE(IGNORE)` on the claim's own write rather than `cashu_escrow_locked_at`. Forcing the miss through that predicate would pin a locked escrow in `Dispute` as something to refuse, and Track D §5C makes it the buyer-wins path here — so the test would assert what TD-3 has to change, and go quiet if the predicate ever moved. Same reasoning I used on #963.

| Mutation | Before | Now |
|---|---|---|
| claim moved back after `settle_seller_hold_invoice` | 1341 passed | `a_lost_claim_refuses_before_the_settle` fails, `LnNodeError` |
| the `return` removed on a claim miss, so the loser settles anyway | 1341 passed | same test fails, `LnNodeError` |
| the release removed | `failed_settle_releases_the_claim` fails | unchanged |

**2. The lost claim refuses — with `InvalidOrderStatus`, not `NotAllowedByStatus`.** Your argument is the one I followed: it is the same situation the guard above catches, a moment later. But that guard is `order.check_status(Status::Dispute)`, and `check_status` answers `InvalidOrderStatus`; #963's guard is a hand-written `NotAllowedByStatus`, which is why the two handlers differ. So each one now agrees with itself, and a solver that retries and trips the guard is told the same thing twice. If you'd rather the two handlers agree with *each other* instead, say so and I'll flip both — it is the same one-line change either way.

**3. The tail.** Done, and wider than the four you listed: the two `PublicKey::from_str` calls in the notification fan-out could abort past the settle too, over a message nobody could have received. The dispute close is now `close_dispute_after_user_resolution`, the helper `release_action` and `cancel.rs` already use — my hand-rolled version published the `Settled` event even when the row update that was supposed to persist it had failed, which the helper does not. You are right that no test reaches this: a successful settle needs a live node, so the whole tail is unpinned and the body says so.

**The race is now in the description**, as you asked — thanks, it is the thing that makes this a fix rather than a defensive change. I cited the RPC server and the service mutex by name rather than by line, since the numbers had moved.

**One residual, same as on #963.** The claim inherits `cashu_escrow_locked_at IS NULL`, and nothing ever clears that lock (`find_locked_cashu_orders_includes_terminal_status_orders` pins it as a deliberate CF-4 contract), so an order carried from Cashu mode into Lightning mode and disputed there can never be admin-settled. Unreachable today, but it arrives with the claim rather than predating it. A CAS of its own here, or filed against TD-3, which needs one anyway?

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1342 passed**, 0 failed, on `bc9d18d`.

Catrya (changes requested):
**Blocking**

1. **Semantic conflict with #967 (rebase on `main`).** `main` now documents `close_dispute_after_user_resolution` as "Never `Settled`, which marks a solver's admin-settle", and this PR calls it with `DisputeStatus::Settled`. It merges cleanly and compiles, but the helper's contract and its log line ("Dispute {} closed automatically after admin settle") are now wrong. Either generalize the helper (name, doc and log wording) so a solver resolution is a supported caller, or keep a local best-effort close in `admin_settle`.

2. **Claiming before the settle exposes an unsettled escrow to the dev-fee job.** `find_unpaid_dev_fees` selects `status = 'settled-hold-invoice'` every 60s from a separate task, with no check that the escrow was actually settled. Between the claim and `settle_seller_hold_invoice` (a window that is longest exactly when LND is slow, i.e. when the settle is about to fail), the job can pay the dev fee. If the settle then fails, the claim is released back to `Dispute` with `dev_fee_paid = 1`, and if the solver later cancels, the node has paid the dev fee from its own funds.
   A crash between the claim and the settle makes it persistent: the order sits in `settled-hold-invoice` with the HTLC still held; `admin_settle` and `admin_cancel` both refuse it (they require `Dispute`); the escrow-deadline guardian ignores it (`active`/`fiat-sent` only); the dev fee gets paid; and once LND cancels the HTLC near expiry, the seller is refunded, the buyer is never paid, and the order is stranded with no recovery path.
   Please handle this in the PR (e.g. make the dev-fee selection require evidence that the escrow was settled, or add a recovery path for this state), or at minimum document it in the description and open a follow-up issue linked from the code comment.

3. **Use `Refs #809`, not `Closes #809`.** #809 also covers `release.rs`, which was deferred on #810. Closing it here drops that half with no issue tracking it. Either open the `release.rs` issue now and link it, or keep #809 open.

4. **Correct the framing in the description and in the first commit message.** "Both settled the escrow… the loser having already moved money" is not what the old code did: a hold invoice settles once, so the loser either got an error from LND before its CAS or missed the CAS after a no-op settle. Either way it never reached `do_payment` or the bond resolution. #809 itself describes the impact as "duplicated settlement attempts and confusing state/log output". The real impact matters here because it is what the new failure modes (dev fee, seller rating inside the window, crash window, Cashu lock) are being traded against.

**Should fix**

5. **Don't borrow `claim_order_status` here.** Its `cashu_escrow_locked_at IS NULL` predicate is unrelated to this transition. It needs a 12-line caveat to justify, and for a stale Cashu lock it makes the handler answer `InvalidOrderStatus` on an order that *is* in `Dispute`, permanently. Moving the previous CAS (`UPDATE orders SET status = ? WHERE id = ? AND status = ?`) above the settle gives the same exclusivity without the extra predicate or the TD-3 / Track D commentary.

6. **Consider splitting the PR.** `bc9d18d` (nothing aborts past the settle) fixes a real, independent defect: a `?` after the settle stranded every `Locked` bond and skipped the buyer's payout. It stands on its own and could land now. The claim reordering for #809 has a much smaller upside and carries the trade-offs above, so it deserves its own discussion.

**Nits**

7. The claim release ignores `Ok(false)` silently. Log it: it means another path moved the order while the settle was failing, which is worth seeing.
8. `settle_claim_admits_exactly_one_winner` and `releasing_a_settle_claim_leaves_other_statuses_alone` only exercise `db::claim_order_status`, not the handler, so they stay green if the handler stops calling it. Move them next to the helper in `db.rs`, or drop them: `a_lost_claim_refuses_before_the_settle` and `failed_settle_releases_the_claim` already cover the handler.
9. The block above the flag match still opens with "Resolve the dispute initiator *before* the settle", but it no longer resolves anything. Rewrite it as "refuse ambiguous flags before the settle".

Point 4 says that, with the old code, the loser either got an error or had its CAS miss. That holds whichever way LND answers a second settle, so it doesn't depend on something I didn't verify against a real node.


**Addendum to point 3: the claim also regresses release vs admin-settle**

`release_action` on a `Dispute` order settles first and CASes afterwards (`status IN (fiat-sent, dispute)`), returning `Ok(())` silently on a miss. With this claim:

1. The seller's release (Nostr) settles the escrow; status is still `Dispute`.
2. An RPC admin-settle claims `Dispute → SettledHoldInvoice`.
3. Release's CAS matches no row → it returns `Ok` and `do_payment` never runs.
4. Admin-settle's settle hits an already-settled invoice. Assuming LND answers with an error (the repo already classifies "already settled" in `bond/slash.rs`), the claim is released back to `Dispute`.

End state: `Dispute`, escrow settled, buyer unpaid. On `main` the same interleaving is safe: the admin settle errors before its CAS and the release CAS succeeds.

The `release.rs` issue from point 3 should describe this as a regression introduced by the claim. The settle-failure arm should also classify "already settled" instead of releasing the claim as if nothing had moved.

ToRyVand:
Taking your point 6, further than you suggested: **the claim comes out entirely** — from this PR, and from #963, which goes to draft.

It lives in SQLite and the effect it guards lives in LND, so on a failed settle the release cannot tell "nothing moved" from "another path already settled this escrow". That is the root of your point 2 and the addendum, and why each patch only moves the symptom; making it atomic needs the durable outbox #855 is blocked on. It also cannot come out of one handler only — the races are between pairs.

**What is left**: past the settle, a `?` on any of the four steps that follow strands every `Locked` bond and skips the buyer's payout; those get logged. The `?` on the settle itself and `return Ok(())` on a CAS miss both stay as `main` has them — nothing has moved at the first, and the loser must not double the payout at the second. Of your nine: 2, 5, 7 and 8 go with the claim; 1 no longer applies (`main`'s inline close does not call the helper — I will fix the publish ordering you flagged there); 3, 4 and 9 are in the rework. `release.rs` gets its issue either way, describing both interleavings as today's behaviour. **#809 stays open**, with this carried back to it.

ToRyVand:
The `release.rs` issue is filed: #972. It records both interleavings as today's behaviour, and carries the reason a claim-first fix does not close them.

ToRyVand:
**Correcting my comment above** on one sentence and on an attribution.

**The sentence.** "The release cannot tell *nothing moved* from *another path already settled this escrow*" is wrong as written: `is_already_settled_error` in `bond/slash.rs` matches exactly that message. Distinguishing is possible; what it does not give is a correct action for the release — letting the claim go is wrong when the escrow did move, holding it is wrong when nothing moved, and neither pays the buyer. That is the argument I should have made, and it is the stronger one. (The same correction applies to the equivalent sentence on #963, where the mechanism I described is also probably not the branch that fires.)

**The attribution.** The scope this PR keeps — that nothing may abort past the settle, because an early return strands every `Locked` bond and skips the buyer's payout — is **@Matobi98's** finding, from the review here on 2026-09-15, not mine. @Catrya's point 6 recognised `bc9d18d` as already fixing it and suggested it could land on its own. My comment named one reviewer twice and the other not at all, on a PR whose whole shape came out of both reviews.

The decision itself does not change.

ToRyVand:
Pushed the rescope: `7e65249`. The claim and its tests are gone; what is left is the post-settle fix — @Matobi98's finding, which you identified as standing on its own.

**One deviation from what I said earlier.** I told you I would keep a local dispute close rather than touch the helper. `/ponytail-review` then pointed out the local version was 43 lines re-implementing `close_dispute_after_user_resolution` step for step — including the publish-only-if-the-row-moved ordering you asked for, which the helper already had. So I took your point 1's *first* option instead: renamed it to `close_dispute_after_resolution`, documented `Settled` as a supported outcome, and called it.

That turned out to fix something for its two existing callers as well: `find_dispute_by_order_id` uses `fetch_one`, so a missing row and a transient DB failure arrived indistinguishably and `if let Ok(..)` discarded both. `release` and `cancel` now log it too.

**Your nine, against the new scope.** 1 taken as above. 4 and 9 done — the description no longer claims the old code had "the loser having already moved money", which you were right to call out. 2, 5, 7 and 8 went with the claim. 3 is `Refs #809`, and #972 tracks the `release.rs` half.

**Three residuals are in the body under _Not covered_**, the first being that no test reaches this code: `settle_seller_hold_invoice` takes a concrete `LndConnector` and `dead_lnd()` fails every call, so every handler test stops at or before the settle. Pinning the rule needs a settle seam that can succeed, which is its own change.

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1348 passed**, 0 failed.

Issue #929, opened 2026-08-31
ci.yml runs on push only, so pull requests are never built


ToRyVand:
## Summary

`.github/workflows/ci.yml` — the workflow that runs `fmt`, `clippy`, `test` and
`build` — triggers on `push` only:

```yaml
name: CI

on:
  push:
```

There is no `pull_request` trigger, and no branch filter. A push to a fork
branch runs in **that fork's** Actions, so this repository never builds a pull
request. Every other workflow that gates anything already has one:

| Workflow | Triggers |
|---|---|
| **`ci.yml`** — fmt, clippy, test, build | **`push`** |
| `markdown.yml` | `push`, `pull_request` |
| `mutation.yml` | `pull_request`, `schedule`, `workflow_dispatch` |
| `cashu.yml` | `pull_request`, `schedule`, `workflow_dispatch` |

So an incoming PR gets its Markdown linted, but its Rust is never compiled.

## Evidence

Check runs on two open PRs, today:

```
#842  mutation-pr: skipped   mutation-baseline: skipped   mint-integration: skipped
#849  mint-integration: skipped   mutation-baseline: skipped   mutation-pr: skipped
```

No `fmt`, no `clippy`, no `test`, no `build` — those jobs are not merely
passing quietly, they are never scheduled.

## Why it matters

A PR that compiled when it was opened stays green-looking after `main` moves
under it. #849 is the worked example: #892 changed `accept_event`'s signature
while it sat open, and the merge stopped compiling. Nothing surfaced that —
not the PR page, not a status check — until it was checked out by hand weeks
later. `cargo build` alone does not even show it; the broken call sites are in
test code, so it takes `cargo build --all-targets`.

The second-order cost is that "is this PR green?" has no answer anyone can
read. A reviewer has to clone and build to know, which is exactly the review
capacity the project is short of.

Credit where it is due: @Catrya identified this from the symptom while
reviewing #849 — *"ci.yml triggers on: push only, so no build ever ran here to
tell either of us"*. This issue is the mechanism and the evidence behind that.

## Suggested fix

Add the trigger, and scope `push` so same-repo branches don't run twice:

```yaml
on:
  push:
    branches: [main]
  pull_request:

concurrency:
  group: ci-${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

Worth deciding alongside it: whether `fmt`/`clippy`/`test`/`build` should
become required checks for merging, which they cannot be today since they
never report on a PR.

[...]

PR #963, opened 2026-09-13
fix(admin-cancel): claim status before refunding (#810)

### LINKED ISSUE #810: [MEDIUM] admin_cancel lacks the atomic CAS status guard that admin_settle has

ToRyVand:
## Summary

Closes #810. `admin_cancel` guarded its `Dispute → CanceledByAdmin` transition with a read and wrote the status at the **tail** of the handler. Between the two, a concurrent transition on the same order could interleave, and both handlers would act on the same escrow.

This claims the transition before the refund, so the second cancel never reaches the escrow.

**Scope.** `src/app/admin_cancel.rs` only, and only the dispute path. The operator cancel of pre-trade orders (#939) is gated above these status checks and takes its own pre-trade CAS: untouched. Built on #825, now merged, and rebased onto it. #809 (the same defect in `admin_settle`) is #961, opened separately as asked.

## What changed

- **Claim first.** `claim_order_status(pool, order.id, Dispute, CanceledByAdmin)` runs before `cancel_hold_invoice`. It is the helper #830 added and the take flow uses; #961 claims with it too, so the two PRs don't carry near-identical private claims. Its `WHERE` also requires `cashu_escrow_locked_at IS NULL`, which never excludes an order here: nothing in Cashu mode reaches this handler (`dispatch_cashu` refuses `AdminCancel`, the admin RPC server is not started in Cashu mode, and `enforce_escrow_deadline`, the other writer of `Dispute`, is Lightning-only).
- **A miss is refused, not silently accepted.** `CantDo(NotAllowedByStatus)`, the same answer the take flow's claim and this file's own pre-trade CAS give. Returning `Ok(())` would have the gRPC `CancelOrder` surface report a cancellation that never happened.
- **Release on a failed refund, in the transient arm only.** Claiming before an irreversible effect adds a failure mode the old ordering did not have: a refund that never happened would leave the order `CanceledByAdmin` with its escrow held and no retry path, since the `Dispute` guard rejects a second admin cancel. An invoice LND has already canceled or does not know (#941, `tolerate_gone_hold_invoice` → `Ok(false)`) is **not** that case: the HTLC is verifiably gone, the claim stands and the dispute closes.
- **Past the claim, the tail is best effort.** The dispute row, the republished event and its `event_id` are logged on failure instead of returned. They used to be retryable because the status write came last; with the claim hoisted, an early return would leave an unretryable order **and** skip the bond resolution, stranding every `Locked` bond. That path is already best effort for exactly this reason, and the DM fan-out above it says so.
- **`event_id`-only patch** afterwards, instead of writing the whole row back from a snapshot read before the claim.

## What this does not close, stated plainly

The cancel/settle pair is only exclusive once `admin_settle` claims before its own settle as well — that is #961. On `main` it still settles **first** and CASes afterwards, so a settle racing a claimed cancel can still pay the buyer while this refunds the seller. This PR closes the second-cancel race and makes the cancel side claimable; the pair closes when both land.

## Tests

| Test | Fails against | What it pins |
|---|---|---|
| `a_transient_refund_failure_releases_the_claim` | the release removed — verified: `left: "canceled-by-admin"`, `right: "dispute"` | a transient LND failure hands the claim back so the solver can retry |
| `a_post_claim_failure_does_not_abort_the_handler` | the `?` restored on the dispute-row update — verified: it returns `DbAccessError(injected)` instead of running on | a failure past the claim does not skip the notifications and the bond resolution |

The already-gone arm (`Ok(false)`) keeps its existing unit coverage in `gone_hold_invoice_is_not_an_error`; the release lives in the `Err` arm by construction, so it cannot fire there.

## Verification

Rebased onto `main` @ `e5ea779` (0.18.7).

- `cargo fmt --check` — clean
- `cargo test --bin mostrod` — **1340 passed**, 0 failed, 2 ignored

Local only: `ci.yml` does not build pull requests (#929).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved admin dispute cancellations to prevent duplicate processing during escrow refunds.
  * Transient refund failures now return disputes to a retryable state.
  * If cancellation ownership is lost before the refund, the escrow remains held and the situation is reported.
  * Cancellation completion is more resilient when follow-up updates or event notifications fail.

* **Tests**
  * Added coverage for refund failures, lost cancellation ownership, and post-cancellation processing errors.

### DIFF
--- a/src/app/admin_cancel.rs
+++ b/src/app/admin_cancel.rs
@@ -3,8 +3,8 @@ use std::str::FromStr;
 use crate::app::bond::{self, BondSlashReason};
 use crate::app::context::AppContext;
 use crate::db::{
-    ensure_dispute_finalize_permission, find_dispute_by_order_id, is_assigned_solver,
-    is_dispute_taken_by_admin,
+    claim_order_status, ensure_dispute_finalize_permission, find_dispute_by_order_id,
+    is_assigned_solver, is_dispute_taken_by_admin,
 };
 use crate::lightning::LndConnector;
 use crate::nip33::{create_dispute_event_tags, new_dispute_event};
@@ -55,7 +55,12 @@ use tracing::{error, info, warn};
 ///
 /// Failures *after* the refund are logged and swallowed rather than
 /// returned (DM fan-out, bond resolution), so the handler always runs to
-/// the end once the escrow has moved. Making that tail atomic is #810.
+/// the end once the escrow has moved.
+///
+/// The `Dispute → CanceledByAdmin` transition is claimed atomically before
+/// the refund, so a second cancel cannot reach that escrow move (#810). The
+/// cancel/settle pair becomes exclusive once `admin_settle` claims before
+/// its own settle too (#809): on `main` it still settles first.
 pub async fn admin_cancel_action(
     ctx: &AppContext,
     msg: Message,
@@ -153,8 +158,8 @@ pub async fn admin_cancel_action(
     // in `Dispute`, so the LN side and the DB disagreed with no way back.
     // This initiator check now precedes the refund, as does the counterparty
     // resolution below. What remains after the refund either can't reject the
-    // call (the DM fan-out is best effort) or is a DB/event write whose
-    // atomicity is tracked separately in #810.
+    // call (the DM fan-out is best effort) or is a DB/event write serialized
+    // by the status claim below (#810).
     let dispute_initiator = match (order.seller_dispute, order.buyer_dispute) {
         (true, false) => "seller",
         (false, true) => "buyer",
@@ -191,14 +196,47 @@ pub async fn admin_cancel_action(
         (None, _) | (_, None) => return Err(MostroInternalErr(ServiceError::InvalidPubkey)),
     };
 
+    // #810: claim `Dispute → CanceledByAdmin` before the escrow moves, so a
+    // concurrent transition cannot act on the same escrow. A miss means
+    // another path owns the transition. The helper's `cashu_escrow_locked_at
+    // IS NULL` predicate never excludes an order here, as nothing in Cashu
+    // mode reaches this handler.
+    if !claim_order_status(pool, order.id, Status::Dispute, Status::CanceledByAdmin).await? {
+        warn!(
+            order_id = %order.id,
+            "admin_cancel: order moved out of dispute concurrently; escrow untouched"
+        );
+        // Refused like the take flow's claim: a silent `Ok` would have the
+        // gRPC surface report a cancellation that never happened.
+        return Err(MostroCantDo(CantDoReason::NotAllowedByStatus));
+    }
+
     if let Some(hash) = order.hash.as_ref() {
         // We return funds to seller. A hold invoice that LND already
         // canceled (CLTV expiry refunded the seller long ago) or does not
         // know at all (the daemon now runs against a different node) is
         // not an error: the HTLC is verifiably gone and the dispute can
         // still be closed. Only a transient LND failure aborts the cancel.
-        if tolerate_gone_hold_invoice(&order.id, ln_client.cancel_hold_invoice(hash).await)? {
-            info!("Order Id {}: Funds returned to seller", &order.id);
+        match tolerate_gone_hold_invoice(&order.id, ln_client.cancel_hold_invoice(hash).await) {
+            Ok(true) => info!("Order Id {}: Funds returned to seller", &order.id),
+            // Already gone: the claim stands and the dispute still closes.
+            Ok(false) => {}
+            Err(e) => {
+                // Only this arm releases: the guard above rejects a second
+                // admin cancel, so a claim standing on a refund that never
+                // happened would strand the escrow with no retry path.
+                if let Err(release) =
+                    claim_order_status(pool, order.id, Status::CanceledByAdmin, Status::Dispute)
+                        .await
+                {
+                    error!(
+                        order_id = %order.id,
+                        "admin_cancel: could not release the cancel claim after a failed refund; \
+                         the order stays canceled-by-admin with its escrow held: {release}"
+                    );
+                }
+                return Err(e);
+            }
         }
     }
 
@@ -210,9 +248,15 @@ pub async fn admin_cancel_action(
         let opened_at = d.created_at;
         // we update the dispute
         d.status = DisputeStatus::SellerRefunded.to_string();
-        d.update(pool)
-            .await
-            .map_err(|e| MostroInternalErr(ServiceError::DbAccessError(e.to_string())))?;
+        // Past the claim the order is committed — the `Dispute` guard rejects
+        // a retry — so a failure here must not skip the notifications and the
+        // bond resolution below. Same reasoning the DM fan-out carries.
+        if let Err(e) = d.update(pool).await {
+            error!(
+                order_id = %order.id,
+                "admin_cancel: could not move the dispute row to seller-refunded: {e}"
+            );
+        }
         // We create a tag to show status of the dispute
         let tags = create_dispute_event_tags(
             DisputeStatus::SellerRefunded.to_string(),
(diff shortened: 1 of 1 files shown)


Matobi98:
Hi @ToRyVand! Read this one alongside #961, and it holds up. One thing before it lands, the same one I raised there.

**Nothing pins the fix.** I mutated the handler and ran `cargo test --bin mostrod` each time:

| Mutation | Result |
|---|---|
| claim moved after the refund | 1340 passed |
| the `return Err(CantDo(NotAllowedByStatus))` removed on a claim miss, so the loser refunds anyway | 1340 passed |
| the release removed | `a_transient_refund_failure_releases_the_claim` fails |

The miss can be forced through the claim's other predicate: an order in `Dispute` with `cashu_escrow_locked_at = Some(_)` passes the handler's status read and misses `claim_order_status`. With a hash and `dead_lnd()`, a refund that gets through comes back as `LnNodeError`:

```rust
order.hash = Some("11".repeat(32));      // a reached refund would hit the dead node
order.cashu_escrow_locked_at = Some(1);  // passes the Dispute read, misses the claim
// ...
assert!(!matches!(result, Err(MostroInternalErr(ServiceError::LnNodeError(_)))));
assert_eq!(stored.status, Status::Dispute.to_string());
```

Tried locally: passes on this branch, fails on both of the first two mutations with `LnNodeError("code=Unknown message=client error (Connect)")`. It leans on the Cashu predicate as the trigger, so a comment on the test saying so would keep it from silently testing something else if that predicate ever leaves the helper.

Verified locally on `adde0ec`: `cargo test --bin mostrod` 1340 passed. With #961 merged on top, 1343 passed.

ToRyVand:
Thanks — reproduced your table first: both mutations left the suite green here too.

**Took the test, changed the trigger.** Forcing the miss through `cashu_escrow_locked_at` pins a locked escrow in `Dispute` as something to *refuse*, and Track D §5C (`docs/cashu/05-track-d-dispute.md`) makes it the seller-wins path — the *never-locked* one is `CantDo(CashuEscrowNotLocked)`. The test would have pinned what TD-3 has to change, and gone quiet if that predicate moved. Instead: a `BEFORE UPDATE … RAISE(IGNORE)` trigger on the claim's own write, the way `a_post_claim_failure_does_not_abort_the_handler` injects its failure. Same `rows_affected() == 0`, no unrelated predicate, survives TD-3.

On `60760ce`:

| Mutation | Before | Now |
|---|---|---|
| claim moved after the refund | 1340 passed | `a_lost_claim_refuses_before_the_refund` fails, `LnNodeError` |
| the refusal removed on a miss | 1340 passed | same test fails, `LnNodeError` |

**Two from my own review, both log-only.** The release reported `Err` but not `Ok(false)` — a release that matched no row is the same operator problem, and the log was silent in half the cases it exists for. And the forward warning named a cause it cannot observe (*"order moved out of dispute concurrently"*); your own suggested test drives that branch with the order still in `Dispute`. The state is identical either side, so nothing pins them — happy to split them out.

**One residual, and it applies to #961 the same way.** The claim inherits `cashu_escrow_locked_at IS NULL`, and nothing ever clears that lock (`find_locked_cashu_orders_includes_terminal_status_orders` pins it as a deliberate CF-4 contract). So a daemon restarted in Lightning mode against a database that ran in Cashu mode can have an order disputed here and then never admin-cancelled. Unreachable today — `dispatch_cashu` refuses `AdminCancel` — but it arrives with this PR rather than predating it, since there was no claim to miss before. A CAS of its own here without the predicate, or file it against TD-3, which needs one anyway? The reasoning is in the comment above the claim.

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --bin mostrod`: **1341 passed**, 0 failed, on `60760ce`.

Matobi98:
Thanks — the trigger is the better test: it forces the claim miss directly, so it won't go quiet when TD-3 changes what a locked Cashu escrow in `Dispute` means.

**Residual: file it against TD-3.** TD-3 rewrites the Cashu branches of `admin_cancel` anyway and needs its own claim there, and a second near-identical claim here is what this PR avoided. It is also not reachable on a Lightning-only node: `cashu_escrow_locked_at` is only written by the Cashu lock flow, so a Lightning order never carries it and admin cancel keeps working for every Lightning order after this lands.

**The two log changes can stay.** They are log-only and don't change behaviour.

Catrya (changes requested):
**Blocking**

1. **The claim turns a benign release/cancel race into stuck funds.** `release_action` (on a `Dispute` order) and, on `main`, `admin_settle` settle the hold invoice first and CAS afterwards, returning `Ok(())` silently on a miss. With this claim in place, a settle that wins at LND can lose the DB race:
   1. The settle path settles the escrow; status is still `Dispute`.
   2. An RPC `admin_cancel` claims `Dispute → CanceledByAdmin`.
   3. The settle path's CAS matches no row → it returns `Ok` and `do_payment` never runs.
   4. `cancel_hold_invoice` fails with "invoice already settled", which `classify_cancel_error` treats as `Transient` → the claim is handed back to `Dispute`.

   End state: order in `Dispute`, escrow settled into the node, buyer unpaid, seller unrefunded. Every retried cancel fails the same way, and admin-settle hits an already-settled invoice. On `main` the same interleaving is safe: the cancel fails without writing anything and the settle's CAS succeeds.

   Since `release.rs` was scoped out on #810, please:
   - Make sure this lands **after #961**, which closes the `admin_settle` half.
   - Make sure the `release.rs` follow-up issue exists and names this stuck state explicitly. After this PR it's a regression, not pre-existing debt.
   - In the `Err` arm, stop treating "invoice already settled" as transient. Classify it (the settle classifier in `bond/slash.rs` already matches that message) and log an explicit error saying the escrow was settled by another path and needs a manual payout, instead of handing the claim back and reporting an LND failure.

2. **The residual-risk note is inaccurate.** The commit message and the doc comment say a settle racing a claimed cancel "can still pay the buyer while this refunds the seller". LND won't both settle and cancel the same HTLC. The real residual risk is the stuck state above, and #961 only closes it for `admin_settle`, not for `release`. Please correct both.

**Should fix**

3. **Don't borrow `claim_order_status` here.** Its `cashu_escrow_locked_at IS NULL` predicate has nothing to do with this transition. It needs a 12-line TD-3 / Track D caveat to justify it, and with a stale Cashu lock it makes the handler refuse an order that *is* in `Dispute`, permanently. A plain `UPDATE orders SET status = ? WHERE id = ? AND status = ?` above the refund gives the same exclusivity without the extra predicate or the commentary.

4. **Early returns remain past the claim.** `new_dispute_event(...)?` and `message.as_json()?` still return early after the claim, which skips the bond resolution and strands every `Locked` bond. That contradicts the PR's own reason for making the tail best effort. Log them like the other post-claim failures.

**Nits**

5. `a_post_claim_failure_does_not_abort_the_handler` only asserts `is_ok()` and the order status. The stated reason is not stranding `Locked` bonds, so add a locked bond and assert it gets resolved.
6. The comment "a DB/event write serialized by the status claim below (#810)" overstates it. The cooperative cancel path (`cancel_cooperative_execution_step_2`) doesn't claim and does a full-row `update`, so it can still overwrite `CanceledByAdmin`. Please reword it.
7. Align the refusal reason with #961: `NotAllowedByStatus` here, `InvalidOrderStatus` there. Either is fine, as long as both handlers answer the same.

ToRyVand:
Going to draft: **the claim comes out**, and this PR was the claim.

The reasoning is on #961. Short version: it lives in SQLite while the effect it guards lives in LND, so on a failed refund the release cannot tell "nothing moved" from "another path already settled this escrow" — the shared root of both interleavings you described. Patching either one only moves the symptom, and making it atomic needs the durable outbox #855 is blocked on. It also cannot come out of one handler only, since the races are between pairs.

That retires points 2, 3, 4, 6 and 7 with it, and 5 with the test that pinned it. Point 1's sub-asks survive differently: the `release.rs` issue still gets filed, describing both interleavings as today's behaviour rather than regressions — without the claim neither PR introduces them. And classifying "already settled" is no longer load-bearing, since `tolerate_gone_hold_invoice` predates this PR and on `main` a cancel that loses to a settle aborts without writing, which you confirmed is safe; I'd still file it as a separate log-quality improvement if you think it earns one.

**#810 stays open**, with this carried back to it: the read-then-act it describes is real, but a CAS before the Lightning effect is not the answer.

ToRyVand:
The `release.rs` issue is filed: #972. It records both interleavings as today's behaviour, and carries the reason a claim-first fix does not close them.

ToRyVand:
**Correcting my comment above** — the mechanism I described is probably not the one that fires, and one sentence in it is contradicted by this repo.

**The sentence.** "The release cannot tell *nothing moved* from *another path already settled this escrow*" is wrong as written: `is_already_settled_error` in `bond/slash.rs` matches exactly that message — the function you pointed me at. Distinguishing is possible. What I should have written is narrower and, I think, stronger: **distinguishing does not give the release a correct action.** Letting the claim go is wrong when the escrow did move; holding it is wrong when nothing moved; and neither pays the buyer.

**The mechanism.** `classify_cancel_error` tests `code=alreadyexists` in its first branch, and this repo states that LND answers `AlreadyExists` for an already-settled HTLC (`slash.rs`, and the fixture `code=AlreadyExists: invoice already settled`). Under that reading the classification is `AlreadyDone`, `tolerate_gone_hold_invoice` returns `Ok(false)`, the `Ok(false) => {}` arm keeps the claim, and the cancel **runs to completion** over an escrow another path settled — leaving `CanceledByAdmin`, the buyer unpaid, and the losing settle already returned. That is terminal, where the state I described was at least still `Dispute`. You hedged this ("assuming LND answers with an error") and I dropped the hedge; which branch actually fires is not resolvable from the tree.

Neither correction changes the decision. Both readings are states `main` does not produce, and the claim is what produces them.

